PR #1, opened 2015-09-17
Fix handling of unset metric type in decoder.

juliusv:


juliusv:
@beorn7 @fabxc

beorn7:
Have you verified this is the only time this is handled incorrectly?
Even if we check for nil, we have to make sure the semantics is correct.
Now we know it doesn't mean "unset", it means "unset or set to default value".

juliusv:
@beorn7 I only checked the `expfmt` package and this is the only misuse of the `Type` field I could find in there. It was using `GetType()` already before we switched to `expfmt` by the way.

Well, semantically, if someone doesn't set it, it seems to mean "counter" now - but even if the intention was "untyped", the code for decoding counters and untyped metrics is identical, so the returned sample vector is the same.

beorn7:
:+1: for fixing Type. But we have to check _everything_. In the PGW, the problem is with label values, as an example.

juliusv:
Yes, of course. This only fixes one regression that currently crashes the Prometheus server. The rest needs a closer look and fix, but I'll leave that up to you if I may.

PR #3, opened 2015-09-18
Separate response format detection from NewDecoder

fabxc:
This turns 

```
dec, err := NewDecoder(r, resp.Header)
```

into

```
respfmt, err := ResponseFormat(resp.Header)
if err != ...
dec, err := NewDecoder(respfmt)
```

This simplifies testing as one can simply initialize a desired decoder with

```
expfmt.NewDecoder(expfmt.FmtText)
```

 @beorn7 @eparis @piosz

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -36,13 +36,13 @@ type DecodeOptions struct {
 	Timestamp model.Time
 }
 
-// NewDecoder returns a new decoder based on the HTTP header.
-func NewDecoder(r io.Reader, h http.Header) (Decoder, error) {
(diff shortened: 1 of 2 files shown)


timstclair:
Yes, this looks much better than our proposed hack of:
```go
hdr := http.Header{"Content-Type": []string{"text/plain"}}
decoder, err := expfmt.NewDecoder(strings.NewReader(body), hdr)
```

juliusv:
:+1:

beorn7:
:+1: Please remember to fix the callers within the Prometheus org. (Go oracle is really handy to find all the callers.)

PR #4, opened 2015-09-18
Fix direct acess of protobuf fields

beorn7:
@fabxc @juliusv Please have a close look. This might or might not have implications.


fabxc:
:+1:

PR #5, opened 2015-09-21
Backport json 0.0.2 format

fabxc:
@beorn7 @juliusv

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -46,12 +46,12 @@ func ResponseFormat(h http.Header) (Format, error) {
 	}
 
 	const (
-		protoType = ProtoType + "/" + ProtoSubType
-		textType  = "text/plain"
+		textType = "text/plain"
+		jsonType = "application/json"
 	)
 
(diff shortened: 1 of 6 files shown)


juliusv:
Curious: Is this "because we can" or was there a new need for this that I'm not aware of yet?

fabxc:
To many *-kits out there, applications with possibly pinned versions of such kits. In total, dragging this with us probably saves us some trouble.

juliusv:
Hmm, is there a reason to assume that any of these kits ever used the JSON format, given that we've never documented it, and at the time of going public, we didn't have any client library anymore which would default to that format?

juliusv:
s/default to that format/not support protobuf when asked/

juliusv:
(or text)

fabxc:
I'm not referring to public kits :)

juliusv:
:+1: after chatting. Ok, let's put this back temporarily, but make a plan for phasing out the remaining JSON endpoints and then getting rid of this forever again...

PR #6, opened 2015-09-21
Simplefmt

fabxc:
Neither `ResponseFormat` nor `NewDecoder` return an error now. If one wants to check whether format was valid, one can validate that `ResponseFormat() != FmtUnknown`. Having a `Format` value that is not one of the constants is a programming flaw, so nothing we should catch via an extra error in `NewDecoder`.

@juliusv @beorn7

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -43,7 +43,7 @@ func ResponseFormat(h http.Header) Format {
 
 	mediatype, params, err := mime.ParseMediaType(ct)
 	if err != nil {
-		return FmtUnkown
+		return FmtUnknown
 	}
 
 	const (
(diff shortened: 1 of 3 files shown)


expfmt/decode.go
```diff
@@ -37,12 +37,13 @@ type DecodeOptions struct {
...
 // ResponseFormat extracts the correct format from a HTTP response header.
-func ResponseFormat(h http.Header) (Format, error) {
+// If no matching format can be found FormatUnknown is returned.
+func ResponseFormat(h http.Header) Format {
 	ct := h.Get(hdrContentType)
 
 	mediatype, params, err := mime.ParseMediaType(ct)
 	if err != nil {
-		return "", fmt.Errorf("invalid Content-Type header %q: %s", ct, err)
+		return FmtUnkown
```
juliusv: Unkown -> Unknown

juliusv:
:+1: otherwise

beorn7:
"Unknown" clearly has too many "n"s... 

:+1:

Issue #9, opened 2015-10-01
Implement CI


mvader:
Implement a CI like travis or something would be great because mistakes that cause a compile error could be easily avoided (or at least detected).

juliusv:
Definitely. Our other main repos have it, but this one's a new one and we still need to add it. On it.

PR #10, opened 2015-10-01
Add .travis.yml for CI.

juliusv:
Fixes https://github.com/prometheus/common/issues/9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+sudo: false
+
+language: go
+go:
+  - 1.4.3
+  - 1.5.1
+  - tip
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Common
+[![Build Status](https://travis-ci.org/prometheus/common.svg)](https://travis-ci.org/prometheus/common)
 
 This repository contains Go libraries that are shared across Prometheus
 components and libraries.


juliusv:
@fabxc @matthiasr @grobie any of you please :)

.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+  - 1.4.3
+  - 1.5.1
+  - tip
+
+install:
+  - go get -d ./...
+
+script:
+  - go test -v ./...
+  - go build -v ./...
```
grobie: What does build catch what's not catched by test?

juliusv: Ah right, `test` even builds packages without tests. Removed the `build`.

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+go:
+  - 1.4.3
+  - 1.5.1
+  - tip
+
+install:
+  - go get -d ./...
+
+script:
+  - go test -v ./...
```
grobie: There is no need to define `install` or `script`. Travis executes exactly these commands by default.

grobie:
:+1:

.travis.yml
juliusv: Ah, Travis magic. Ok, removed.

PR #11, opened 2015-10-10
Add structured logging to json and support for syslog

mattkanwisher:
We use kibana currently to manage our centralized logging, adding json support will allow native integration with tools like this and logstash. In addition I added syslogger support so people don't need to do manual piping to syslog

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -16,6 +16,8 @@ package log
 import (
 	"flag"
 	"fmt"
+	"net/url"
+	"os"
 	"runtime"
 	"strings"
 
@@ -39,10 +41,68 @@ func (f levelFlag) Set(level string) error {
 	return nil
 }
 
+func setSyslogFormatter(appname, facility string) error {
+	if appname == "" || facility == "" {
+		return fmt.Errorf("your missing a appname(%s) or local(%s)", appname, facility)
+	}
+
+	fmter, err := newSyslogger(appname, facility, origLogger.Formatter)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
+		origLogger.Errorf("can't connect logger to syslog: %v", err)
+		return err
+	}
+	origLogger.Formatter = fmter
+	return nil
+}
+
+func setJSONFormatter() {
+	origLogger.Formatter = &logrus.JSONFormatter{}
+}
+
+type logFormatFlag struct{ uri string }
+
+// String implements flag.Value.
+func (f logFormatFlag) String() string {
+	return f.uri
+}
+
+// Set implements flag.Value.
+func (f logFormatFlag) Set(format string) error {
+	f.uri = format
+	u, err := url.Parse(format)
+	if err != nil {
+		return err
+	}
+	if u.Scheme != "logger" {
+		return fmt.Errorf("invalid scheme %s", u.Scheme)
+	}
+	jsonq := u.Query().Get("json")
+	if jsonq == "true" {
+		setJSONFormatter()
+	}
+
+	switch u.Opaque {
+	case "syslog":
+		appname := u.Query().Get("appname")
+		facility := u.Query().Get("local")
+		return setSyslogFormatter(appname, facility)
+	case "stdout":
+		origLogger.Out = os.Stdout
+	case "stderr":
+		origLogger.Out = os.Stderr
+
+	default:
+		return fmt.Errorf("unsupported logger %s", u.Opaque)
+	}
+	return nil
+}
+
 func init() {
-	// In order for this flag to take effect, the user of the package must call
+	// In order for these flags to take effect, the user of the package must call
 	// flag.Parse() before logging anything.
 	flag.Var(levelFlag{}, "log.level", "Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].")
+	flag.Var(logFormatFlag{}, "log.format", "If set use a syslog logger or JSON logging. Example: logger:syslog?appname=bob&local=7 or logger:stdout?json=true. Defaults to stderr.")
 }
 
 type Logger interface {
--- a/log/syslog_formatter.go
+++ b/log/syslog_formatter.go
@@ -1,0 +1,84 @@
+package log
+
+import (
+	"fmt"
+	"log/syslog"
+	"os"
+
+	"github.com/Sirupsen/logrus"
+)
+
+var ceeTag = []byte("@cee:")
+
+type syslogger struct {
+	wrap logrus.Formatter
+	out  *syslog.Writer
+}
+
+func newSyslogger(appname string, facility string, fmter logrus.Formatter) (*syslogger, error) {
+	priority, err := getFacility(facility)
+	if err != nil {
+		return nil, err
+	}
+	out, err := syslog.New(priority, appname)
+	return &syslogger{
+		out:  out,
+		wrap: fmter,
+	}, err
+}
+
+func getFacility(facility string) (syslog.Priority, error) {
+	switch facility {
+	case "0":
+		return syslog.LOG_LOCAL0, nil
+	case "1":
+		return syslog.LOG_LOCAL1, nil
+	case "2":
+		return syslog.LOG_LOCAL2, nil
+	case "3":
+		return syslog.LOG_LOCAL3, nil
+	case "4":
+		return syslog.LOG_LOCAL4, nil
+	case "5":
+		return syslog.LOG_LOCAL5, nil
+	case "6":
+		return syslog.LOG_LOCAL6, nil
+	case "7":
+		return syslog.LOG_LOCAL7, nil
+	}
+	return syslog.LOG_LOCAL0, fmt.Errorf("invalid local(%s) for syslog", facility)
+}
+
+func (s *syslogger) Format(e *logrus.Entry) ([]byte, error) {
+	data, err := s.wrap.Format(e)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "syslogger: can't format entry: %v\n", err)
+		return data, err
+	}
+	// only append tag to data sent to syslog (line), not to what
+	// is returned
+	line := string(append(ceeTag, data...))
+
+	switch e.Level {
+	case logrus.PanicLevel:
+		err = s.out.Crit(line)
+	case logrus.FatalLevel:
+		err = s.out.Crit(line)
+	case logrus.ErrorLevel:
+		err = s.out.Err(line)
+	case logrus.WarnLevel:
+		err = s.out.Warning(line)
+	case logrus.InfoLevel:
+		err = s.out.Info(line)
+	case logrus.DebugLevel:
+		err = s.out.Debug(line)
+	default:
+		err = s.out.Notice(line)
+	}
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "syslogger: can't send log to syslog: %v\n", err)
+	}
+
+	return data, err
+}


juliusv:
@fabxc @brian-brazil @beorn7 So I've been neglecting this PR here for a while, but talked to @mattkanwisher from DigitalOcean at VelocityConf in NYC about this.

What do you think? Is it worth adding these features to our logging library directly? It would mean that any user of the logging library needs to call the `SetFormatter()` (or whatever it will be named in the end) initialization method to initialize logging properly. That's because there are multiple interdependent flags, so configuring the logger can't just be solved in each flag's custom parsing code, like we currently do it with the log level.

Alternatively, what about adding only JSON support, because that is harder to solve externally? Prometheus would then still focus on only logging to stdout and leave any syslog support up to a wrapper.

brian-brazil:
I think having syslog out of the box would be useful, as pretty much everything log-related integrates with it. JSON seems a little more custom, so we should probably draw a line in the sand now rather than dealing with feature creep later.

mattkanwisher:
@brian-brazil structured logging isn't really a feature creep as the library already is doing it. Honestly the current structured logging isn't useful to ingest into logging systems like kibana. We can do something where have a Database like DSN to specify logging settings "rsylog:local7?format=json" or something.

juliusv:
Yeah, it's harder to structure the logs externally once they're already outputted unstructured.

Supporting machine-readable structured output would however mean that any subsequent change to the format (whether through switching to a different underlying library than logrus or else) we'll have to treat as a breaking interface change in all components using the library (with only human-readable format, you usually don't care). Just a thing to keep in mind.

taotetek:
Please note that syslog-ng and rsyslog both support JSON parsing natively.  See:

* https://algernon.blogs.balabit.com/2012/02/cee-handling-with-syslog-ng-ose/
* http://www.rsyslog.com/doc/mmjsonparse.html

Which is part of our interest in @cee JSON support.

beorn7:
Not really deep into this subject right now. Intuitively, I'd say if our logging is structured internally, we should have a way to output it in a structured form.
Happy to accept any outcome...

mattkanwisher:
@juliusv we can certainly change the code so the interface doesn't change for all components downstream

mattkanwisher:
@juliusv ok fixed the interface is no longer changed downstream. Do we have any other issues here?

./prometheus -log.format="logger:syslog?json=true&appname=bob&local=local7

log/log.go
```diff
@@ -39,10 +41,57 @@ func (f levelFlag) Set(level string) error {
...
+	if appname != "" && local != "" {
+		fmter, err := newSyslogger(appname, local, origLogger.Formatter)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
+			origLogger.Errorf("can't connect logger to syslogs: %v", err)
+		} else {
+			origLogger.Formatter = fmter
+		}
+	}
+	return nil
```
grobie: Try to reduce indentation in go.

```go
if appname == "" || local == "" {
  return nil
}

fmter, err := newSyslogger(appname, local, origLogger.Formatter)
if err != nil {
  fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
  origLogger.Errorf("can't connect logger to syslogs: %v", err)
  return nil
}

origLogger.Formatter = fmter
return nil
```

I don't really understand, why your method signature includes a `error` return value. You never check for the return value and always return `nil` anyway.

log/syslog_formatter.go
```diff
@@ -0,0 +1,84 @@
...
+		return syslog.LOG_LOCAL3
+	case "4":
+		return syslog.LOG_LOCAL4
+	case "5":
+		return syslog.LOG_LOCAL5
+	case "6":
+		return syslog.LOG_LOCAL6
+	case "7":
+		return syslog.LOG_LOCAL7
+	}
```
grobie: This seems to be easier solved by a map than by running through a switch statement every time?

```go
var localMapping = map[string]syslog.Priority{
  "0": syslog.LOG_LOCAL0,
  "1": syslog.LOG_LOCAL1,
  // ...
}

func getLocal(level string) syslog.Priority {
  p, ok := localMapping[level]
  if !ok {
    return syslog.LOG_LOCAL0
  }
  return p
}
```

mattkanwisher: The function only runs once ever

log/log.go
```diff
@@ -39,10 +41,57 @@ func (f levelFlag) Set(level string) error {
...
+		setFormatterSylog(appname, local)
+	}
+	return nil
+}
+
 func init() {
 	// In order for this flag to take effect, the user of the package must call
 	// flag.Parse() before logging anything.
 	flag.Var(levelFlag{}, "log.level", "Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].")
+	flag.Var(formatLog{}, "log.format", "If set use a syslog logger or json logging. Example: logger:syslog?appname=bob&local=local7 or logger:stdout@?json=true ")
```
grobie: So this example shows `local=local7`. The `getLocal` method in the formatter only knows about `local=7` though. Am I missing where you remove the `local` part from `local7`?

mattkanwisher: Example is wrong good catch should say local=7, updated

log/syslog_formatter.go
grobie: Nevermind, it's only used during initialization anyways.

log/syslog_formatter.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"log/syslog"
+	"os"
+
+	"github.com/Sirupsen/logrus"
+)
+
+var ceeTag = []byte("@cee:")
+
+var _ logrus.Formatter = &syslogger{}
```
grobie: What the purpose of this line?

log/syslog_formatter.go
```diff
@@ -0,0 +1,84 @@
...
+	case "6":
+		return syslog.LOG_LOCAL6
+	case "7":
+		return syslog.LOG_LOCAL7
+	}
+	return syslog.LOG_LOCAL0
+}
+
+func (s *syslogger) Format(e *logrus.Entry) ([]byte, error) {
+
```
grobie: No empty line after func declaration please.

log/syslog_formatter.go
mattkanwisher: not needed removed, this was copied from our internal logging library that also uses logrus

log/log.go
```diff
@@ -39,10 +41,59 @@ func (f levelFlag) Set(level string) error {
...
+		return setFormatterSylog(appname, local)
+	}
+	return nil
+}
+
 func init() {
 	// In order for this flag to take effect, the user of the package must call
 	// flag.Parse() before logging anything.
 	flag.Var(levelFlag{}, "log.level", "Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].")
+	flag.Var(formatLog{}, "log.format", "If set use a syslog logger or json logging. Example: logger:syslog?appname=bob&local=7 or logger:stdout@?json=true ")
```
grobie: What's the purpose of the @ sign in the stdout example?

mattkanwisher: none fixed

grobie:
:+1:

Structured logs are so much easier to deal with, I'd always prefer them. Especially as Prometheus de-facto already uses structured logging, having a well defined format makes a lot of sense.

Syslog is so common that we should support. The majority of software I've been working with only supports syslog style logging and nothing else.

log/log.go
```diff
@@ -39,10 +41,59 @@ func (f levelFlag) Set(level string) error {
...
+	}
+	if u.Scheme != "logger" {
+		return fmt.Errorf("Invalid scheme %s", u.Scheme)
+	}
+	jsonq := u.Query().Get("json")
+	if jsonq == "true" {
+		setJsonFormatter()
+	}
+
+	if u.Opaque == "syslog" {
```
grobie: What do you think about being explicit about the acceptable loggers? Might help to avoid confusion.

```go
switch u.Opague {
case "syslog":
  // ...
case "stdout":
default:
  return fmt.Error("unsupported logger", u.Opague)
}
```

log/log.go
```diff
@@ -39,10 +41,59 @@ func (f levelFlag) Set(level string) error {
...
+		appname := u.Query().Get("appname")
+		local := u.Query().Get("local")
+		return setFormatterSylog(appname, local)
+	}
+	return nil
+}
+
 func init() {
 	// In order for this flag to take effect, the user of the package must call
```
grobie: s/this flag/these flags/

log/log.go
```diff
@@ -39,10 +41,59 @@ func (f levelFlag) Set(level string) error {
...
+}
+
+// Set implements flag.Value.
+func (f formatLog) Set(format string) error {
+	u, err := url.Parse(format)
+	if err != nil {
+		return err
+	}
+	if u.Scheme != "logger" {
+		return fmt.Errorf("Invalid scheme %s", u.Scheme)
```
grobie: Please use lowercase error strings: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

log/log.go
mattkanwisher: i like the switch fixed

log/log.go
mattkanwisher: lowercase fixed

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+	}
+
+	switch u.Opaque {
+	case "syslog":
+		appname := u.Query().Get("appname")
+		local := u.Query().Get("local")
+		return setFormatterSylog(appname, local)
+	case "stdout":
+	default:
+		return fmt.Errorf("unsupported logger", u.Opaque)
```
grobie: Ah sorry, I forgot there isn't a `fmt.Error` method like the `log.Error` one. Please add the missing ` %s` to print the logger value (`go vet` should also mark this problem).

mattkanwisher: fixed, make vet doesn't cover common ill do it manually ;)

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
 
+func setFormatterSylog(appname, local string) error {
+	if appname != "" && local != "" {
+		return nil
+	}
+
+	fmter, err := newSyslogger(appname, local, origLogger.Formatter)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
+		origLogger.Errorf("can9't connect logger to syslogs: %v", err)
```
juliusv: "can9't" -> "can't"

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
 	return nil
 }
 
+func setFormatterSylog(appname, local string) error {
```
juliusv: "Sylog" -> "Syslog"

This function name (`setFormatterSyslog`) is inconsistent with the other one (`setJsonFormatter`). Maybe rename it to `setSyslogFormatter`?

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
+		origLogger.Errorf("can9't connect logger to syslogs: %v", err)
+		return err
+	}
+	origLogger.Formatter = fmter
+	return nil
+}
+
+func setJsonFormatter() {
```
juliusv: "Json" -> "JSON", see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+	}
+	return nil
+}
+
 func init() {
-	// In order for this flag to take effect, the user of the package must call
+	// In order for these flag to take effect, the user of the package must call
 	// flag.Parse() before logging anything.
 	flag.Var(levelFlag{}, "log.level", "Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].")
+	flag.Var(formatLog{}, "log.format", "If set use a syslog logger or json logging. Example: logger:syslog?appname=bob&local=7 or logger:stdout?json=true ")
```
juliusv: "json logging" -> "JSON logging"

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+	case "stdout":
+	default:
+		return fmt.Errorf("unsupported logger %s", u.Opaque)
+	}
+	return nil
+}
+
 func init() {
-	// In order for this flag to take effect, the user of the package must call
+	// In order for these flag to take effect, the user of the package must call
```
juliusv: "flag" -> "flags"

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+	}
+	return nil
+}
+
 func init() {
-	// In order for this flag to take effect, the user of the package must call
+	// In order for these flags to take effect, the user of the package must call
 	// flag.Parse() before logging anything.
 	flag.Var(levelFlag{}, "log.level", "Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].")
+	flag.Var(formatLog{}, "log.format", "If set use a syslog logger or JSON logging. Example: logger:syslog?appname=bob&local=7 or logger:stdout?json=true ")
```
juliusv: Naming consistency: if we name the above `levelFlag`, then we should also name this one `logFormatFlag` instead of `formatLog`.

mattkanwisher:
Cool updated all the feedback

log/log.go
```diff
@@ -39,10 +41,63 @@ func (f levelFlag) Set(level string) error {
...
+
+func setJSONFormatter() {
+	origLogger.Formatter = &logrus.JSONFormatter{}
+}
+
+type logFormatFlag struct{}
+
+// String implements flag.Value.
+func (f logFormatFlag) String() string {
+	return origLogger.Level.String()
```
juliusv: This now should return something different than the log level? Guess this was a copy-and-paste from the `levelFlag.String()` implementation?

See also the wrong default values in the flag help output now:

```
   -log.format "info"
      If set use a syslog logger or JSON logging. Example: 
      logger:syslog?appname=bob&local=7 or logger:stdout?json=true
  
   -log.level "info"
      Only log messages with the given severity or above. Valid levels: 
      [debug, info, warn, error, fatal].
```

mattkanwisher: so I'm just going to return the original log format uri, since thats probably the most useful

juliusv: (I guess it should return the empty string for an empty default value?)

mattkanwisher: Probably, I'm not sure how useful this really is. Normally I woudn't do custom struct flags

juliusv: Yeah, what you did is actually correct I think.

log/log.go
```diff
@@ -39,10 +41,64 @@ func (f levelFlag) Set(level string) error {
...
 
+func setSyslogFormatter(appname, local string) error {
+	if appname != "" && local != "" {
+		return nil
+	}
+
+	fmter, err := newSyslogger(appname, local, origLogger.Formatter)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
+		origLogger.Errorf("can't connect logger to syslogs: %v", err)
```
juliusv: Is this supposed to read "syslogs" instead of "syslog"?

mattkanwisher:
I think we have 5 hours in on this PR now ;) how about we merge this sucker

log/log.go
mattkanwisher: ;) fixed

log/syslog_formatter.go
```diff
@@ -0,0 +1,81 @@
...
+func newSyslogger(appname string, level string, fmter logrus.Formatter) (*syslogger, error) {
+	priority := getLocal(level)
+	out, err := syslog.New(priority, appname)
+	return &syslogger{
+		out:  out,
+		wrap: fmter,
+	}, err
+}
+
+func getLocal(level string) syslog.Priority {
```
juliusv: Would `getPriority` be a clearer name for this?

mattkanwisher: no maybe Facility would be better https://en.wikipedia.org/wiki/Syslog but its not a priority

juliusv: Well, it returns a `syslog.Priority`. The stdlib docs say:

"The Priority is a combination of the syslog facility and severity."

So I guess priority would actually be correct here...

juliusv:
Looks good to me now, but want another person to do a final :+1: and merge, since it's touching the whole ecosystem and I'm vacationing :)

log/syslog_formatter.go
grobie: This method won't ever return severity, so I don't think getPriority makes sense. +1 for getFacility.

juliusv: Ok.

log/syslog_formatter.go
```diff
@@ -0,0 +1,81 @@
...
+	case "4":
+		return syslog.LOG_LOCAL4
+	case "5":
+		return syslog.LOG_LOCAL5
+	case "6":
+		return syslog.LOG_LOCAL6
+	case "7":
+		return syslog.LOG_LOCAL7
+	}
+	return syslog.LOG_LOCAL0
```
grobie: @mattkanwisher So I was thinking about this after the introduction of the `switch Opague`. It might be a good idea to not return a default value for all unknown levels, but to return an error. That way a user will have immediate feedback what's wrong with `local=local7` instead of trying to understand why their log messages are delivered using the wrong facility. What do you think?

mattkanwisher: ok sure returning error now

grobie:
Awesome! Thanks so much for sticking with us ;)

log/log.go
```diff
@@ -39,10 +41,64 @@ func (f levelFlag) Set(level string) error {
 	return nil
 }
 
+func setSyslogFormatter(appname, facility string) error {
+	if appname != "" && facility != "" {
```
grobie: Wait, this should have been changed to be the other way around. `appname == "" || facility == ""`.

log/log.go
```diff
@@ -39,10 +41,64 @@ func (f levelFlag) Set(level string) error {
...
+		return nil
```
grobie: Might be good to return an error here instead of nil. This code path is only reached if someone set `logger:syslog`. We should tell them that appname and facility are mandatory.

If you think it makes sense, we could alternatively set facility to some default value.

log/log.go
mattkanwisher: good catch fixed

log/log.go
```diff
@@ -39,10 +41,64 @@ func (f levelFlag) Set(level string) error {
 	return nil
 }
 
+func setSyslogFormatter(appname, facility string) error {
+	if appname != "" || facility != "" {
```
grobie: This is still wrong. If either of them is set we'll return early here. It needs to be `==` in both comparisons.

A test might make sense at this point ;)

fabxc:
Chiming in late – I derived this package a few weeks back exactly because I wanted more structured logging. While changing a structured log message is more explicitly breaking, it is still an order of magnitude better than parsing unstructured logs.

In the end we have to decide whether our logging output is ruled by our semantic versions or not – I doubt that it makes sense.

Thanks for writing this and everyone else for reviewing :+1:

log/log.go
```diff
@@ -39,10 +41,64 @@ func (f levelFlag) Set(level string) error {
...
+	if jsonq == "true" {
+		setJSONFormatter()
+	}
+
+	switch u.Opaque {
+	case "syslog":
+		appname := u.Query().Get("appname")
+		facility := u.Query().Get("local")
+		return setSyslogFormatter(appname, facility)
+	case "stdout":
```
fabxc: logrus defaults to `stderr` though. So People will provide stdout, and will get stderr, while providing stderr will fail.

https://github.com/Sirupsen/logrus/commit/f8f08842cce7ed05c296430852f60f287458c529

mattkanwisher: nice catch fixed

fabxc:
Thanks a bunch! :+1:

PR #18, opened 2015-10-26
Import ww/goautoneg to internal/

### LINKED ISSUE #17: Dependency on bitbucket.org/ww/goautoneg is the only mercurial dependency
Hi! In various places, I don't have mercurial available (for example, CoreOS). I've also taken to using git submodules for vendoring. Out of the 7 external packages required to import `prometheus/client_golang`, the only non-git dependency is bitbucket.org/ww/goautoneg. See https://godoc.org/github.com/prometheus/client_golang/prometheus?import-graph&hide=2.

`ww/goautoneg` has not had a commit since 2012. It's a very small amount of code: https://bitbucket.org/ww/goautoneg/src/75cd24fc2f2c2a2088577d12123ddee5f54e0675/autoneg.go?at=default&fileviewer=file-view-default

Any chance the source could be imported wholesale, and live where it is needed? (perhaps to `/internal/autoneg`, for example).

---

As a side note in the project I'm using this, there are 26 external repositories. All of which live in git repositories and happily live in the git submodule ecosystem. This is the first package in more than a year of development on this project where I've hit a non-git dependency.

pwaller:
Copies the code and readme wholesale to `internal/`, eliminating
the only non-git import.

Please test this works for you before merging! :)

Fixes #17.

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -18,9 +18,9 @@ import (
 	"io"
 	"net/http"
 
-	"bitbucket.org/ww/goautoneg"
 	"github.com/golang/protobuf/proto"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
+	"github.com/prometheus/common/internal/bitbucket.org/ww/goautoneg"
 
 	dto "github.com/prometheus/client_model/go"
 )
--- a/internal/bitbucket.org/ww/goautoneg/README.txt
+++ b/internal/bitbucket.org/ww/goautoneg/README.txt
@@ -1,0 +1,67 @@
+PACKAGE
+
+package goautoneg
+import "bitbucket.org/ww/goautoneg"
+
+HTTP Content-Type Autonegotiation.
+
+The functions in this package implement the behaviour specified in
+http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
+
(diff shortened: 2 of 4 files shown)


fabxc:
The new go1.5 vendoring which we are using should support library-level vendoring. From my understanding `internal/` is for self-written packages that are just not available for import outside of the libraries scope.

So while a rewrite of the logic would fit into `internal/`. I wonder whether this would be candidate for `common/vendor/`. Not sure whether this would solve your problem though, since it is also already vendored by applications importing `common/expfmt`.

pwaller:
Yes, I use the vendoring mechanism in lots of places, too. It's generally my opinion that things used as libraries should avoid vendor if possible. In this PR it is my intent was really as if the code is being brought in-tree, removing the external dependency, in a sense. I based that decision partly on how long it has been since the dependency in question was last changed. It also ensures that consumers not using `GO15VENDOREXPERIMENT=1` will work, too. Happy to move it over to `vendor/` though if you think that's more appropriate.

fabxc:
Given that the vendoring is still experimental you have a point here for a library. Let's roll with this for now, licenses and all are included so this should be fine.

fabxc:
:+1: Thanks

pwaller:
:+1:

PR #20, opened 2015-11-12
The FNV trick again

alicebob:
@beorn7: Same idea as in client_golang, although here it only saves a sync.Pool.

There are two commits: the first one removes the sync.Pool, and the second one moves the FNV code to an external package. If you don't like the external package just ignore that commit, the first commit is standalone (or fork it into the prometheus namespace). Otherwise we'll also change client_golang to use it.

(with help from @robx)


fabxc:
Thanks! :+1: in general.

I do not feel to strongly about having an additional external dependency. I think that they are a non-issue given that a language's vendoring is designed properly. For Go I consider this to be the case at this point (1.6 will polish the last bits).

Generally we've been trying to reduce the dependencies that are transitively pulled in through the client library. I do not deem this necessary but @brian-brazil has a point that some people might be driven away by it.
So given the amount of code, for now it's probably best to inline it via the first commit.

beorn7:
Given the shortness of the code and the issues that still exist around dependencies of libraries and their vendoring, I'd vote for inlining here.

Thanks for your efforts.

alicebob:
I've reverted the external package commit and added the license header to fnv.go.

beorn7:
Thanks here, too. :+1:

PR #21, opened 2015-11-26
Add generatorURL to Alert struct

fabxc:
@brian-brazil 

Wondering how to handle this field though. Should it really always be an URL. I don't think all clients necessarily fulfill that.

### DIFF
--- a/model/alert.go
+++ b/model/alert.go
@@ -35,8 +35,9 @@ type Alert struct {
 	Annotations LabelSet `json:"annotations"`
 
 	// The known time range for this alert. Both ends are optional.
-	StartsAt time.Time `json:"startsAt,omitempty"`
-	EndsAt   time.Time `json:"endsAt,omitempty"`
+	StartsAt     time.Time `json:"startsAt,omitempty"`
+	EndsAt       time.Time `json:"endsAt,omitempty"`
+	GeneratorURL string    `json:"generatorURL"`
 }
 
 // Name returns the name of the alert. It is equivalent to the "alertname" label.


model/alert.go
```diff
@@ -35,8 +35,9 @@ type Alert struct {
 	Annotations LabelSet `json:"annotations"`
 
 	// The known time range for this alert. Both ends are optional.
-	StartsAt time.Time `json:"startsAt,omitempty"`
-	EndsAt   time.Time `json:"endsAt,omitempty"`
+	StartsAt     time.Time `json:"startsAt,omitempty"`
+	EndsAt       time.Time `json:"endsAt,omitempty"`
+	GeneratorURL string    `json:"generatorURL"`
```
brian-brazil: Standard JSON naming would be generator_url, starts_at etc. as it follows Javascript language conventions.

brian-brazil:
> Wondering how to handle this field though. Should it really always be an URL. I don't think all clients necessarily fulfill that.

There's two purposes to the field.

The most important is so that a user has something that'll let them block alerts from a given client if it breaks badly (e.g. machine is inaccessible but still spewing alerts). For that case a defaulting it to `http://client_ip/` handles any non-Prometheus clients that haven't implemented it, bar in cases where there's proxies or multi-IP NAT in play.

Where users will more commonly hit it is in the UI and long-form notifications where it'll link back to the relevant alert on Prometheus.  That's the external url on the Prometheus side so all that is fine. For non-Prometheus clients, the above fallback will at least get you to the machine in question.

Given the utility of the 2nd use case, I think a URL makes sense as it's only really clients that don't do HTTP at all where it's likely to be a problem. We could split the two uses into two separate fields, but I don't think it's worth the hassle as most implementers of clients would only see the utility of the 2nd case with the 1st being a nice to have rather than the other way around.

model/alert.go
fabxc: Uhm, we had this discussion a while back when specifying the Prometheus API. I remember that I was going for the underscore based naming and you arguing for camelCase. We then decided we will consistently do camelCase.

fabxc:
Yes, for the first use case it just has to be some identifier. URLs can be turned into HTML links on the fly in the UI (as happening for annotations for example). So I would argue for just naming it `generatorID`.

model/alert.go
brian-brazil: Sorry, I got confused with all the different naming conventions I need to remember. What you have is correct.

brian-brazil:
> URLs can be turned into HTML links on the fly in the UI (as happening for annotations for example). So I would argue for just naming it generatorID.

That's fine for the 1st case, but for the 2nd I want to be a URL for the UI as it'll be a much nicer experience for the user. That implies that it needs to be explicit that we're expecting a URL and going to use it as a link target.

We can't make any assumptions about the content of annotations, as they're opaque blobs to the alertmanager.

fabxc:
What I mean is, we can use javascript to detect URLs in content and turn them into HTML links.

brian-brazil:
That'd make sense for a big block of text, but given the main thing we're going to use this for is a link target I'd say it's simpler to just request one from the user up front.

fabxc:
Generally I don't mind too much. Probably having `URL` in the name also encourages short-lived clients (alert at the end of batch job or so) to send something meaningful like an URL to a wiki page.

brian-brazil:
:+1: 

> Probably having URL in the name also encourages short-lived clients (alert at the end of batch job or so) to send something meaningful like an URL to a wiki page.

I don't think that's something we should be encouraging. Alertmanager clients should be long-lived and continue to send alerts until the issue is resolved, we can't guarantee reliability otherwise. This sounds like more a case for the push gateway.

For the first use case we want something that is the IP or dnsname of the batch job, not any wiki page.

fabxc:
I don't want to make assumptions about usage of other components in our eco-system, even if it would be a better option.

I can imagine this very well – some batch jobs with one-time alerts fail over the course of the day. Once a day post a report to a webhook. Just one possibility. We have persistence now. If that works as it should, there should be no issue with (successfully) sending an alert just once (modulo our hard-coded auto-resolve time at the moment).

brian-brazil:
There's a fair few assumptions built in there, such as the request to the AM not failing and the setup of the notification intervals permitting a notification to be sent before the alert expires. It also offers no way to stop the alert on the part of the client.

This sounds more like an event logging use case, and thus something that we should be wary of supporting in favour of the metric-based alerting that we encourage.

fabxc:
All fair points. Just saying people might come up with usages we may not anticipate.

brian-brazil:
Of definitely, I expect users to try things like that and more.

PR #22, opened 2015-12-01
Validate label names

fabxc:
@beorn7

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -19,6 +19,7 @@ import (
 	"math"
 	"mime"
 	"net/http"
+	"unicode/utf8"
 
 	dto "github.com/prometheus/client_model/go"
 
@@ -110,6 +111,10 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 	if err != nil {
 		return err
 	}
+	// MatchString already returns false if !utf8.ValidString(v.GetName()).
+	if !model.MetricNameRE.MatchString(v.GetName()) {
+		return fmt.Errorf("invalid metric name %q", v.GetName())
+	}
 	for _, m := range v.GetMetric() {
 		if m == nil {
 			continue
@@ -118,6 +123,10 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 			if l == nil {
 				continue
 			}
+			if !utf8.ValidString(l.GetValue()) {
+				return fmt.Errorf("invalid label value %q", l.GetValue())
+			}
+			// MatchString already returns false if !utf8.ValidString(l.GetName()).
 			if !model.LabelNameRE.MatchString(l.GetName()) {
 				return fmt.Errorf("invalid label name %q", l.GetName())
 			}
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -19,7 +19,6 @@ import (
 	"math"
 	"mime"
 	"net/http"
-	"unicode/utf8"
 
 	dto "github.com/prometheus/client_model/go"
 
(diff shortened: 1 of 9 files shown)


fabxc:
Will throw some tests in there later.

expfmt/decode.go
```diff
@@ -107,7 +107,23 @@ type protoDecoder struct {
...
+		if m == nil {
+			continue
+		}
+		for _, l := range m.GetLabel() {
+			if l == nil {
+				continue
+			}
+			if !model.LabelNameRE.MatchString(l.GetName()) {
+				return fmt.Errorf("invalid label name %q", l.GetName())
+			}
```
beorn7: I'm 80% sure that protobuf can transport invalid UTF-8 strings (while JSON can't). So (just yesterday) I ran into a library function to check if a string is a valid UTF-8 string (on which we depend in our hashing function).

So I'd add
`if !utf8.ValidString(l.GetValue()) { ... } `

beorn7:
Otherwise :+1:

expfmt/decode.go
fabxc: That would have to be applied to metric names and label values too then, right?

beorn7: The regexp match will also catch invalid strings (as not-matching, that is). I tried that out.

fabxc:
Added unicode validation for proto strings.

expfmt/decode.go
```diff
@@ -107,7 +108,30 @@ type protoDecoder struct {
 // Decode implements the Decoder interface.
 func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 	_, err := pbutil.ReadDelimited(d.r, v)
-	return err
+	if err != nil {
+		return err
+	}
+	if !utf8.ValidString(v.GetName()) {
```
beorn7: We have a regexp somewhere for metric names. We should check against the regexp here (which will include checking for valid UTF-8).

fabxc: Not in common/model at least. Until now.

fabxc:
Performance A/B test:

Fresh servers, 32 cores, 64GB RAM, ~2800 targets.

Without checks:
 ~524,000 samples/sec
load avg ~12.5

With checks:
 ~514,000 samples/sec
load avg ~18.8

Certainly a significant CPU impact. Slight reduction in samples/sec.
We will see if that changes over time. Probably the samples/sec are not sustainable anyway.

fabxc:
Updated. A/B test running.

model/metric.go
```diff
@@ -79,3 +83,19 @@ func (m Metric) Fingerprint() Fingerprint {
...
+// IsValidMetricName returns true iff name matches the pattern of MetricNameRE.
+func IsValidMetricName(n LabelValue) bool {
+	if len(n) == 0 {
+		return false
+	}
+	if !((n[0] >= 'a' && n[0] <= 'z') || (n[0] >= 'A' && n[0] <= 'Z') || n[0] == '_' || n[0] == ':') {
+		return false
+	}
+	for _, b := range n[1:] {
+		if !((b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || b == ':' || (b >= '0' && b <= '9')) {
```
beorn7: I think the previous `if` block can be easily avoided by taking the index in the for loop and loop over the whole string (`for i, b := range n {`) and then expand the condition to `... || (b >= '0' && b <= '9' && i>0))`.

Same above.

fabxc:
This version has no notable impact on CPU load or samples/sec. Seems good to go.

model/metric.go
fabxc: Less code but more checks per subsequent iterations. Probably doesn't matter either way.

beorn7: The "more checks" only kick in for numbers, so pretty rarely. With the amount of replicated code (and potential of unintentional deviations between L92 and L96), I'd vote pretty clearly for the one line version.

PR #23, opened 2015-12-04
Validate protobuf label names on decode.

jpittis:
Based on prometheus/prometheus/issues/1232, we are not validating protobuf label names on decode.

This patch provides a [test](https://github.com/jpittis/common/blob/5fde791b57d7901fdf17b3f0439769467329949f/expfmt/decode_test.go#L297-L314) to confirm bad labels normally decode without error and adds [validation](https://github.com/jpittis/common/blob/5fde791b57d7901fdf17b3f0439769467329949f/expfmt/decode.go#L112-L120).

For the test, I stole a passing test case and mutated the label name to be incorrect.

I'm unsure if this is the best place to perform validation. I am also unsure that the error being returned from `Decode` is clear enough.

ping: @beorn7 @fabxc

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -19,6 +19,7 @@ import (
 	"math"
 	"mime"
 	"net/http"
+	"strings"
 
 	dto "github.com/prometheus/client_model/go"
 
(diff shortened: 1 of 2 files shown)


jpittis:
I just noticed #22 is already opened to fix this.

beorn7:
Sorry for the duplicate work...

PR #25, opened 2015-12-10
Remove the last occurrence of `utf8.ValidString`

beorn7:
It's all redirected to model.LabelValue.IsValid now.  We use string as
type for Matcher.Value because it could be a regexp rather than a
LabelValue.

@fabxc


fabxc:
:+1:

PR #26, opened 2016-01-04
Limit syslogger to unix systems

fabxc:
Currently packages importing `common/log` will not compile on Windows and other systems due to the missing build constraints for the syslog formatter.

@brian-brazil

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -41,23 +41,8 @@ func (f levelFlag) Set(level string) error {
 	return nil
 }
 
-func setSyslogFormatter(appname, local string) error {
-	if appname == "" {
-		return fmt.Errorf("missing appname parameter")
-	}
-	if local == "" {
-		return fmt.Errorf("missing local parameter")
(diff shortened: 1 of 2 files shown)


log/log.go
```diff
@@ -41,23 +41,7 @@ func (f levelFlag) Set(level string) error {
...
-	fmter, err := newSyslogger(appname, local, origLogger.Formatter)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "error creating syslog formatter: %v\n", err)
-		origLogger.Errorf("can't connect logger to syslog: %v", err)
-		return err
-	}
-	origLogger.Formatter = fmter
-	return nil
-}
+var setSyslogFormatter func(string, string) error
```
brian-brazil: A comment on why this is done this way would be useful

brian-brazil:
:+1:

juliusv:
:+1:

PR #29, opened 2016-01-29
Extend Duration to years and weeks

beorn7:
I'm currently working on unifying all Duration parsing. While it's
arguably rarely useful to use y or w, I'm pretty sure that we have
users doing it (and we have separate parsing code still in place
supporting it (and I'm working on removing that other parsing code)).

I'd hate to break people's expressions and configs just because we
think 1w is too long for a range...

@fabxc

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -163,10 +163,10 @@ func (t *Time) UnmarshalJSON(b []byte) error {
 // This type should not propagate beyond the scope of input/output processing.
 type Duration time.Duration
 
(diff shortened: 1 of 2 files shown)


grobie:
:+1:

fabxc:
Yes, when working on that bit recently I realized that we have duplicated parsing code etc.
Thanks for picking this up.

:+1:

juliusv:
:+1:

PR #30, opened 2016-02-11
Return meaningful error instead of EOF during text parsing

beorn7:
So far, we are just passing through the io.EOF error created when
trying to read from the input stream. That io.EOF error doesn't tell
the caller the position where it has happened, and worse, it could
easily misunderstood as a signal for the legitimate end of the input
stream.

@fabxc FIxes https://github.com/prometheus/prometheus/issues/1135


fabxc:
:+1:

beorn7:
Thanks. Of course, now we need to update vendoring at some point... ;)

PR #35, opened 2016-03-18
Empty vector is the same as a non-empty vector without that label.

brian-brazil:
This is the cause of https://github.com/prometheus/prometheus/issues/1489

### DIFF
--- a/model/signature.go
+++ b/model/signature.go
@@ -98,7 +98,7 @@ func labelSetToFastFingerprint(ls LabelSet) Fingerprint {
 // specified LabelNames into the signature calculation. The labels passed in
 // will be sorted by this function.
 func SignatureForLabels(m Metric, labels ...LabelName) uint64 {
-	if len(m) == 0 || len(labels) == 0 {
+	if len(labels) == 0 {
 		return emptyLabelSignature
 	}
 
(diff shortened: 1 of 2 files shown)


beorn7:
:+1:

I believe that's the right fix to be consistent with the current situation.

However, looking at it, I was thinking about our rule that a label with an empty value should count as a label not present at all. Shouldn't we bake that into our signature generation and simply ignore labels with an empty value altogether?

fabxc:
Yes

brian-brazil:
Go map semantics means we'll get an empty string if the label isn't present, which is what we want.

beorn7:
But then we still take the label _name_ into account. So for the hash value, an existing label name with an empty value is different from a label that isn't there at all.

brian-brazil:
We're passing in that list of label names, so it makes no difference for this function.

beorn7:
Indeed, it doesn't make a difference for this function in particular. Therefore my :+1: and my statement above. This PR is good for merging.

However, I was thinking about the more general problem. Should a metric with an unset label value make it into the storage, it will have a different fingerprint as a metric with that label not set at all. We throw away labels with emtpy value upon ingestion, but I think we should make the treatment consistent in fingerprint calculation. I'll file a separate issue for that.

PR #38, opened 2016-05-01
Add version informations as a common package

sdurrheimer:
Ref. prometheus/node_exporter#236

Usage as mentionned by @grobie 

```go
import (
	"flag"
	"fmt"
	"os"

	"github.com/prometheus/common/log"
	"github.com/prometheus/common/version"
)

func init() {
	prometheus.MustRegister(version.NewCollector("<program>"))
}

func main() {
	var showVersion = flag.Bool("version", false, "Print version information.")
	flat.Parse()
	
  	if *showVersion {
		fmt.Fprintln(os.Stdout, version.Print("<program>"))
		os.exit(0)
 	}

  	log.Infoln("Starting <program>", version.Info())
  	log.Infoln("Build context", version.BuildContext())
}
```

I had a [second implementation](https://gist.github.com/sdurrheimer/2363d7832e7a79091209fec16e4ca120), but it was a bit overkill for a single metric.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ sudo: false
 language: go
 go:
   - 1.4.3
-  - 1.5.1
+  - 1.5.4
+  - 1.6.2
   - tip
--- a/version/info.go
+++ b/version/info.go
@@ -1,0 +1,89 @@
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package version
+
+import (
+	"bytes"
+	"fmt"
+	"runtime"
+	"strings"
+	"text/template"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Build information. Populated at build-time.
+var (
+	Version   string
+	Revision  string
+	Branch    string
+	BuildUser string
+	BuildDate string
+	GoVersion = runtime.Version()
+)
+
+// NewCollector returns a collector which exports metrics about current version information.
+func NewCollector(program string) *prometheus.GaugeVec {
+	buildInfo := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: program,
+			Name:      "build_info",
+			Help: fmt.Sprintf(
+				"A metric with a constant '1' value labeled by version, revision, branch, and goversion from which %s was built.",
+				program,
+			),
+		},
+		[]string{"version", "revision", "branch", "goversion"},
+	)
+	buildInfo.WithLabelValues(Version, Revision, Branch, GoVersion).Set(1)
+	return buildInfo
+}
+
+// versionInfoTmpl contains the template used by Info.
+var versionInfoTmpl = `
+{{.program}}, version {{.version}} (branch: {{.branch}}, revision: {{.revision}})
+  build user:       {{.buildUser}}
+  build date:       {{.buildDate}}
+  go version:       {{.goVersion}}
+`
+
+// Print returns version information.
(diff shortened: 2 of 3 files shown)


version/info.go
```diff
@@ -0,0 +1,92 @@
...
+
+	var buf bytes.Buffer
+	if err := t.ExecuteTemplate(&buf, "version", Map); err != nil {
+		panic(err)
+	}
+	return strings.TrimSpace(buf.String())
+}
+
+// VersionLog returns version, branch and revision information.
+func VersionLog() string {
```
grobie: Guess log might be a confusing name. Suggestions? `VersionInfo`?

version/info.go
```diff
@@ -0,0 +1,92 @@
...
+	return fmt.Sprintf("(version=%s, branch=%s, revision=%s)", Version, Branch, Revision)
+}
+
+// BuildLog returns goVersion, buildUser and buildDate information.
+func BuildLog() string {
```
grobie: `BuildContext`?

version/info.go
grobie: `version.Details()`?

version/info.go
```diff
@@ -0,0 +1,92 @@
...
+	Version   string
+	Revision  string
+	Branch    string
+	BuildUser string
+	BuildDate string
+	GoVersion = runtime.Version()
+)
+
+// Map provides the iterable version information.
+var Map = map[string]string{
```
grobie: I'm unsure about making this part of the interface. It doesn't provide a clear contract and all the variables are already available. I'd move that map into the Info() method as local variable.

version/info.go
sdurrheimer: What about renaming :
- `version.Info()` to `version.Print()` or `version.PrintString()`.
- `version.VersionLog()` to `version.Info()` or `version.InfoString()`
- `version.BuildLog()` to `version.BuildContext()` or `version.BuildString()`

version/info.go
sdurrheimer: If we are sure that this iterable interface is never gonna be used elsewhere, I guess we can move it to `version.Info()`.

version/info.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Info returns version information.
+func Info(program string) string {
+	Map["program"] = program
+
+	t := template.Must(template.New("version").Parse(versionInfoTmpl))
+
+	var buf bytes.Buffer
+	if err := t.ExecuteTemplate(&buf, "version", Map); err != nil {
+		panic(err)
```
sdurrheimer: I'm considering changing the `panic` here (https://github.com/golang/go/wiki/CodeReviewComments#dont-panic)

.travis.yml
```diff
@@ -3,5 +3,5 @@ sudo: false
 language: go
 go:
   - 1.4.3
-  - 1.5.1
+  - 1.5.4
```
sdurrheimer: Given that this repository is about libs, should we continue to test against `go 1.4.3` and also add `go 1.6.2` ?

version/info.go
grobie: Sounds good to me. I'd go with the first version, the return type should make clear it's a string.

version/info.go
grobie: There is one usage outside of the version package using the Map in prometheus/prometheus. Though, they expect specific keys to be available. As the interface can't be ensured (the map could be empty and still pass the compilation), I think the version package's only interface should be the package variables. Alternatively we could provide a struct.

version/info.go
grobie: I think this is fine here. The template input is controlled by the same package, so any error would be a programming mistake and should be catched during development.

.travis.yml
grobie: SGTM

version/info.go
sdurrheimer:
```bash
$ grep -nr 'version.Map' .
./cmd/prometheus/main.go:236:	if err := t.ExecuteTemplate(&buf, "version", version.Map); err != nil {
./cmd/promtool/main.go:199:	if err := tmpl.ExecuteTemplate(&buf, "version", version.Map); err != nil {
./storage/local/storagetool/main.go:54:	if err := tmpl.ExecuteTemplate(&buf, "version", version.Map); err != nil {
./web/web.go:322:		Info:   version.Map,
./web/web.go:328:	if err := dec.Encode(version.Map); err != nil {
```
The first 3 matches will be replaced by `version.Print()` normally.

grobie:
:+1: very nice!

PR #40, opened 2016-05-19
Make SampleValue.Equal return true for NaNs.

beorn7:
This is different from pure floats, but needed in identifying no-ops
in the storage (PR will follow).

Finally a use for the Equal method (which was never used before).

Even the other Equal methods fixed here are used nowhere. But this change is actually important for consistency (because it is doing address comparison first, which would also yield true for two NaN values).

@fabxc

### DIFF
--- a/model/value.go
+++ b/model/value.go
@@ -16,6 +16,7 @@ package model
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


brian-brazil:
Are we using this anywhere that'd affect PromQL semantics?

beorn7:
As said above, these are used nowhere so far.

brian-brazil:
:+1:

PR #41, opened 2016-05-26
add ErrorLogger

alileza:
This function can be use to connect HTTP Server ErrorLog with this common/log.

Here's one of the issue 
https://github.com/prometheus/prometheus/issues/1651

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -16,6 +16,7 @@ package log
 import (
 	"flag"
 	"fmt"
+	"log"
 	"net/url"
 	"os"
 	"runtime"
@@ -301,4 +302,17 @@ func Fatalln(args ...interface{}) {
 // Fatalf logs a message at level Fatal on the standard logger.
 func Fatalf(format string, args ...interface{}) {
 	baseLogger.sourced().Fatalf(format, args...)
+}
+
+type errorLogWriter struct{}
+
+func (errorLogWriter) Write(b []byte) (int, error) {
+	baseLogger.sourced().Error(string(b))
+	return len(b), nil
+}
+
+// NewErrorLogger returns a log.Logger that is meant to be used
+// in the ErrorLog field of an http.Server to log HTTP server errors.
+func NewErrorLogger() *log.Logger {
+	return log.New(&errorLogWriter{}, "", 0)
 }


log/log.go
```diff
@@ -302,3 +303,15 @@ func Fatalln(args ...interface{}) {
...
+
+type errorLogWriter struct{}
+
+func (errorLogWriter) Write(b []byte) (int, error) {
+	baseLogger.sourced().Error(string(b))
+	return len(b), nil
+}
+
+// ErrorLogger return log.Logger to be used as ErrorLog to catch HTTP server errors.
+func ErrorLogger() *log.Logger {
```
grobie: Sorry for all the late comments. In order to be consistent with the golang naming, let's call it `NewErrorLogger`. The comment  needs to be updated as well then.

log/log.go
```diff
@@ -302,3 +303,15 @@ func Fatalln(args ...interface{}) {
...
 }
+
+type errorLogWriter struct{}
+
+func (errorLogWriter) Write(b []byte) (int, error) {
+	baseLogger.sourced().Error(string(b))
+	return len(b), nil
+}
+
+// ErrorLogger return log.Logger to be used as ErrorLog to catch HTTP server errors.
```
grobie: s/return/returns

grobie:
:+1: Please squash your commits to a single one.

log/log.go
alileza: what do you mean about this comment?

alileza:
@grobie updated !

log/log.go
```diff
@@ -302,3 +303,15 @@ func Fatalln(args ...interface{}) {
...
 }
+
+type errorLogWriter struct{}
+
+func (errorLogWriter) Write(b []byte) (int, error) {
+	baseLogger.sourced().Error(string(b))
+	return len(b), nil
+}
+
+// NewErrorLogger return log.Logger to be used as ErrorLog to catch HTTP server errors.
```
juliusv: "return" -> "returns a"

How about: "NewErrorLogger returns a log.Logger that is meant to be used in the ErrorLog field of an http.Server to log HTTP server errors."

juliusv:
:+1: otherwise

log/log.go
alileza: ah, alright !

alileza:
@juliusv changes done !

juliusv:
:+1: Thank you! :)

PR #42, opened 2016-05-27
Ignore UTF8 BOMs from funny windows clients.

clavoie-acquisio:
This change is [Reviewable](https://reviewable.io/reviews/prometheus/common/42)

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -25,6 +25,7 @@ import (
 	dto "github.com/prometheus/client_model/go"
 
 	"github.com/golang/protobuf/proto"
+        "github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 )
 
@@ -125,6 +126,15 @@ func (p *TextParser) reset(in io.Reader) {
 	} else {
 		p.buf.Reset(in)
 	}
+
+	// Some clients like to throw in UTF-8 BOM chars
+	maybeBOM, err := p.buf.Peek(3)
+	if len(maybeBOM) == 3 && maybeBOM[0] == 0xEF && maybeBOM[1] == 0xBB && maybeBOM[2] == 0xBF && err == nil {
+		// The common case dealing with funny Windows clients
+		log.Debug("Discarding BOM")
+		p.buf.Discard(3)
+	}
+
 	p.err = nil
 	p.lineCount = 0
 	if p.summaries == nil || len(p.summaries) > 0 {


grobie:
Cool! Can you add a test please.

clavoie-acquisio:
With pleasure. I'm just wondering how you'd like that done -- adding random
BOMs to the existing test suite feels awkward: they're effectively
invisible, or downright disallowed in the middle of files (wikipedia says
"BOM use is optional, and, if used, should appear at the start of the text
stream.")

Happy to do it... how?

grobie:
Would it be enough to create a byte array with the bom bytes at the
beginning and then following a normal textfile line?

juliusv:
Given https://github.com/prometheus/prometheus/issues/1679#issuecomment-222252507, should we allow BOMs? I don't care either way, but it seems the recommendation is not to.

grobie:
My first reaction was to follow the robustness principle and be forgiving on the input, I do agree with that recommendation though. BOM has always been an ugly wart, I'm surprised it's still exists. Or maybe not ...

@clavoie-acquisio What client is that which sends the BOM? What client library is being used which sets that? What prevents you from not removing the BOM header instead?

beorn7:
Given https://tools.ietf.org/html/rfc3629 , I'm inclined to not allow BOMs, but explicitly add to the format specification that BOMs are not allowed.

However, if this is something that is a pain to get rid of on the side of the generator, we might go for being  "forgiving on the input".

BTW: The test only fail on Go 1.4 which doesn't compile our dependencies anymore. We should remove it from the test.

clavoie-acquisio:
Yeah, I'm fond of allowing BOMs for two big reasons:

* Valid prometheus text format streams cannot be confused for BOMs --
either '#' or [a-zA-Z:_]. So there's no chance of confusion with other UTF8
characters.

* A lot of text tools (especially on Windows) have basically forgotten that
ASCII exists. Quoth wikipedia again: "Microsoft
<https://en.wikipedia.org/wiki/Microsoft> compilers[10]
<https://en.wikipedia.org/wiki/Byte_order_mark#cite_note-10> and
interpreters, and many pieces of software on Microsoft Windows
<https://en.wikipedia.org/wiki/Microsoft_Windows> such as Notepad
<javascript:void(0)> treat the BOM as a required magic number
<javascript:void(0)> rather than use heuristics. These tools add a BOM when
saving text as UTF-8, and cannot interpret UTF-8 unless the BOM is present,
or the file contains only ASCII bytes."

I could also make this configurable on a per-target basis, and not
'pollute' the parser's purity on sane scrape targets, though my preference
is to just assume that a "text format" is what it is in the modern world:
UTF-8.

grobie:
Your reference to the modern world is exactly the reason why we have the feeling it should not be supported. A byte order is completely meaningless in UTF-8 encoding and one should expect that current software should not need such a wart anymore to handle UTF-8 correctly. As the recommendation is to not set it, I think we should have good arguments why it's necessary to support it anyways.

Do you save the textfiles with your editor? Can you just disable the BOM header in your editor? If some client library writes out the BOM header, would it be possible to change that?

clavoie-acquisio:
On Sat, May 28, 2016 at 5:31 PM, Tobias Schmidt <notifications@github.com>
wrote:

> Your reference to the modern world is exactly the reason why we have the
> feeling it should not be supported. A byte order is completely meaningless
> in UTF-8 encoding and one should expect that current software should not
> need such a wart anymore to handle UTF-8 correctly. As the recommendation
> is to not set it, I think we should have good arguments why it's necessary
> to support it anyways.
>
Because as mentioned, it doesn't hurt other clients, and most (all?) modern
Windows text processing tools require it, and won't allow disabling it (as
far as I can tell).

> Do you save the textfiles with your editor? Can you just disable the BOM
> header in your editor? If some client library writes out the BOM header,
> would it be possible to change that?
>
I'm working on fixing the libraries I can (with a few PR sent on their
respective projects), but the use case that I won't be able to fix is
indeed someone writing text files and not being able to disable BOM headers.

Christian Lavoie

>

beorn7:
Given that the text format is our most lenient fallback (e.g. if no content-type header is present), where we just try to parse anything as text format, I tend to allow BOMs, but treat them by ignoring. We should also add that to the spec, as in "BOMs are tolerated but completely ignored, i.e. the content is always considered UTF-8, no matter what the BOMs say."

And BTW: https://github.com/prometheus/common/pull/43 will fix the test.

grobie:
Sounds like it's not easy to get rid of BOM headers in Windows environments, so I agree that we should handle it.

Please add a test which verifies the correct handling. The format specification @beorn7 suggested needs to be added in the prometheus/doc repository.

expfmt/text_parse.go
```diff
@@ -125,6 +126,15 @@ func (p *TextParser) reset(in io.Reader) {
 	} else {
 		p.buf.Reset(in)
 	}
+
+	// Some clients like to throw in UTF-8 BOM chars
+	maybeBOM, err := p.buf.Peek(3)
+	if len(maybeBOM) == 3 && maybeBOM[0] == 0xEF && maybeBOM[1] == 0xBB && maybeBOM[2] == 0xBF && err == nil {
+		// The common case dealing with funny Windows clients
+		log.Debug("Discarding BOM")
```
beorn7: We don't use any logging anywhere in this package. It might come as a surprise to the (possibly very diverse) users, and it's pulling in another dependency. I think it's fine to silently swallow the BOMs.

beorn7:
Since there don't seem to be no further objections, let's go with what @grobie said:
@clavoie-acquisio please add a test as suggested. I can take care of documenting, if you don't beat me to it.

expfmt/text_parse.go
clavoie-acquisio: Will do.

beorn7:
@clavoie-acquisio Are you still on this?

clavoie-acquisio:
Yup, got distracted with deployment of prometheus. Should be able to update
the PR today or tomorrow. Touch wood.

fabxc:
ping @clavoie-acquisio :)

juliusv:
Another ping @clavoie-acquisio

judwhite:
@clavoie-acquisio @fabxc @juliusv I'm happy to pick this up and add some tests if there's still interest.

grobie:
Please go for it Jud!

brian-brazil:
Potentially relevant: https://github.com/dimchansky/utfbom

brian-brazil:
The current decision is that we won't be allowing this, as OpenMetrics forbids it.

PR #47, opened 2016-07-20
Allow logging flags to be added to custom flag set

beorn7:
First step in solving https://github.com/prometheus/prometheus/issues/1826

Need to merge this first, then update vendoring, and then send you another PR for prometheus/prometheus.

@juliusv @fabxc


juliusv:
:+1:

PR #48, opened 2016-07-25
Use strings.Replacer instead of hand-rolled escapeString

lmb:
This reduces run-time (since replacement is done via table lookup) and memory consumption (since strings that do not need replacing are returned as they are).

    name      old time/op    new time/op    delta
    Create-4    66.9µs ± 3%    55.0µs ±12%  -17.82%  (p=0.008 n=5+5)

    name      old alloc/op   new alloc/op   delta
    Create-4    19.2kB ± 0%    13.7kB ± 0%  -28.44%  (p=0.008 n=5+5)

    name      old allocs/op  new allocs/op  delta
    Create-4       617 ± 0%       498 ± 0%  -19.29%  (p=0.008 n=5+5)

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -14,7 +14,6 @@
 package expfmt
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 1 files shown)


expfmt/text_create.go
```diff
@@ -285,21 +284,18 @@ func labelPairsToText(
...
-		case c == '\\':
-			result.WriteString(`\\`)
-		case includeDoubleQuote && c == '"':
-			result.WriteString(`\"`)
-		case c == '\n':
-			result.WriteString(`\n`)
-		default:
-			result.WriteRune(c)
-		}
+	r := escapeReplacer
```
brian-brazil: Do we need this variable?

fabxc: We can just directly call `escapeReplacer.Replace(v)` etc.

fabxc:
Thanks, should be good after comment.

lmb:
Yup, done.

fabxc:
Thank you 👍

PR #52, opened 2016-08-29
Convert exposition format to proto3.

juliusv:


juliusv:
The CI fails because this depends on https://github.com/prometheus/client_golang/pull/233, which in turn depends on this change here. The two repos import each other circularly :-/

fabxc:
👍

juliusv:
We'll probably have to revert this, given the discussion in https://github.com/prometheus/client_model/issues/13 - giving it a day more to settle, but doesn't look good right now.

grobie:
Please don't wait another day. This is regardless of the outcome of the discussion, at the moment all `go get`s of Prometheus projects are broken.

juliusv:
Push a revert directly to master: https://github.com/prometheus/common/commit/616e90af75cc300730196d04f3676f838d70414f

Issue #54, opened 2016-09-08
Can't handle federation text output


brian-brazil:
Prometheus can't scrape the text version of it's own federation output, e.g.

```
# TYPE cpu untyped
cpu{job="promclient_triton_5s",instance="container-metrics:9092",gs_instanceid="aabd3924",gs_svcuser="svcuser2",type="idle",gs_dc="us-east-1",gs_env="dev",id="prom-1"} 0.0886400014181114 1473307588759
# TYPE cpu untyped
cpu{instance="container-metrics:9092",gs_instanceid="aabd3924",gs_svcuser="svcuser2",type="wait",gs_dc="us-east-1",gs_env="dev",job="promclient_triton_5s",id="prom-1"} 0.6554035022739826 1473307588759
# TYPE cpu untyped
cpu{instance="container-metrics:9093",job="promclient_triton_5s",gs_dc="us-east-1",gs_env="dev",gs_instanceid="231deaa1",gs_svcuser="svcuser2",type="sys",id="prom-1"} 0.3062938875174634 1473307586226
# TYPE cpu untyped
cpu{job="promclient_triton_5s",gs_instanceid="fb310edz",gs_svcuser="svcuser1",type="user",gs_dc="us-east-1",instance="container-metrics:9091",gs_env="dev",id="prom-1"} 0.5589919972685718 1473307585070
```

From https://groups.google.com/forum/#!topic/prometheus-developers/ZVb2NjjUJ5A

beorn7:
As said in the discussion, this is a problem on the generation side, not on the parse side, see https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details

I guess the problem is that the federation code creates one metric family per time series and doesn't bundle them. That's already wrong in the protobuf representation, and the text parser mirrors that.

fabxc:
With the storage having no notion of metric families, that seems tricky to
do differently unless we want to apply some sorting to the entire output
first.

beorn7:
I'll look into how much effort that is. I could imagine it's worth the savings in bandwidth.

We can also approach the problem from the other side (not mutually exclusive) by loosing the requirement of not repeating the TYPE and HELP lines (last mention wins).
(I vaguely remember that I had implemented it in the loose way back then but deliberately enforced it later because of the spec... there might have been a concrete reason to do so but I don't remember.)

RichiH:
Instead of ordering, keeping a list of already-exposed metric TYPEs might be more efficient.

[...]

PR #55, opened 2016-09-08
Documentation fixes

beorn7:
This acknowledges the fact that duplicate metrics in the slice passed
to MetricFamilyToText will result in invalid text format.

This is meant to document the status quo and bears no implication for
the possible loosening of the text format spec, cf. #54.

@brian-brazil @fabxc


brian-brazil:
:+1:

PR #56, opened 2016-09-09
Add support for logging to Windows event log

carlpett:
Adds a new log fomatter which sends the event to the Windows Event log. Needed for the wmi_exporter.

### DIFF
--- a/log/eventlog_formatter.go
+++ b/log/eventlog_formatter.go
@@ -1,0 +1,89 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build windows
+
+package log
+
+import (
+	"fmt"
+	"os"
+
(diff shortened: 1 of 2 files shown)


log/eventlog_formatter.go
```diff
@@ -0,0 +1,90 @@
...
+		fallthrough
+	case logrus.ErrorLevel:
+		err = s.log.Error(102, e.Message)
+	case logrus.WarnLevel:
+		err = s.log.Warning(101, e.Message)
+	case logrus.InfoLevel:
+		err = s.log.Info(100, e.Message)
+	case logrus.DebugLevel:
+		if s.debugAsInfo {
+			err = s.log.Info(100, e.Message)
```
juliusv: Isn't this `err` always unused and overwritten due to the `fallthrough` below?

log/log.go
```diff
@@ -45,6 +46,9 @@ func (f levelFlag) Set(level string) error {
 // setSyslogFormatter is nil if the target architecture does not support syslog.
 var setSyslogFormatter func(string, string) error
 
+// setEventlogFormatter is nil if the target os does not support Eventlog (eg, is not Windows)
```
juliusv: os -> OS

eg -> i.e.

End doc comments with a period.

log/eventlog_formatter.go
carlpett: Indeed it is, and we also always log, and potentially twice. Removing the `fallthrough`. Not sure how that slipped in, good catch!

log/log.go
carlpett: Ok. As a non-native, I tend to confuse i.e. vs e.g., but now I've googled up some good memory tricks :)

carlpett:
@juliusv Thanks for the comments, fixed and squashed again

juliusv:
:+1: thanks!

PR #57, opened 2016-09-10
Standardize log flag descriptions

grobie:
Set the default values of -log.level and -log.format so that the flag
package can read them.

Before:

    -log.format value
          If set use a syslog logger or JSON logging. Example: logger:syslog?appname=bob&local=7 or logger:stdout?json=true. Defaults to stderr.
    -log.level value
          Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal].

After:

    -log.format value
          Set the log target and format. Example: "logger:syslog?appname=bob&local=7" or "logger:stdout?json=true" (default "logger:stderr")
    -log.level value
          Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal] (default "info")

@juliusv

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -28,11 +28,11 @@ import (
 	"github.com/Sirupsen/logrus"
 )
 
-type levelFlag struct{}
+type levelFlag string
 
 // String implements flag.Value.
 func (f levelFlag) String() string {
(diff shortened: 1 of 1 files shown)


juliusv:
:+1:

PR #61, opened 2016-09-21
route: Allow per-request custom context injection

juliusv:
**Motivation**

Currently, Prometheus's web API knows of two different contexts:

- The main Prometheus server context which is passed into the web API
  and through to the query layer.
- The per-request context which is generated by the route.Router. This
  is currently only used for storing and retrieving URL-path parameters,
  and it's not possible to influence this context at the top of the
  chain.

Since merging of contexts is not a concept, we ultimately want the
router to support a per-request context which:

a) can be derived from the main server context,
b) can be decorated with custom values/timeouts/etc. per-request

The concrete motivation for b) is for Weaveworks Frankenstein to be able
to extract multi-tenancy authentication headers and store user
information in the context. This will then be passed on into the query
layers and decide what data a user gets to see.

Next step: update Prometheus to make use of this, get rid of the context
in the web API itself. This will make the API package reusable by
Frankenstein.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -32,10 +32,38 @@ func WithParam(ctx context.Context, p, v string) context.Context {
 	return context.WithValue(ctx, param(p), v)
 }
 
-// handle turns a Handle into httprouter.Handle
-func handle(h http.HandlerFunc) httprouter.Handle {
-	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
-		ctx, cancel := context.WithCancel(context.Background())
+type contextFn func(r *http.Request) context.Context
+
+// Router wraps httprouter.Router and adds support for prefixed sub-routers
+// and per-request context injections.
+type Router struct {
+	rtr    *httprouter.Router
+	prefix string
+	ctxFn  contextFn
+}
+
+// New returns a new Router.
+func New(ctxFn contextFn) *Router {
+	if ctxFn == nil {
+		ctxFn = func(r *http.Request) context.Context {
+			return context.Background()
+		}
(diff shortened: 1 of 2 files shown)


juliusv:
@fabxc

route/route.go
```diff
@@ -32,67 +32,79 @@ func WithParam(ctx context.Context, p, v string) context.Context {
...
+// Router wraps httprouter.Router and adds support for prefixed sub-routers
+// and per-request context injections.
+type Router struct {
+	rtr    *httprouter.Router
+	prefix string
+	ctxFn  contextFn
+}
+
+// New returns a new Router.
+func New(ctxFn contextFn) *Router {
```
fabxc: Do you envision a use case where a context function here would be useful instead of a base context?
Seems like values passed into the context would be static to a base context anyway. 

Seems like only a timeout for all requests applied to a new context would be sane.
Could be more explicit to configure as a directly. But both approaches seem equally fine.

juliusv: Yes, I need a context function exactly for the Frankenstein case: to pull out per-request multi-tenancy information from the HTTP request headers and stuff that user info into the context.

Or did I misunderstand your question?

juliusv: Ping :)

juliusv:
@fabxc Any chance to get this and the related https://github.com/prometheus/prometheus/pull/2012 merged?

fabxc:
Yea, seems fine in general and doesn't break anything 👍 go ahead on both.

juliusv:
Thanks!

PR #63, opened 2016-10-02
route: Allow contextFn to return an error

juliusv:
The contextFn may be unable to create a request context if e.g. it tries
to extract authentication information from a header and the information
is not present or invalid.


juliusv:
@fabxc

fabxc:
:+1:

PR #64, opened 2016-11-07
Avoid precision loss in text output

disq:
In text-encoded output generated by MetricFamilyToText(), metric values larger than 1 million are exposed with exponents such as `1e+06`. This leads to a loss of precision. By formatting the metric value with [strconv.FormatFloat](https://golang.org/pkg/strconv/#FormatFloat) we can avoid it.

@fabxc

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"strconv"
 	"strings"
 
 	dto "github.com/prometheus/client_model/go"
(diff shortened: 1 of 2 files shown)


brian-brazil:
The presence of an exponent doesn't automatically mean there's a loss of prevision. Can you provide a unittest showing the issue?

disq:
Good point, I tried and could not prove it. Maybe it's Grafana's fault and I jumped the gun. I'll check. Thanks!

beorn7:
@disq Note what's written in https://golang.org/pkg/strconv/#FormatFloat : "The special precision -1 uses the smallest number of digits necessary such that ParseFloat will return f exactly."
Not losing precision here is by design.

disq:
The problem was somewhere else, sorry for the mixup.

PR #65, opened 2016-11-08
Adjust the MetricNameRE to reflect the actual behavior

beorn7:
Note that the validation in
https://github.com/prometheus/common/blob/master/model/metric.go#L93
does not actually use MetricNameRE. Right now, the regexp is more
documentation than anything else, although some might use it (which is
one of the reasons to keep it around).

In any case, the regexp should match exactly the same strings that are
matched by the IsValidMetricName function in
https://github.com/prometheus/common/blob/master/model/metric.go#L93
matches.

@brian-brazil

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -22,7 +22,7 @@ import (
 
 var (
 	separator    = []byte{0}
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1:

juliusv:
:+1:

PR #66, opened 2016-11-11
Remove last remaining use of LabelNameRE

beorn7:
The IsValid method is much faster.

Same as MetricNameRE, the regexp is mostly there for documentation
reasons.

This commit also clarifies that in the doc comments and encourages
using the IsValid method (or IsValidMatricName function) for
performance reasons.

@grobie @stuartnelson3

### DIFF
--- a/model/labels.go
+++ b/model/labels.go
@@ -80,14 +80,18 @@ const (
 	QuantileLabel = "quantile"
 )
 
-// LabelNameRE is a regular expression matching valid label names.
+// LabelNameRE is a regular expression matching valid label names. Note that the
+// IsValid method of LabelName performs the same check but faster than a metch
+// with this regular expression.
 var LabelNameRE = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 
 // A LabelName is a key for a LabelSet or Metric.  It has a value associated
 // therewith.
 type LabelName string
 
-// IsValid is true iff the label name matches the pattern of LabelNameRE.
+// IsValid is true iff the label name matches the pattern of LabelNameRE. ;This
+// method, however, does not use LabelNameRE for the check but a much faster
+// hardcoded implementation.
 func (ln LabelName) IsValid() bool {
 	if len(ln) == 0 {
 		return false
@@ -106,7 +110,7 @@ func (ln *LabelName) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal(&s); err != nil {
 		return err
 	}
-	if !LabelNameRE.MatchString(s) {
+	if !LabelName(s).IsValid() {
 		return fmt.Errorf("%q is not a valid label name", s)
 	}
 	*ln = LabelName(s)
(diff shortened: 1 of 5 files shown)


model/labels.go
```diff
@@ -80,14 +80,18 @@ const (
...
+// IsValid method of LabelName performs the same check but faster than a metch
+// with this regular expression.
 var LabelNameRE = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 
 // A LabelName is a key for a LabelSet or Metric.  It has a value associated
 // therewith.
 type LabelName string
 
-// IsValid is true iff the label name matches the pattern of LabelNameRE.
+// IsValid is true iff the label name matches the pattern of LabelNameRE. ;This
```
brian-brazil: The ; is meant to be a newline I guess?

model/metric.go
```diff
@@ -21,7 +21,10 @@ import (
 )
 
 var (
-	separator    = []byte{0}
+	separator = []byte{0}
+	// MetricNameRE is a regular expression matching valid metric
+	// names. Note that the IsValidMetricName function performs the same
+	// check but faster than a metch with this regular expression.
```
brian-brazil: match

model/metric_test.go
```diff
@@ -111,6 +111,14 @@ func TestMetricNameIsValid(t *testing.T) {
...
+			mn:    ":leading_colon",
+			valid: true,
+		},
+		{
+			mn:    "colon:in:the:middle",
+			valid: true,
+		},
 	}
 
 	for _, s := range scenarios {
```
brian-brazil: Should we also check the regex here?

beorn7:
Also added test for colons in metric names and label names, as that triggered this whole revisit..

model/labels.go
```diff
@@ -80,14 +80,18 @@ const (
 	QuantileLabel = "quantile"
 )
 
-// LabelNameRE is a regular expression matching valid label names.
+// LabelNameRE is a regular expression matching valid label names. Note that the
+// IsValid method of LabelName performs the same check but faster than a metch
```
grobie: match

model/metric_test.go
grobie: +1 for testing both implementations in tests.

model/metric.go
beorn7: Fixed.

beorn7:
Added tests for the regexps, too.

model/labels.go
```diff
@@ -80,14 +80,18 @@ const (
...
+// IsValid method of LabelName performs the same check but faster than a match
+// with this regular expression.
 var LabelNameRE = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 
 // A LabelName is a key for a LabelSet or Metric.  It has a value associated
 // therewith.
 type LabelName string
 
-// IsValid is true iff the label name matches the pattern of LabelNameRE.
+// IsValid is true iff the label name matches the pattern of LabelNameRE. This
```
stuartnelson3: I'm assuming `iff` is for "if and only if", but I personally find it unexpected when scanning programming documentation and jumps out to me as a typo. I would consider writing out the whole statement for clarity, but if this is an established pattern that's fine by me.

brancz: Agreed, in one of my first PRs I tried fixing it 😄

beorn7: I went through the same evolution of "typo" → "pretty standard". I always blamed myself not being a native speaker. Now Stuart fell into the typo trap, too. But I concur that it is used quite commonly. Once you start looking for it, it's everywhere.

beorn7:
The hand-coded functions have been introduced ages ago, with benchmarks and everything. This PR does not touch any of this, just removes left-over uses of the regexp. No need to justify the function again.

PR #67, opened 2016-12-01
route: make contextFn public type

fabxc:
Types showing up in signatures of other exported types
or functions should not be unexported.
Func is more consistent with the stdlib than Fn.


brancz:
👍

Issue #68, opened 2016-12-18
Log: Sylog logs should not contain cee tag unless they are json formatted


cagedmantis:
Logs sent to syslog currently tag logs (regardless of format) with the `@cee:` tag. 

`$ prometheus -log.format logger:syslog?appname=pm\&local=7`

Produces these incorrect logs in syslog:

```
Dec 18 09:26:48 Argon pm[62206]: @cee:ERRO[0000] Could not lock data/DIRTY, Prometheus already running?  source=persistence.go:181
Dec 18 09:26:48 Argon pm[62206]: @cee:ERRO[0000] Error opening memory series storage: resource temporarily unavailable  source=main.go:184
```

`$ prometheus -log.format logger:syslog?appname=pm\&local=7\&json=true`

Produces these correct logs in syslog:

```
Dec 18 09:25:04 Argon pm[62168]: @cee:{"level":"error","msg":"Could not lock data/DIRTY, Prometheus already running?","source":"persistence.go:181","time":"2016-12-18T09:25:04-05:00"}
Dec 18 09:25:04 Argon pm[62168]: @cee:{"level":"error","msg":"Error opening memory series storage: resource temporarily unavailable","source":"main.go:184","time":"2016-12-18T09:25:04-05:00"}
```

Syslog log entries should not be cee-enhanced unless they are json formatted.

PR #69, opened 2016-12-18
log: Only add cee tag to cee-enhanced logs. 

cagedmantis:
This PR is a follow-up to https://github.com/prometheus/common/issues/68

Changed cee tags so that they are only added to json formatted logs. Begin adding test to the syslog functionality.


beorn7:
Thanks, :+1:

PR #73, opened 2017-01-08
expfmt: Don't panic on decoding corrupt data

### LINKED ISSUE #72: expfmt: Don't panic on decoding corrupt data
https://github.com/prometheus/common/blob/master/expfmt/decode.go#L182 panics upon an unknown metric family type. This can be triggered by corrupt input data. We should return an error instead to not crash e.g. the whole Prometheus server during ingesting metrics with corrupted data.

beorn7:
Also, fix lint warnings.

Fixes #72 .

@juliusv 

Will fix the caller of expfmt.ExtractSamples in separate PR.

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -31,6 +31,7 @@ type Decoder interface {
 	Decode(*dto.MetricFamily) error
 }
 
+// DecodeOptions contains options used by the Decoder and in sample extraction.
 type DecodeOptions struct {
 	// Timestamp is added to each value from the stream that has no explicit timestamp set.
 	Timestamp model.Time
@@ -142,6 +143,8 @@ func (d *textDecoder) Decode(v *dto.MetricFamily) error {
 	return nil
 }
 
+// SampleDecoder wraps a Decoder to extract samples from the metric families
+// decoded by the wrapped Decoder.
 type SampleDecoder struct {
 	Dec  Decoder
 	Opts *DecodeOptions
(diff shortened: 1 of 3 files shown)


expfmt/decode.go
```diff
@@ -142,44 +143,56 @@ func (d *textDecoder) Decode(v *dto.MetricFamily) error {
...
+	*s, err = extractSamples(&sd.f, sd.Opts)
+	return err
 }
 
-// Extract samples builds a slice of samples from the provided metric families.
-func ExtractSamples(o *DecodeOptions, fams ...*dto.MetricFamily) model.Vector {
-	var all model.Vector
+// ExtractSamples builds a slice of samples from the provided metric
+// families. If an error occurs during sample extraction, it continues to
+// extract from the remaining metric families. The returned error is the lat
```
juliusv: lat -> last

expfmt/decode.go
```diff
@@ -142,44 +143,56 @@ func (d *textDecoder) Decode(v *dto.MetricFamily) error {
...
+// error that has occured.
```
juliusv: The code doesn't look like it's returning the last error, but just the last `err` value, which could be `nil`, even when there was a previous error? The generally less surprising behavior for me would be to not continue at all when there's an error.

beorn7: *bummer* Programming is hard.

beorn7:
Both fixed.

expfmt/decode.go
beorn7: About the "not surprising": I think there are two use cases if an error is returned: (1) play safe and discard everything (2) best effort, use as much as you can. 

For the 1st case, we would return nil in case of an error. For the 2nd, we do what I have implemented. Even with (2) implemented, you can still do (1) by simply ignoring the result. Just returning all the samples up to the error would not serve any of the use cases.

beorn7:
Being a good boy, I have now added a unit test, too.

PR #75, opened 2017-02-16
Fix comparison bug in Sample.Equal()

juliusv:


juliusv:
@beorn7

model/value.go
```diff
@@ -129,7 +129,7 @@ func (s *Sample) Equal(o *Sample) bool {
 	if !s.Timestamp.Equal(o.Timestamp) {
 		return false
 	}
-	if s.Value.Equal(o.Value) {
+	if !s.Value.Equal(o.Value) {
```
beorn7: Couldn't this just be
```Go
return s.Timestamp.Equal(o.Timestamp)
```
?

juliusv: Yup. Changed.

beorn7:
:+1:

PR #76, opened 2017-03-20
Fix typo in comment for LabelSet.Equal().

bweston92:


brian-brazil:
Thanks for the PR. The comment is correct, "iff" means "if and only if".

bweston92:
@brian-brazil that is madness but Google confirmed aha.

PR #79, opened 2017-04-08
Remove global map of contexts, use request.Context()

tomwilkie:
Part of https://github.com/prometheus/prometheus/pull/2554


tomwilkie:
Can we disable go 1.5 and go 1.6 tests on this repo?  request.Context was introduced in 1.7

brian-brazil:
client_golang depends on this repo, so it's a question of what we support for the client library.

tomwilkie:
Ah I see.  I guess I can try and do this in a backwards compatible way..

client_golang doesn't need the route package, and TBH now the context stuff has gone away there is very little value in this package.  I could just remove it use from prometheus.  WDYT?  Does anything else (other than alertmanager) use it?

tomwilkie:
@brian-brazil I've split the implementation for pre1.7 and post.  I'm not particularly keen on this, I'd prefer to move this package elsewhere and only support 1.7+.  The whole opentracing stuff won't work pre-1.7 anyway

brian-brazil:
This is ultimately @beorn7's call.

beorn7:
Please enlighten me where `client_golang` is using this package. I can't find it right now.

brian-brazil:
It pulls in `expfmt`.

beorn7:
But these changes only affect the `route` package.

grobie:
Let's remove old go versions here, as client_golang can specify their own
go versions it wants to test against.

I also don't know of a single go project which deliberately sticks to old
versions. In contrast to interpreted languages, it's easy to require recent
versions.

tomwilkie:
Have removed the compatibility code; @beorn7 let me know what you think!  I'll do a PR into prometheus and alertmanager to update this dependancy once this is merged.

beorn7:
Yes, let's remove Go 1.5 and Go 1.6 from the tests here.

tomwilkie:
@beorn7 sorry for the delay, have been on holiday the last week.  I removed go 1.5 & 1.6 from travis - PTAL.

beorn7:
:+1: 

@fabxc This is originally your code. Perhaps you want to do a final check? I have little context in my head how this code is used.

fabxc:
With Context now being part of the regular request object, this sure makes sense.
The whole route package becomes somewhat obsolete by that of course.

👍

tomwilkie:
> The whole route package becomes somewhat obsolete by that of course.

Yeah, its not super useful.  Its still pushes any 'path-matched-variables' into the context, and add prefixes to paths, which is kinda nice I guess.

PR #83, opened 2017-06-08
Add go-kit/log interface compatibility

fabxc:
@brian-brazil @grobie 

To some degree this is best effort I suppose. But practically this should work well for handing our log.Logger to any library using oklog or similar.

The lines a the middle are coming from tsdb.

screen shot 2017-06-08 at 17 12 04

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -14,6 +14,7 @@
 package log
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -149,11 +150,50 @@ type Logger interface {
 	Fatalln(...interface{})
 	Fatalf(string, ...interface{})
 
+	Log(keyvals ...interface{}) error
+
 	With(key string, value interface{}) Logger
 }
 
 type logger struct {
 	entry *logrus.Entry
+}
+
+func (l logger) Log(keyvals ...interface{}) (err error) {
+	if len(keyvals)%2 != 0 {
+		return errors.New("invalid arg count")
+	}
+	var lv string
+	var msg interface{}
+	fields := make(logrus.Fields, 5)
+
+	for i := 0; i < len(keyvals); i += 2 {
+		switch k := keyvals[i].(string); k {
+		case "level":
+			lv = fmt.Sprint(keyvals[i+1])
+		case "msg":
+			msg = fmt.Sprint(keyvals[i+1])
+		default:
+			fields[fmt.Sprint(keyvals[i])] = fmt.Sprint(keyvals[i+1])
+		}
+	}
+	ll := l.entry.WithFields(fields)
+
+	if msg == nil {
+		msg = ""
+	}
+	switch lv {
+	case "debug":
+		ll.Debug(msg)
+	case "warn":
+		ll.Warn(msg)
+	case "error":
+		ll.Error(msg)
+	default:
+		ll.Info(msg)
+	}
+
+	return nil
 }
 
 func (l logger) With(key string, value interface{}) Logger {


log/log.go
```diff
@@ -149,13 +150,52 @@ type Logger interface {
...
+	Log(keyvals ...interface{}) error
+
 	With(key string, value interface{}) Logger
 }
 
 type logger struct {
 	entry *logrus.Entry
 }
 
+func (l logger) Log(keyvals ...interface{}) (err error) {
```
grobie: Remove named return value.

log/log.go
```diff
@@ -149,13 +150,52 @@ type Logger interface {
...
 	Fatalf(string, ...interface{})
 
+	Log(keyvals ...interface{}) error
+
 	With(key string, value interface{}) Logger
 }
 
 type logger struct {
 	entry *logrus.Entry
 }
```
grobie: Missing doc string, should explain the behavior, the expectations of the input interface and what errors will be raised.

grobie: Should also mention that this implements go-kit/log

brian-brazil:
The middle lines are lacking the source code line number, that's a pretty useful piece of information in an application log.

This PR also doesn't the change the fact that we'd still be using two different Go logging libraries. The number we should be using is exactly one.

fabxc:
@brian-brazil there have been long discussions around how a standard Go logging interface should look like. @peterbourgon can tell you more.

The go-kit library requires implementing a single simple interface, which is trivial to do when having to plug an arbitrary logging library, like ours, into another package like prometheus/tsdb. If we were to use prometheus/common/log in prometheus/tsdb, this would be nearly impossible given the high amount of methods we require.

Showing the logged line can be helpful, but depending on your call nesting gives you results that aren't really useful. @peterbourgon, any way do work around that when using log/levels?
In general I think the logged message should be clear enough through usage of precise log key/vals and wrapping the error trace with pkg/errors.

brian-brazil:
I'm okay with switching, after we've all discussed this and come to a consensus about switching. Until then we should be using the existing agreed library.

Last time around when we had the big discussion that resulted in common/log existing, having line numbers was seen as a critical feature. So it's not something to be dropped lightly.

I don't have skin in the game beyond having one system and having line numbers. Searching for where a log line came from without knowing the file and line number is quite irritating.

peterbourgon:
@fabxc What is not working at the moment?

```go
package main

import (
	"os"

	"github.com/go-kit/kit/log"
	"github.com/go-kit/kit/log/level"
)

func main() {
	var logger log.Logger
	logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout))
	logger = level.NewFilter(logger, level.AllowWarn())
	logger = log.With(logger, "caller", log.DefaultCaller)
	logger = log.With(logger, "ts", log.DefaultTimestampUTC)

	foo(log.With(logger, "func", "foo"))
}

func foo(logger log.Logger) {
	level.Warn(logger).Log("hello", "world")
}
```

```
$ go run main.go
level=warn caller=main.go:21 ts=2017-06-19T09:54:12.208580733Z func=foo hello=world
```

fabxc:
Weird! I had `log.DefaultCaller` resulting in `caller` being set to files of the log package. Maybe I wired it up incorrectly.

peterbourgon:
The things that return log.Valuer typically need to go into a log.With statement, they aren't designed to stand alone in a logger.Log invocation. But glad that sorts you out :) The order of the keys in the log output is controlled by the order of the decorator declarations, by the way, so if you want e.g. timestamp-level-caller, it's also easy.

fabxc:
Ah no, now I recall. The problem was that this log package attaches the source information right before printing. That's what I added here too and what then spit out oklog source files. That's more of a general problem of this package when it comes to plugging it into adapters.
I don't think logrus has the same capability as logrus to do that right thing here.

@brian-brazil  Can we merge this and I'll submit a PR to Prometheus integrating it. We'll wire up there to keep printing the source.

brian-brazil:
> Can we merge this and I'll submit a PR to Prometheus integrating it. We'll wire up there to keep printing the source.

There's more to this than Prometheus. This is used in basically every other Go exporter too, and several 3rd party integrations.
We need consensus that this is the way we're going before we start doing all that work.

In any case we should probably finish the Cobra switch first, as that's a similarly large project that's just getting going.

fabxc:
Nothing changes for any of those. So do you want to block this?
That means a) we switch all of TSDB to this log package, which I'll strongly oppose b) we do nothing and have TSDB just log for its own like it does now.

brian-brazil:
I'm saying that there isn't presently a consensus on switching to go-kit for logging.
I'd suggest starting a thread on -dev with the aim of achieving consensus for your proposal to switch everything over.

If there is not consensus to switch, we would follow the standing consensus which would mean a).

My concern is primarily around having a homogeneous codebase, not which exact log library we use.

brian-brazil:
This has all been done in other changes.

PR #84, opened 2017-06-09
Check for invalid utf-8 values in label values.

brian-brazil:
https://github.com/prometheus/prometheus/issues/2822

@fabxc


brian-brazil:
I haven't looked at the code yet, when I tested it I got an error.

fabxc:
👍

PR #85, opened 2017-06-15
Fix logrus import

abhinav:
The author of logrus has changed his GitHub username from Sirupsen to
sirupsen. The case change breaks logrus imports for packages whose
dependencies import logrus using the other casing.

See also, sirupsen/logrus#384


brian-brazil:
Thanks!

You'll likely need to re-vendor the repos using this.

PR #87, opened 2017-06-15
Add functions to reduce use of globals.

Gouthamve:
These functions will be used to construct and pass around the logger in prometheus.

@grobie @fabxc @brian-brazil

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/common/87)

### DIFF
--- a/log/eventlog_formatter.go
+++ b/log/eventlog_formatter.go
@@ -25,18 +25,18 @@ import (
 )
 
 func init() {
-	setEventlogFormatter = func(name string, debugAsInfo bool) error {
+	setEventlogFormatter = func(l logger, name string, debugAsInfo bool) error {
 		if name == "" {
 			return fmt.Errorf("missing name parameter")
 		}
 
-		fmter, err := newEventlogger(name, debugAsInfo, origLogger.Formatter)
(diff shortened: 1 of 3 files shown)


grobie:
LGTM but I'm concerned about the amount of code we write without tests.

fabxc:
@grobie that's more of a package structure than a semantic change. The larger added code block is a straight copy of https://github.com/prometheus/common/blob/master/log/log.go#L67-L107 AFAICS.

@Gouthamve maybe we could deduplicate that?

Gouthamve:
Done. No clue why I did not do that before, sorry.

fabxc:
:+1:

grobie:
@fabxc I'm aware that it was mostly copied code. I don't see how that justifies not having tests?

fabxc:
It's a logging library wrapper. I know 90%+ test coverage would be fantastic. As it always has been, we do not have remotely the resources to invest that kind of effort.
We have a 2.0 release to get out. Thew new TSDB is significantly better tested than this package here. I hope that's somewhat reassuring :)

grobie:
It is!

PR #89, opened 2017-06-16
Add pflag Value support to Duration

Gouthamve:
Ref prometheus/prometheus#1883

@grobie @fabxc @brian-brazil

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/common/89)

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -163,9 +163,21 @@ func (t *Time) UnmarshalJSON(b []byte) error {
 // This type should not propagate beyond the scope of input/output processing.
 type Duration time.Duration
 
+// Set implements pflag/flag.Value
+func (d *Duration) Set(s string) error {
+	var err error
+	*d, err = ParseDuration(s)
+	return err
+}
+
+// Type implements pflag.Value
+func (d *Duration) Type() string {
+	return "model.Duration"
+}
+
 var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 
-// StringToDuration parses a string into a time.Duration, assuming that a year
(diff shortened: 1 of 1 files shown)


model/time.go
```diff
@@ -163,9 +163,20 @@ func (t *Time) UnmarshalJSON(b []byte) error {
 // This type should not propagate beyond the scope of input/output processing.
 type Duration time.Duration
 
+// Set implements pflag/flag.Value
+func (d *Duration) Set(s string) (err error) {
```
grobie: Nit: A named return variable is stutter here.

Gouthamve: Umm, so it would make it an extra step as one cannot do `*d, err := ` with the receiver `*d`. 

I can change, but isn't naming the return value a good practice?

grobie: Only if they add to the documentation. `err error` is stutter and clutters the documentation. `(max int, min int)` is good and preferred over `(int, int)`.

https://go-review.googlesource.com/c/20024/

Gouthamve: TIL. Made the change. Thanks!

model/time.go
```diff
@@ -163,9 +163,21 @@ func (t *Time) UnmarshalJSON(b []byte) error {
...
+// Set implements pflag/flag.Value
+func (d *Duration) Set(s string) error {
+	var err error
+	*d, err = ParseDuration(s)
+	return err
+}
+
+// Type implements pflag.Value
+func (d *Duration) Type() string {
+	return "model.Duration"
```
grobie: @Gouthamve I wasn't aware this is the value which will be printed in the help page / usage. I don't think it makes sense to use our internal data model here. Probably just `duration`?

PR #90, opened 2017-06-16
Make Type() return duration for UX

Gouthamve:
@grobie Sorry about that. You are right, this is better UX.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/common/90)


grobie:
:+1:

PR #94, opened 2017-07-13
Fall back on environment http proxies

Fra-nk:
If no explicit proxy URL is defined, try to use the proxies (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) defined in the process' environment.

Pull request as followup of https://github.com/prometheus/blackbox_exporter/issues/190

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -121,9 +121,16 @@ func NewHTTPClientFromConfig(cfg *HTTPClientConfig) (*http.Client, error) {
 		return nil, err
 	}
 
+	var proxyFunc func(*http.Request) (*url.URL, error)
+	if cfg.ProxyURL.URL != nil {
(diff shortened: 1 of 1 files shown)


brian-brazil:
This isn't the one Prometheus uses, you want https://github.com/prometheus/prometheus/blob/master/util/httputil/client.go#L45

Fra-nk:
I"m slightly confused.
Blackbox exporter includes `github.com/prometheus/common/config` in `http.go` and should use the code-line that I changed.

So is your comment meant as "please first contribute it to Prometheus server, it will then be forwarded/ported to Prometheus common, if it's accepted" or am I reading the code wrong?

brian-brazil:
Yes, that's what I mean.

PR #95, opened 2017-07-21
Remove program prefix from build_info metric

infoverload:
In reference to:

https://github.com/prometheus/prometheus/issues/2566


brian-brazil:
Thanks for the PR. I was going off the title of that bug rather than the content, so this change is something that'll need further discussion.

brian-brazil:
Thanks for your patience, we've determined that https://github.com/prometheus/client_golang/issues/328 is a better option here.

PR #96, opened 2017-07-29
Switch to kingpin in log

carlpett:
This switches the flag package to kingpin as discussed in [Prometheus#2455](https://github.com/prometheus/prometheus/issues/2455) and [influxdb_exporter#15](https://github.com/prometheus/influxdb_exporter/pull/15).

I've also tested building and running Prometheus with the dev-2.0 branch to ensure these changes does not impact how logging flags are initialized there, and it seems to work as expected.

People not vendoring this package will get a surprise though if this is merged. Is this something that needs to be considered?

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -24,9 +24,8 @@ import (
 	"strconv"
 	"strings"
 
-	"gopkg.in/alecthomas/kingpin.v2"
-
 	"github.com/sirupsen/logrus"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 // setSyslogFormatter is nil if the target architecture does not support syslog.
--- a/log/log.go
+++ b/log/log.go
@@ -14,7 +14,6 @@
 package log
 
 import (
-	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -25,25 +24,10 @@ import (
 	"strconv"
 	"strings"
 
+	"gopkg.in/alecthomas/kingpin.v2"
+
 	"github.com/sirupsen/logrus"
 )
-
-type levelFlag string
-
-// String implements flag.Value.
-func (f levelFlag) String() string {
-	return fmt.Sprintf("%q", origLogger.Level.String())
-}
-
-// Set implements flag.Value.
-func (f levelFlag) Set(level string) error {
-	l, err := logrus.ParseLevel(level)
-	if err != nil {
-		return err
-	}
-	origLogger.Level = l
-	return nil
-}
 
 // setSyslogFormatter is nil if the target architecture does not support syslog.
 var setSyslogFormatter func(logger, string, string) error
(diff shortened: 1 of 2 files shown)


carlpett:
Missed mentioning @fabxc 
Not sure about the failing test, it works on my machine (tm)... 
Also @Gouthamve and @brian-brazil, does this line up with the discussions on the other PR?

Gouthamve:
While I would agree with the function, I don't think we should have an init function. What if someone wants to define his own flags?

See: https://github.com/prometheus/prometheus/issues/2455#issuecomment-317676475

carlpett:
This appears to be the area of contest, as @brian-brazil wants to avoid having client needing to call this initialization function. 
My own use-case for overriding these flags for the wmi_exporter has disappeared because apparently powershell does not do the same sort of tokenization for double-dashed flags (I'm guessing the reason for that happening is that `-SomeFloatFlag.5` is to be a valid way to pass `0.5`). Are there any other cases where someone would realistically want to use the Prometheus log package without following Prometheus patterns for flags?
I guess we could have some global bool which could be set in a consuming package's `init` to disable it in order to cater to both camps here (set flags by default, possible to disable), but I'm not sure it is warranted?

brian-brazil:
I suggested this without the init(), as there are some rare use cases where users don't want this to happen automatically and are willing to take on the maintenance burden of defining their own flags.

carlpett:
@brian-brazil Aha, my misunderstand then. Anyone fine with the normal flags will then call `log.AddFlags`. I'll update the PR.

log/log.go
```diff
@@ -55,38 +39,30 @@ func setJSONFormatter() {
...
-// Set implements flag.Value.
-func (f logFormatFlag) Set(format string) error {
-	return baseLogger.SetFormat(format)
-}
-
-func init() {
-	AddFlags(flag.CommandLine)
-}
-
-// AddFlags adds the flags used by this package to the given FlagSet. That's
```
brian-brazil: This comment should be kept and adjusted, as this is a public function.

carlpett: Oops. Re-added.

brian-brazil:
:+1: 

There's still a failing test.

carlpett:
Ah, now I see, non-latest version of logrus locally. The latest version of logrus apparently stops appending a space after the log line, which breaks the regexp test.

carlpett:
Apparently caused by this PR: https://github.com/sirupsen/logrus/pull/591.

log/log.go
```diff
@@ -25,26 +24,11 @@ import (
 	"strconv"
 	"strings"
 
+	"gopkg.in/alecthomas/kingpin.v2"
+
```
grobie: Please remove the extra line and list all external package dependencies in one block.

carlpett: @grobie Fixed.

brian-brazil:
Thanks!

PR #108, opened 2017-10-16
Include go1.8 and go1.9

taion809:
This change adds the last 2 versions of go into the build matrix.


brian-brazil:
This is outdated, closing.

PR #113, opened 2017-12-22
add color logging capabilities for promlog

cstyan:
Add's color logging capabilities to promlog when we're logging to the terminal, otherwise we just get a standard go-kit logger when logging to a file. 
We'll have to update deps/vendor packages in prometheus after we merge.

cc: @krasi-georgiev @juliusv

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/go-kit/kit/log/term"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


cstyan:
cc: @fabxc

krasi-georgiev:
Did you check if colours can be applied to selected name=value groups rather than the entire line?

cstyan:
AFAICT, that would require a change to the `go-kit/log/term` package. So unless anyone thinks it's really necessary, I'll just leave the colors change as is for now.

brian-brazil:
Is this still wanted?

SuperQ:
This PR seems fine, but it needs a rebase for the change to `go-kit/log`

promlog/log.go
```diff
@@ -31,6 +32,25 @@ type AllowedLevel struct {
...
+			continue
+		}
+		switch keyvals[i+1] {
+		case "debug":
+			return term.FgBgColor{Fg: term.Blue}
+		case "warn":
+			return term.FgBgColor{Fg: term.Yellow}
+		case "error":
+			return term.FgBgColor{Fg: term.Red}
+		default:
```
juliusv, line 47: How about green for INFO?

SuperQ: Sure

SuperQ:
This package is deprecated (https://github.com/prometheus/common/pull/720).

PR #114, opened 2018-01-01
Improve performance of Time to/from JSON

bboreham:
A few changes to speed up marshalling/unmarshalling of `model.Time` - see individual commit messages.

Benchmarks: before:
```
BenchmarkMarshalJSON-2     	 5000000	       303 ns/op	      64 B/op	       3 allocs/op
BenchmarkUnmarshalJSON-2   	 5000000	       314 ns/op	      53 B/op	       5 allocs/op
```

after:
```
BenchmarkMarshalJSON-2     	10000000	       219 ns/op	      16 B/op	       1 allocs/op
BenchmarkUnmarshalJSON-2   	10000000	       124 ns/op	      16 B/op	       2 allocs/op
```

@fabxc

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -117,7 +117,7 @@ func (t Time) String() string {
 
 // MarshalJSON implements the json.Marshaler interface.
 func (t Time) MarshalJSON() ([]byte, error) {
-	return []byte(t.String()), nil
+	return strconv.AppendFloat(nil, float64(t)/float64(second), 'f', -1, 64), nil
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface.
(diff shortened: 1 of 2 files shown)


bboreham:
Note prometheus/prometheus#3536 overrides `MarshalJSON` in `Point` so this PR has less impact.  I still have cases where I'm marshalling `Time` directly.

#3536 doesn't touch the `UnmarshalJSON` side.

brian-brazil:
Where is this code getting used? We're trying to move away from model generally.

bboreham:
Aha, hadn't picked up on that shift.

Used here: https://github.com/weaveworks/cortex/blob/cc8cdf499162077c4a21ae05d162ca8116228718/pkg/chunk/chunk.go#L44-L45
no biggy to change it to underlying types.

brian-brazil:
Changing to underlying types is probably best, it'll be less of a problem when this code eventually goes away.

bboreham:
In the end I moved my code to use `jsoniter` and added type overrides to achieve the same effect.

You might consider a note in the README or CONTRIBUTING here that `model` is deprecated.

PR #116, opened 2018-01-15
Expose Filter() function on AllowedLevel

bboreham:
This makes it possible to use AllowedLevel outside of this package.

This change would help with https://github.com/weaveworks/cortex/issues/655

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -53,11 +53,16 @@ func (l *AllowedLevel) Set(s string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


bboreham:
@fabxc as maintainer any thoughts?  or @juliusv ?

juliusv:
Looks ok to me, but I'd prefer someone not associated with Cortex to also :+1:

brian-brazil:
This looks to be obsolete.

PR #119, opened 2018-02-28
Add charset to text/plain Content-Type

robbiet480:
This fixes encoding issues when viewing `/metrics` directly.

See also https://github.com/prometheus/docs/issues/557, https://github.com/prometheus/prometheus/issues/1988

### DIFF
--- a/expfmt/expfmt.go
+++ b/expfmt/expfmt.go
@@ -26,7 +26,7 @@ const (
 
 	// The Content-Type values for the different wire protocols.
 	FmtUnknown      Format = `<unknown>`
-	FmtText         Format = `text/plain; charset=utf-8; version=` + TextVersion
+	FmtText         Format = `text/plain; version=` + TextVersion + `; charset=utf-8`
 	FmtProtoDelim   Format = ProtoFmt + ` encoding=delimited`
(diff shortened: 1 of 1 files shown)


expfmt/expfmt.go
```diff
@@ -26,7 +26,7 @@ const (
 
 	// The Content-Type values for the different wire protocols.
 	FmtUnknown      Format = `<unknown>`
-	FmtText         Format = `text/plain; version=` + TextVersion
+	FmtText         Format = `text/plain; charset=utf-8; version=` + TextVersion
```
brian-brazil: We put it at the end usually: https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L28

robbiet480:
@brian-brazil I thought about that but couldn’t find definitive proof either way that was “legal” or not. I’ll change it to your suggestion momentarily.

brian-brazil:
I believe it's valid both ways offhand, but may aswell keep things consistent in case someone is doing over-simplistic handling of the header.

robbiet480:
LGTM

brian-brazil:
Thanks!

PR #120, opened 2018-03-01
route: support for instrumentation

simonpasquier:


route/route.go
```diff
@@ -33,9 +34,22 @@ func New() *Router {
...
+}
+
 // WithPrefix returns a router that prefixes all registered routes with prefix.
 func (r *Router) WithPrefix(prefix string) *Router {
-	return &Router{rtr: r.rtr, prefix: r.prefix + prefix}
+	return &Router{rtr: r.rtr, prefix: r.prefix + prefix, instrh: r.instrh}
+}
+
+// InstrumentHandler returns the handler wrapped by the instrumentation handler.
+func (r *Router) InstrumentHandler(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
```
brian-brazil: This shouldn't be exposed in the API, instead use it within the class

simonpasquier: Do you mean wrapping with the instrumentation handler every time a route is registered (eg in `Get()`, `Post()`, ...) using `path` as the `handler` label value?

brian-brazil: Yes, that's one way.

simonpasquier:
@brian-brazil please take a look.

brian-brazil:
Thanks!

PR #121, opened 2018-03-18
add a common makefile

krasi-georgiev:
an idea how to reuse common rules in a Makefile across projects
fixes https://github.com/prometheus/prometheus/issues/3904

the common makefile is to gather common rules
it will be pulled down by projects that want to use it but each project will also keep a local cached copy Makefile.common  so a developper can still work offline.

here is how the prometheus makefile will look like.
https://github.com/prometheus/prometheus/pull/3978/files

cc @brian-brazil


krasi-georgiev:
also noticed that check_license.sh should also go in common.

PR #123, opened 2018-03-24
Support http client timeout

plaflamme:
This adds support for configuring the `http.Client` timeout parameter, allowing the specification of a global request timeout.

@fabxc This is required to refactor the marathon SD to use the common HTTP client configuration. The current version allows specifying the timeout: https://github.com/prometheus/prometheus/blob/master/discovery/marathon/marathon.go#L88

There are alternate approaches obviously, but this seemed like more useful generally.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -21,7 +21,9 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
+	"github.com/prometheus/common/model"
 	"gopkg.in/yaml.v2"
 )
 
@@ -74,6 +76,8 @@ type HTTPClientConfig struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
+	// Timeout to apply to every request.
+	Timeout model.Duration `yaml:"timeout,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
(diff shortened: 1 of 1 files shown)


brian-brazil:
This doesn't belong here, many of the places using this code already have a timeout and this would cause duplication.

plaflamme:
@brian-brazil I was afraid this might cause problems with other usage. Perhaps it's a style requirement, but the fact that the http client config is inlined with other configuration seems like it creates more problems than it solves. It seems to me like the best approach here would be for the http client config to require its own block of properties: this would make all usage consistent, allow documenting it separately and eventually allow more settings to be configurable.

Given that it would be a pretty invasive change, I see these alternate solutions:

* rename the configuration property to reduce the chance of conflict (e.g.: `http_request_timeout`, `http_client_timeout`)
* take a `Timeout` parameter of type `model.Duration` on `NewHTTPClientFromConfig`
* make no changes here and mutate the returned client in `marathon.go` to override the `Timeout` member (not sure if that's an acceptable style)
* drop support for `Timeout` in `marathon_sd` (seems doable given it's [not documented](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<marathon_sd_config>))

Please let me know what you feel would be the best approach here (the ultimate goal is to refactor `marathon.go` to use `HTTPClientConfig`)

plaflamme:
Forgot another alternative solution:

* create a new function that only returns an `http.Transport` instead of a full-blown `http.Client`, e.g.: `NewHTTPTransportFromConfig`

plaflamme:
Please see https://github.com/prometheus/common/pull/124 for a take on the last alternative mentioned. I'll close whichever PR you feel is unnecessary, potentially both.

brian-brazil:
The simplest option would be to remove the timeout configuration option for this SD mechanism, none of the other HTTP SDs have a timeout option and that has worked out fine.

brian-brazil:
Also the way to configure a timeout is to use a context.

plaflamme:
Thanks for the feedback.

PR #124, opened 2018-03-24
Expose function to build a RoundTripper from HTTPClientConfig

plaflamme:
This allows clients to build their own `http.Client` while reusing `HTTPClientConfig`.

@brian-brazil This is a take on one of the alternative approaches. Please let me know if this is a reasonable change.

I wasn't sure about the use of pointers here (I'm not a golang dev). The existing functions on `RoundTripper` aren't using pointers, so I used the same style, but the similar `NewHTTPClientFromConfig` function is returning a pointer to the `http.Client`, so perhaps the new function should also return a pointer?

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -115,9 +115,9 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return checkOverflow(a.XXX, "basic_auth")
 }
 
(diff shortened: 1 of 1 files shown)


plaflamme:
This is no longer necessary as the support for specifying the client timeout is dropped entirely from `marathon_sd`

PR #127, opened 2018-04-02
Replace yaml Unmarshal workaround with UnmarshalStrict.

### LINKED ISSUE #126: remove checkOverflow workaround from the config package
`checkOverflow` and `XXX map[string]interface{} `yaml:",inline"` are  no longer needed as the yaml package introduced the `UnmarshalStrict()` to detect unknown or duplicated fields. 

After removing `checkOverflow` need to find all packages that use `common/config` , update the yaml package and start using the new `UnmarshalStrict()`.

manosf:
This resolves #126.
As with prometheus/prometheus/pull/4033, after updating the yaml vendor package, `checkOverflow()` is no longer needed since `UnmarshalStrict()` covers duplicate and unknown fields in yaml files.
Also some cleanup after deleting some -no longer necessary- variables and `checkOverflow()` calls.
@fabxc, there are no major changes, so I believe this can be merged right after prometheus/prometheus/pull/4033, if everything looks good of course.
cc @krasi-georgiev

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,23 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This package no longer handles safe yaml parsing. In order to
+// ensure correct yaml unmarshalling, use "yaml.UnmarshalStrict()".
+
 package config
-
-import (
-	"fmt"
-	"strings"
-)
-
-func checkOverflow(m map[string]interface{}, ctx string) error {
-	if len(m) > 0 {
-		var keys []string
-		for k := range m {
-			keys = append(keys, k)
-		}
-		return fmt.Errorf("unknown fields in %s: %s", ctx, strings.Join(keys, ", "))
-	}
-	return nil
-}
 
 // Secret special type for storing secrets.
 type Secret string
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -32,9 +32,6 @@ type BasicAuth struct {
 	Username     string `yaml:"username"`
 	Password     Secret `yaml:"password,omitempty"`
 	PasswordFile string `yaml:"password_file,omitempty"`
-
-	// Catches all undefined fields and must be empty after parsing.
-	XXX map[string]interface{} `yaml:",inline"`
 }
 
 // URL is a custom URL type that allows validation at configuration load time.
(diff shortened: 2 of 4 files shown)


krasi-georgiev:
can you make a list of all packages that use this so we keep track which would need updating after this change.

brian-brazil:
To be safe we should probably check this in last, so noone picks this up before they have converted to strict. We should also have a comment somewhere that we presume strict parsing.

manosf:
I made the list of Prometheus programs' files which import `github.com/common/config`\[1\].
The names of the programs that do not import this package are only referenced on the bottom.

I'm going to open a PR in each of these programs soon too, in order to replace `Unmarshal()` with `UnmarshalStrict()` on the files that have not yet been converted.

\[1\]: [commonconfigimports.txt](https://github.com/prometheus/common/files/1873452/commonconfigimports.txt)

brian-brazil:
There's also yaml unmarshalling in the snmp exporter, but it doesn't use this library.

krasi-georgiev:
@manosf you misunderstood my comment. I meant we need to make a list and update all packages that import `prometheus/common` so that they use use `yaml.UnmarshalStrict` and not rely on the common package to throw an error for missing or duplicated fields.
Also we need a comment somewhere that  `prometheus/common` doesn't handle missing or duplicated fields and any package that want to use it should call `yaml.UnmarshalStrict`

manosf:
@krasi-georgiev That's what I did. I updated all Prometheus' external modules that relied on `prometheus/common` for safe yaml unmarshalling to use `yaml.UnmarshalStrict()`.
I do not know where to add such a comment or even if it is something that I have the authorization to do (I don't think that a simple comment inside one among dozens of files is clear enough).

krasi-georgiev:
Ah , ok , so why  `commonconfigimports.txt` list includes many packages that don't import the `common` package?

I did a quick search and these came up (most are not even ours, but couldn't find a way to filter.)
https://github.com/search?p=1&q=%22github.com%2Fprometheus%2Fcommon%2Fconfig%22+extension%3A.go+language%3AGo+language%3AGo&type=Code&utf8=%E2%9C%93
so if these have been updated to use UnmarshalStrict we can merge this one after a rebase.

About the comment maybe a good place would be just above the package definition.

```
// .....
package config
```

manosf:
I wrote on the comment in which I provided the txt that I am simply referencing the modules that do not import `common/config` on the bottom, for the sake of being thorough/clear.
I'll rebase then.

manosf:
@krasi-georgiev Done.

krasi-georgiev:
@manosf thanks , all looks ok now.

@brian-brazil please review

brian-brazil:
Thanks!

PR #129, opened 2018-04-10
support HTTP basic_auth.password_file 

adamdecaf:
Issue: https://github.com/prometheus/prometheus/issues/4074
Issue: https://github.com/prometheus/prometheus/issues/4076


adamdecaf:
I assume the deps in prometheus need updated after this is merged?

adamdecaf:
Oh. I realize now that prometheus has its own implementation. I'm going to just PR that also. 

https://github.com/prometheus/prometheus/blob/master/util/httputil/client.go#L77-#L79

adamdecaf:
Let me know on https://github.com/prometheus/prometheus/issues/4076 which route I should take with this. Thanks!

adamdecaf:
@brian-brazil Updated.

config/http_config.go
```diff
@@ -146,7 +151,15 @@ func NewHTTPClientFromConfig(cfg *HTTPClientConfig) (*http.Client, error) {
 	}
 
 	if cfg.BasicAuth != nil {
-		rt = NewBasicAuthRoundTripper(cfg.BasicAuth.Username, Secret(cfg.BasicAuth.Password), rt)
+		pass := cfg.BasicAuth.Password
+		if string(pass) == "" || cfg.BasicAuthPasswordFile != "" {
+			bs, err := ioutil.ReadFile(cfg.BasicAuthPasswordFile)
```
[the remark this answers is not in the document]
adamdecaf: I'm not quite sure I follow. Should I drop the `string(pass) == ""` and always prefer a password_file if the path is non-empty?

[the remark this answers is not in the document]
brian-brazil: If there's both a password and a password_file that should be rejected when the config is unmarahslled.

adamdecaf:
@brian-brazil This is ready to merge.

config/http_config.go
```diff
@@ -88,6 +89,12 @@ func (c *HTTPClientConfig) Validate() error {
 	if c.BasicAuth != nil && (len(c.BearerToken) > 0 || len(c.BearerTokenFile) > 0) {
 		return fmt.Errorf("at most one of basic_auth, bearer_token & bearer_token_file must be configured")
 	}
+	if c.BasicAuth != nil && c.BasicAuth.Username == "" {
+		return fmt.Errorf("basic_auth requires a username")
+	}
+	if c.BasicAuth != nil && (string(c.BasicAuth.Password) != "" && c.BasicAuth.PasswordFile != "") {
+		return fmt.Errorf("don't specify a basic_auth password and password_file")
```
simonpasquier: minor but it would be good to be consistent with the other error messages, eg `"at most one of basic_auth password & password_file must be configured"`.

adamdecaf: Good catch. Fixed.

config/http_config.go
```diff
@@ -29,6 +29,7 @@ import (
 type BasicAuth struct {
 	Username string `yaml:"username"`
 	Password Secret `yaml:"password"`
```
brian-brazil: Shouldn't this be omitempty now too?

config/http_config.go
```diff
@@ -146,7 +153,15 @@ func NewHTTPClientFromConfig(cfg *HTTPClientConfig) (*http.Client, error) {
 	}
 
 	if cfg.BasicAuth != nil {
-		rt = NewBasicAuthRoundTripper(cfg.BasicAuth.Username, Secret(cfg.BasicAuth.Password), rt)
+		pass := cfg.BasicAuth.Password
+		if cfg.BasicAuth.PasswordFile != "" {
+			bs, err := ioutil.ReadFile(cfg.BasicAuth.PasswordFile)
```
brian-brazil: This still needs to be read on every request.

adamdecaf: @brian-brazil The Bearer auth doesn't read it's file on each request. I'm fine with changing this, but I think both options should be consistent with each other. 

https://github.com/prometheus/common/blob/master/config/http_config.go#L133-L142

config/http_config.go
adamdecaf: Whoops. Updated.

config/http_config.go
brian-brazil: This is due to us ending up with duplicate code over in Prometheus. They both should read on every request.

adamdecaf: @brian-brazil Hmm. ok. Can we break the `NewBasicAuthRoundTripper` interface? I need to pass through the passwordFile param.

brian-brazil: https://github.com/prometheus/prometheus/blob/master/util/httputil/client.go already implements this, you can copy that and we'll then delete that duplication later.

adamdecaf:
@brian-brazil Ok. Apologies for the back and forth on this. I copied the http client code from prometheus as you linked.

config/http_config.go
```diff
@@ -115,19 +124,37 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 	if err != nil {
 		return nil, err
 	}
+	return newClient(rt), nil
+}
 
-	// It's the caller's job to handle timeouts
+// NewRoundTripperFromConfig returns a new HTTP RoundTripper configured for the
+// given config.HTTPClientConfig. The name is used as go-conntrack metric label.
+func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTripper, error) {
```
brian-brazil: Copy all this stuff wholesale from Prometheus, this code currently isn't handling bearer tokens correctly. The new code you add should follow how Prometheus currently does bearer tokens.

adamdecaf: @brian-brazil Done.

adamdecaf:
@brian-brazil How's this look?

brian-brazil:
Thanks!

PR #131, opened 2018-05-05
Adding json support for model.Duration

bradbeam:
Otherwise it will attempt to unmarshal as time.Duration int64 and panic because
ParseDuration will not be called.

Ex,
```
rulefmt.RuleGroups.Groups: []rulefmt.RuleGroup: rulefmt.RuleGroup.Rules: []rulefmt.Rule: rulefmt.Rule.For: readUint64: unexpected character: \xff, error found in #10 byte of ...|n\",\"for\":\"5m\",\"label|..., bigger context ...|ut(instance) (rate(requests_total[5m]))\\n\",\"for\":\"5m\",\"labels\":{\"severity\":\"critical\"}}]}]}}\n|..."}
```

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"regexp"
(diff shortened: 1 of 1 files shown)


bradbeam:
We're using this to expose RuleGroups as a CRD to better manage alert rules.

Ex,
```
type Alert struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata"`
        Spec              rulefmt.RuleGroups `json:"spec"`
        Status            AlertStatus        `json:"status,omitempty"`
}
```

Specifically, https://github.com/operator-framework/operator-sdk/blob/v0.0.5/pkg/util/k8sutil/k8sutil.go#L57-L72 is where we are running into issues.

brian-brazil:
These are not a public API, so we don't do things that encourage use of them. Model is also generally meant to be going away,

PR #132, opened 2018-05-15
Support carriage-returns in expfmt parser

carlpett:
This PR adds support for \r in the expfmt text parser. This makes it easier to use in a Windows environment (such as the textfile collector in node_exporter or wmi_exporter).

The approach taken here is pretty simple, and just treat \r the same as \n. I'm not sure if it is sufficient, or if there actually needs to be a new state in the parser for this?

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -152,6 +152,8 @@ func (p *TextParser) startOfLine() stateFn {
 		return p.startComment
 	case '\n':
 		return p.startOfLine // Empty line, start the next one.
+	case '\r':
+		return p.startOfLine // Empty line, start the next one.
 	}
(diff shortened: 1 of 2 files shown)


brian-brazil:
This is not something that we want to do. Adding an additional supported line ending would be a breaking change for other parsers, and is not in line with what openmetrics is going to do.

carlpett:
Oh, ok. We'll have to do some work-around then.
Could you explain how it would be breaking, and what the conflict is with openmetrics?

brian-brazil:
It'd be an expansion of what the text format is, so older parsers and parsers not implemented by us would break on this new format. OpenMetrics is the next version (roughly) and also only has \n.

PR #134, opened 2018-05-17
Fix type passed to fmt

brian-brazil:
Signed-off-by: Brian Brazil <brian.brazil@robustperception.io>


brian-brazil:
@discordianfish This fixes the build failure your PR is turning up.

PR #135, opened 2018-05-18
config: basic_auth doesn't need to require a username

adamdecaf:
Issue: https://github.com/prometheus/prometheus/issues/4173


brian-brazil:
Thanks!

Issue #117, opened 2018-01-30
model.ParseDuration bug


AlekSi:
`"1m0s"` can be parsed normally by `time.ParseDuration` (https://play.golang.org/p/QbEuk34UKSm) but not by `model.ParseDuration`:
```
not a valid duration string: "1m0s"
```

fabxc:
That's intended behavior. We do support days, weeks, and years (which stdlib `time` does not) but only support a single unit argument. This is not likely to change.

AlekSi:
That issue comes from a code working with Prometheus configuration file parsing and rendering, but not a part of Prometheus itself. We pass `string`s across API boundaries, not `time.Duration`, and both client (which uses `time.Duration` internally) and server (which uses `model.Duration`) should play nicely with each other. Will you accept a change if it is strictly backward-compatible with current implementation?

brian-brazil:
I believe we're all consistent in this these days, and as Fabian says it's not planned to extend this syntax.

PR #137, opened 2018-06-04
Support upstream time.ParseDuration (#117)

jmcfarlane:
This change supports model time.ParseDuration and is backwards
compatible with upstream.

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -359,7 +359,7 @@ func (p *TextParser) startLabelValue() stateFn {
 		}
 		return p.readingValue
 	default:
-		p.parseError(fmt.Sprintf("unexpected end of label value %q", p.currentLabelPair.Value))
+		p.parseError(fmt.Sprintf("unexpected end of label value %q", p.currentLabelPair.GetValue()))
 		return nil
 	}
 }
--- a/model/time.go
+++ b/model/time.go
@@ -182,6 +182,10 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 func ParseDuration(durationStr string) (Duration, error) {
 	matches := durationRE.FindStringSubmatch(durationStr)
 	if len(matches) != 3 {
+		// If single unit validation fails, still support time.ParseDuration
+		if d, err := time.ParseDuration(durationStr); err == nil {
+			return Duration(d), nil
+		}
 		return 0, fmt.Errorf("not a valid duration string: %q", durationStr)
 	}
 	var (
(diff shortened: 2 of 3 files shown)


brian-brazil:
I believe we have it this way on purpose, and this change would result in an inconsistent parser. Why should 1y2d not work but 1h2m does? What particular problem are you trying to solve.

Makefile changes should be handled separately up in the Prometheus repo first, as we've 30 repos and ad-hoc changes to the build infrastructure make maintenance challenging.

jmcfarlane:
I don't claim to understand the design decision behind supporting a custom `time.Duration` parser, but I do feel it violates the principle of least surprise. I first learned of the current behavior while testing https://github.com/coreos/prometheus-operator/pull/1333 which enables alert specification via Kubernetes [CRD](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/), where the `time.Duration` is typed as an `int64` due to [OpenAPI](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#dataTypes) type constraints. After validation (which is super useful) the `time.Duration` is converted to a `string` (presumably) via stdlib and results in Prometheus failing to start because the string value contains more than one unit (ie is a typical golang `time.Duration` string).

To answer your question specifically, I would say: The reason why 1h2m works but 1y2d not, is because the former is a standard golang `time.Duration` string format and the latter is a short hand single unit format supported by Prometheus for convenience, of which there are only three: [d, w, y].
 
Ack on the Makefile has been removed.

brian-brazil:
From memory the reason we have our own is that the Go one only goes up to hours, and it was annoying for users who regularly wanted to specify values in days/weeks/years. It took quite a while to even get the project consistently using this.

@brancz Sounds like you've a bug in the operator.

> To answer your question specifically

That's not a great answer if you want to avoid surprising users.

jmcfarlane:
As it happens a subsequent operator [commit](https://github.com/coreos/prometheus-operator/commit/afe9711ce2ef5dd0e0b5091f921ae9bc0aab3711#diff-939e2d86e2e0e8a4c6798bc787f524c4) switched from `int64` to `string`, though it's unclear how validation now works.

jmcfarlane:
I understand it's unrelated to Prometheus directly, but with the `int64` > `string` change validation now allows invalid `time.Duration` through:

```
caller=manager.go:479 component="rule manager" msg="loading groups failed" err="not a valid duration string: \"foobar\""
caller=main.go:607 msg="Failed to apply configuration" err="error loading rules, previous rule set restored"
```
where ^ successfully passed crd validation.

@brancz Happen to have a feeling on the right way to facilitate validation of the `for` field, but without surprising users with two parsing strategies for `time.ParseDuration`?

brian-brazil:
It sounds like this is obsolete, and I'm still not convinced this change would make sense from a UX perspective.

PR #138, opened 2018-06-16
Ensure configuration reload upon certificate change

roidelapluie:
Fixes https://github.com/prometheus/prometheus/issues/4155

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -14,12 +14,15 @@
 package config
 
 import (
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
I see what you're trying to do, but this is inconsistent with the rest of the configuration - we don't require you to have a hash of your password_file or bearer_token_file for those to be picked up automatically. It'd also be quite a bit of work for users, if they can do this then they can also have their configuration management system write out my_cert_file.HASH and change the config accordingly so overall this isn't a saving in effort it's just adding extra steps.

If this is to work, it needs to do so completely automatically.

roidelapluie:
@brian-brazil we do not require it. it is computed on unmarshalling.

roidelapluie:
For the user there is no difference.

roidelapluie:
(we are not adding any configuration option here)

brian-brazil:
This is the use configuration struct that you're changing, anything in there is a user configuration option. You need to do this deeper in the code.

brian-brazil:
And in addition, it should not require a reload to work in order to be consistent with password_file and bearer_token_file.

roidelapluie:
I was inspired by https://github.com/prometheus/prometheus/pull/4245/files where we also add an extra config option to ensure uniqueness.... (Source)

I find this quite elegant and it will work for all the prometheus project ootb

roidelapluie:
Thanks for your comments Brian, I will look how to do this is a better fashion

PR #139, opened 2018-06-16
Reload TLS certificates when needed

roidelapluie:
Fixes https://github.com/prometheus/prometheus/issues/4155

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strings"
 	"time"
 
@@ -146,6 +147,8 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTri
 		),
 	}
 
+	rt = NewTLSConfigRoundTripper(cfg, tlsConfig, name, rt)
+
 	// If a bearer token is provided, create a round tripper that will set the
 	// Authorization header correctly on each request.
 	if len(cfg.BearerToken) > 0 {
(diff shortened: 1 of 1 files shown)


config/http_config.go
```diff
@@ -162,6 +165,33 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTri
...
 }
 
+type tlsConfigRoundTripper struct {
+	cfg       HTTPClientConfig
+	tlsConfig *tls.Config
+	name      string
+	rt        http.RoundTripper
+}
+
+// NewTLSConfigRoundTripper reads the tls configuration upton each request
```
brian-brazil: upon

config/http_config.go
```diff
@@ -162,6 +165,33 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTri
...
+// and updates the TLS configuration as needed
```
brian-brazil: fullstop

config/http_config.go
```diff
@@ -162,6 +165,33 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTri
...
+func NewTLSConfigRoundTripper(cfg HTTPClientConfig, tlsConfig *tls.Config, name string, rt http.RoundTripper) http.RoundTripper {
+	return &tlsConfigRoundTripper{cfg, tlsConfig, name, rt}
+}
+
+func (rt *tlsConfigRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	tlsConfig, err := NewTLSConfig(&rt.cfg.TLSConfig)
+	if err != nil {
+		return nil, err
+	}
+	if !reflect.DeepEqual(tlsConfig, rt.tlsConfig) {
```
brian-brazil: I think there's mutable state in a tls.Config, so this may always return true and cause spurious reconnects

config/http_config.go
```diff
@@ -162,6 +165,33 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string) (http.RoundTri
...
+		rt.rt, err = NewRoundTripperFromConfig(rt.cfg, rt.name)
```
brian-brazil: This feels like a layering violation, it'd be cleaner to take in a function that'd create a new round tripper

roidelapluie:
I have ended by writing a http proxy that deals with that kind of things. Any other solution was too expensive (we do 8 scrapes/s with mutual ssl)

PR #141, opened 2018-06-29
Update HTTPConfig with KeepAlive setting.

RussellRollins:
In previous development of Prometheus, it has been possible to configure the KeepAlive settings of specific scrapes. However, that functionality was intentionally removed:

https://github.com/prometheus/prometheus/pull/3173

I'd be interested in reviving the discussion around this.

We attempt to scrape an endpoint that is behind a "black box" type load balancer, whose logic we don't really control. What we'd like to achieve is a roughly equal percent of the balanced resources receiving Prometheus requests to their `/metrics` endpoints. However, after some discussion with support on that product, we have determined that this won't work as long as the client (Prometheus in this case) is not sending `Connection: close` type requests.

The specific client I'd like to alter is the one created here:

https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/scrape/scrape.go#L145-L149

I apologize if this is a dead discussion, and we have some ideas to make other workarounds. However, I do think it's a useful case for Prometheus to cover, I can imagine other endpoints that work better without HTTP keep-alive.

@fabxc

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -74,6 +74,8 @@ type HTTPClientConfig struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
+	// If set, override whether to use HTTP KeepAlive - defaults ON
+	KeepAlive *bool `yaml:"keep_alive,omitempty"`
 }
 
 // Validate validates the HTTPClientConfig to check only one of BearerToken,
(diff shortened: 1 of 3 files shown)


brian-brazil:
This config is used in more than one place, and for this to be considered to be made configurable it must make sense for it to be configurable in all of those places.

Prometheus is expected to scrape individual targets directly, and going via a load balancer almost always causes problems. If you really need this I'd suggest using a load balancer that supports per-request rather than per-connection load balancing, and failing that you could always put a proxy in a way that closes the connections.

RussellRollins:
Hi @brian-brazil, thanks for the feedback. We have some other solutions to look at, but we wanted to make sure this was a deliberate design decision Prometheus side, not just an oversight. Thanks!

PR #142, opened 2018-08-03
Ignore UTF8 BOM

sgissi:
Add requested tests to PR #42
Signed-off-by: Silvio Gissi <silvio@gissilabs.com>

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -125,6 +125,11 @@ func (p *TextParser) reset(in io.Reader) {
 	} else {
 		p.buf.Reset(in)
 	}
+	// Some clients might throw in UTF-8 BOM chars, ignore if found.
+	maybeBOM, err := p.buf.Peek(3)
+	if len(maybeBOM) == 3 && maybeBOM[0] == 0xEF && maybeBOM[1] == 0xBB && maybeBOM[2] == 0xBF && err == nil {
+		p.buf.Discard(3)
+	}
 	p.err = nil
 	p.lineCount = 0
 	if p.summaries == nil || len(p.summaries) > 0 {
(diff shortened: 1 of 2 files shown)


simonpasquier:
LGTM

grobie:
@brian-brazil @RichiH Is this in line with the open metrics work? Does it still make sense to gracefully handle BOM?

RichiH:
It's currently not in the spec and requiring BOM to make certain the data is UTF-8 would have impact on other users.

That being said, being strict about sending out, flexible in receiving, etc. As it's non-printable, it's the perfect trap for implementers who don't test against the OpenMetrics test suite, though...

All in all, I would lean towards explicitly forbidding it, but I am not 100% convinced either way.

grobie:
Alright, I propose we close this and the other PR, and add a note to https://prometheus.io/docs/instrumenting/exposition_formats/ that BOM is explicitly forbidden following https://tools.ietf.org/html/rfc3629.

> A protocol SHOULD forbid use of U+FEFF as a signature for those textual protocol elements that the protocol mandates to be always UTF-8, the signature function being totally useless in those cases.

wdyt?

brian-brazil:
> I would not see this is a breaking change considering that it does nothing if there is no UTF8 BOM and ignores it if is exists.

It takes something which is currently invalid and makes it valid, thus all other parsers would need to be updated. That's a breaking change.

RichiH:
The problem is that others might not accept it and then error out when
Prometheus is happy.

Making this a hard error is better.

Sent by mobile; please excuse my brevity.

sgissi:
@brian-brazil @RichiH Fair points, I'll update the branch with an hard error instead.

sgissi:
@brian-brazil @RichiH PTAL. Let me know if the BOM checks are too bloated to live inside TextToMetricFamilies, if so I can move them to a separate function.

sgissi:
Seems like I missed the sign-off on the last commits, sorry about that.
@brian-brazil @RichiH Is this issue still relevant? I can rebase and sign the commit if there is interest on merging it, otherwise I recommend we close all issues/PRs related with a comment that UTF BOMs doesn't need special handling.

brian-brazil:
The current decision is that we won't be allowing this, as OpenMetrics forbids it.

PR #148, opened 2018-10-14
 Make MetricFamilyToText allocation-free

### LINKED ISSUE #146: Optimize `MetricFamilyToText`
This [code](https://github.com/prometheus/common/blob/master/expfmt/text_create.go) has been written with the assumption it would never be a in a hot code path. With Prometheus 2.x dropping protobuf support, scrapes by Prometheus are now always using the text format, and thus every monitored Go program is currently running this code quite a lot. (@mxinden has just told me how the kube-state-metrics CPU profiling is dominated by [`labelPairsToText`](https://github.com/prometheus/common/blob/master/expfmt/text_create.go#L243).)

client_golang will undergo more invasive changes in the future, which will provide opportunities to solve the problem in another way, but `MetricFamilyToText` has a lot of low hanging fruit to make the current usage of client_golang more efficient.

beorn7:
@mxinden as promised.
@grobie we should run a kube-state-metrics instance with this vendored in.
Fixes #146.

This makes use of interface upgrades of the `io.Writer`. If the
necessary upgrades are not available, a `bytes.Buffer` is used in
between.

Number formatting is performed with the `strconv.Append...` functions,
using an intermediate byte slice as buffer. (Perhaps surpsingly, a
plain `fmt.Fprint` with an int or float as single argument causes
quite a bit of allocations, despit the usage of `sync.Pool` by the
`fmt` package.)

Both the `bytes.Buffer` and the intermediate `[]byte` buffer are
managed by a `sync.Pool` each.

Benchmark comparison:

BEFORE
```
$ go test -bench=Create -run=NONE -benchmem
goos: linux
goarch: amd64
BenchmarkCreate-4                  50000             23854 ns/op            4433 B/op        285 allocs/op
BenchmarkCreateBuildInfo-4       1000000              1592 ns/op             216 B/op         19 allocs/op
```

AFTER
```
$ go test -bench=Create -run=NONE -benchmem
goos: linux
goarch: amd64
BenchmarkCreate-4                 100000             12644 ns/op               0 B/op          0 allocs/op
BenchmarkCreateBuildInfo-4       2000000               892 ns/op               0 B/op          0 allocs/op
```

Bottom line: About twice as fast, and no allocations at all in the long run.

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -14,13 +14,44 @@
 package expfmt
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"math"
-	"strings"
+	"strconv"
+	"sync"
+
+	"github.com/prometheus/common/model"
 
 	dto "github.com/prometheus/client_model/go"
-	"github.com/prometheus/common/model"
+)
+
+// enhancedWriter has all the enhanced write functions needed here. bytes.Buffer
+// implements it.
+type enhancedWriter interface {
+	io.Writer
+	WriteRune(r rune) (n int, err error)
+	WriteString(s string) (n int, err error)
+	WriteByte(c byte) error
+}
+
+const (
+	initialBufSize    = 512
+	initialNumBufSize = 24
+)
+
+var (
+	bufPool = sync.Pool{
+		New: func() interface{} {
+			return bytes.NewBuffer(make([]byte, 0, initialNumBufSize))
+		},
+	}
+	numBufPool = sync.Pool{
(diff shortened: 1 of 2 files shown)


fabxc:
@beorn7 have you seen https://godoc.org/golang.org/x/tools/cmd/benchcmp?
It's pretty nice for comparing benchmark outputs.

beorn7:
I know the tool, but I found it a bit over-tech'd for the six deltas involved here, especially given that four of them went down to zero...

PR #149, opened 2018-10-19
fix: use initialBufSize

orisano:
Is it expected? I think that it is a mistake.


brian-brazil:
Yeah, that looks to be the case. @beorn7

beorn7:
Yes. Thanks for spotting.

PR #150, opened 2018-10-19
feat: use strings.Replacer

orisano:
Signed-off-by: Nao YONASHIRO <yonashiro@r.recruit.co.jp>

improve performance.
```
name               old time/op  new time/op  delta
Create-4           13.1µs ± 2%   8.7µs ±10%  -33.99%  (p=0.000 n=9+10)
CreateBuildInfo-4   920ns ± 1%   422ns ±10%  -54.12%  (p=0.000 n=9+10)
```

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"math"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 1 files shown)


beorn7:
The problem with the `strings.Replacer` is that it causes allocs, see https://golang.org/src/strings/strings.go?s=22223:22269#L840

The runtime gain is significant, but allocations can bite us in different ways.

I think, however, most of the speed gain happens because most strings don't need escaping at all and could just be copied directly. Thus, we could try to use the same strategy as in `strings.Replacer`, i.e. first check if escaping is needed at all, and if not, just call `w.WriteString` directly. That should have an even better effect than the one in this PR.

beorn7:
Sorry, misread `Replace` vs. `Replacer`.

I now think this is doing exactly the right thing. Will just test for allocs to be sure onec I'm at my development computer.

beorn7:
Confirmed allocation-free-ness. This is exactly what I was looking for but failed to spot in the `scrings` package. Many thanks for this.

Compared to the original benchmark in #148, this is a reduction in runtime of 54% for `Create` and `67% for `CreateBuildInfo`. Really nice.

Issue #147, opened 2018-10-10
Update MAINTAINERS.md


beorn7:
Current MAINTAINERS.md lists @fabxc as the maintainer, but with his CoreOS mail address (which presumably wouldn't work anymore as he doesn't work for CoreOS anymore).

There has been very little commit activity an this repo recently, but there are a bunch of open PRs, new and old, and a lot of meaningful issues, again new and old. Most of recent activities on issues and PRs is (unsurprisingly) by @brian-brazil . The current maintainer doesn't seem to have shown any recent activity here.

In short: This repo appears abandoned. However, it is used by many others across the project, thus having this properly maintained is crucial. Many of the changes here, and in particular some of the long-standing open issues, touch quite basic decisions that have impact on the project as a whole and need consistency across the project (including the meta issue of what should go into `common` at all – the different opinions about this have been colliding here and there over the last years, up to the current day… #97 is just the tip of the iceberg). I would love to see somebody who is not only actively maintaining this but also has the ability to mediate all those basic decisions into a productive consensus.

juliusv:
@fabxc Could you provide an updated email address? I assume you would be fine with a maintainership change if someone volunteered (which doesn't seem like the case right now)?

fabxc:
I advocated for phasing out this repo as much as possible, others argued for the opposite and more functionality has been added since. I understand arguments for either and don't feel too strongly but can't really be a maintainer for things I didn't add and wouldn't agree to add.

I think we discussed this to the same extent some time ago on IRC. I suggest @brian-brazil takes maintainership here since he was the main advocate for centralising config structs etc. here.

juliusv:
@fabxc Makes sense, I'll send a PR with @brian-brazil as the maintainer and see if he complains :)

PR #152, opened 2018-10-26
Add http_config option for unprefixed (api) token authentication.

0xDiddi:
As of now, the`http_config` only accepts OAuth bearer tokens and basic user+password authentication.

With these changes, you can specify an `api_token` or an `api_token_file` that will be used as the authentication header without any further prefixes, suffixes or other modifications.

This PR should (unless I've missed something) contain all required bits and pieces, including validation and unit tests.

cc @fabxc
as per #147 also cc @beorn7


brian-brazil:
Thanks for your PR. I'm not sure we should support arbitrary non-standard authentication mechanisms, as that's getting a bit too niche.

hphde:
What's really required is something that satisfies https://tools.ietf.org/html/rfc2616#page-107
Some part that sends out `Authorization: xyz` where as `xyz` can be anything like `String` or `Identifier String` or whatever you can think of.

hphde:
@brian-brazil what do you say about rfc2616?

brian-brazil:
I say that we comply with rfc2616, that doesn't mean we have to support every possible implementation variation of it though.

hphde:
That's true, current Bearertoken implementation is compliant with the RFC but the PR satisfies it to a much broader extend. I don't know if it's an implementation of all variations - that would be awesome, since it's quite a small change.

brian-brazil:
I'm going to close this off, as we still only support standard bearer tokens. We explicitly aren't trying to support every bespoke auth system that has been created, and you can always use a http proxy if you need something niche.

PR #153, opened 2018-11-01
add wildcard label name

kirillsablin:
I need this change to proceed with https://github.com/prometheus/alertmanager/pull/1588


kirillsablin:
@fabxc

brian-brazil:
This is not a valid labelname, this implementation detail of the AM should not leak elsewhere.

PR #156, opened 2018-11-16
promlog: add comment for Config

alexander-yu:
Forgot to add a comment for the Config struct in #136; should satisfy any lint issues for people.


brian-brazil:
Thanks!

PR #157, opened 2018-11-16
*: support Go modules

simonpasquier:
This also enables additional checks in Travis CI: style, staticcheck and license headers.


brian-brazil:
:+1:

go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/prometheus/common
+
```
SamWhited: Consider adding a language line (if you recreate this with a recent version of Go built from source it will add one for you, so we might as well go ahead and add it so that it doesn't have to be done later). The earliest version of Go that has basic support for reading the `go.mod` file is 1.9.7 so something like `go 1.9` is enough.

simonpasquier: Sorry do you have a link to some documentation? I've never seen such things.

SamWhited: Like most things about modules, it's not documented (*sigh*) but if you create the go.mod file with a newer version of Go 1.12 (I forget when it was added; it will be in the next release though so building from source should do it) it will add it. I can try to dig up the list discussion, if I can find it (I can never find anything searching the list archives) I'll paste it here.

SamWhited: My apologies, it is documented, I just never found it: https://tip.golang.org/cmd/go/#hdr-The_go_mod_file

simonpasquier: I've dug into the Git history and found this [CL](https://go-review.googlesource.com/c/go/+/125940/) which says:

```
cmd/go: add 'go version' statement in go.mod

We aren't planning to use this or advertise it much yet,
but having support for it now will make it easier to start
using in the future - older go commands will understand
what 'go 1.20' means and that they don't have go 1.20.
```

Personally I wouldn't add it until it is "officially" announced. But many thanks for the tip anyway!

SamWhited: I'm sorry, I forgot about this, but ended up finding the documentation: https://tip.golang.org/cmd/go/#hdr-The_go_mod_file

The problem is that it's already in 1.12 if you build it from source or use beta1, so any time someone builds this library with those versions it adds this line and you now have uncommitted changes in the repo.

SamWhited: Ah yah, interesting, they must have either disabled that feature or added it after the beta was cut. When I'm at my laptop I'll check if I'm on a version before or after the beta.

Either way, it exists in the Go tool, and new files created with the beta definitely add it by default, so we should probably be up to date with what `go1.12beta1 mod init` does.

beorn7: Perhaps we should just wait until the feature has made it at least into an RC? At beta stage, things might still change, and updating it here later is not a big deal. (FYI: Most prometheus repos have been converted already, without the language line, anyway.)

SamWhited: It won't change from beta; it's just bug fixes from here on in.

SamWhited: It's not; I can dig up the commit where it was added if you really want. Just trying to save having to make changes later and users who build this later suddenly having uncommitted changes all the time.

beorn7: You are missing my point. 

My point is that the effort of this discussion is not worth the saving it can possibly create.

To prove my point, I'll drop out of this discussion now. @simonpasquier should make the call. I'l :+1: whatever he decides.

simonpasquier: I'd agree with @beorn7 and wait for 1.12 to be officially out before bothering.
Anyway thanks for the hints @SamWhited!

PR #160, opened 2018-11-25
Add unit tests to increase coverage for encode and decode

tariq1890:
Adding to the unit test coverage for decode and encode.

### DIFF
--- a/expfmt/decode_test.go
+++ b/expfmt/decode_test.go
@@ -259,6 +259,65 @@ func TestProtoDecoder(t *testing.T) {
 			},
 		},
 		{
+			in: "\u007f\n\x1drequest_duration_microseconds\x12\x15The response latency.\x18\x04\"E:C\b\x85\x15\x11\xcd\xcc\xccL\x8f\xcb:A\x1a\v\b{\x11\x00\x00\x00\x00\x00\x00Y@\x1a\f\b\x9c\x03\x11\x00\x00\x00\x00\x00\x00^@\x1a\f\b\xd0\x04\x11\x00\x00\x00\x00\x00\x00b@\x1a\f\b\xf4\v\x11\x9a\x99\x99\x99\x99\x99e@",
+			expected: model.Vector{
+				&model.Sample{
+					Metric: model.Metric{
+						model.MetricNameLabel: "request_duration_microseconds_count",
+					},
+					Value: 2693,
+					Timestamp:testTime,
+				},
+				&model.Sample{
+					Metric: model.Metric{
+						"le": "+Inf",
+						model.MetricNameLabel: "request_duration_microseconds_bucket",
+					},
+					Value: 2693,
+					Timestamp:testTime,
+				},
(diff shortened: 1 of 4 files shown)


expfmt/decode_test.go
```diff
@@ -258,6 +258,65 @@ func TestProtoDecoder(t *testing.T) {
 				},
 			},
 		},
+		{
+			in: "\u007f\n\x1drequest_duration_microseconds\x12\x15The response latency.\x18\x04\"E:C\b\x85\x15\x11\xcd\xcc\xccL\x8f\xcb:A\x1a\v\b{\x11\x00\x00\x00\x00\x00\x00Y@\x1a\f\b\x9c\x03\x11\x00\x00\x00\x00\x00\x00^@\x1a\f\b\xd0\x04\x11\x00\x00\x00\x00\x00\x00b@\x1a\f\b\xf4\v\x11\x9a\x99\x99\x99\x99\x99e@",
```
brian-brazil: @beorn7 Looking at the spec, we just say "whitespace" rather than tabs and spaces, so we should accept other whitespace characters. Does this seem right to you?

beorn7: It's probably safer to make the spec more precise/restrictive by explicitly only allow spaces and tabs (as it is implemented right now in `expfmt`). Interpreting whitespace widely would include all kind of line-break characters, which we could definitely not allow. But even for non-line-breaking whitespace, there are a _lot_ of whitespace characters in Unicode, which would make parsing much harder if we needed to check for all of them. See https://en.wikipedia.org/wiki/Whitespace_character#Definition_and_ambiguity

beorn7: In different news: You commented the line with the raw protobuf. Where is whitespace involved in that one? Perhaps I'm missing your point.

expfmt/encode_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	err := delimEncoder.Encode(metric)
+	if err != nil {
+		t.Errorf("unexpected error during encode: %s", err.Error())
+	}
+
+	out := buff.Bytes()
+	if len(out) == 0 {
+		t.Errorf("expected the output bytes buffer to be non-empty")
```
beorn7: This could test the actual content. Same below where only non-emptiness as asserted.

expfmt/decode_test.go
brian-brazil: Brainfart on my part, I though this was an exotic text format.

brian-brazil:
Thanks!

expfmt/decode_test.go
beorn7: But still a good point that we mean “tab and space” when we say "whitespace”. We should clarify (or wait for OpenMetrics to clarify...).

brian-brazil: OM has only the space character, probably not worth clarifying for Prometheus format at this point.

PR #162, opened 2018-12-06
Ensure there's always a . or e in floats.

brian-brazil:
This is for future consistency with OpenMetrics, so we can fully distinguish
ints and floats.

This has no impact on allocs, and is a 5% hit on CPU.

@SuperQ @beorn7

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -436,11 +436,11 @@ func writeEscapedString(w enhancedWriter, v string, includeDoubleQuote bool) (in
 func writeFloat(w enhancedWriter, f float64) (int, error) {
 	switch {
 	case f == 1:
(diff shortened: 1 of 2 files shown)


beorn7:
Implementation-wise, this looks fine.

However, the buckets of a histogram and the _count of a summary and a histogram _are_ ints. We render them as floats with this change.

brian-brazil:
I was thinking that's something we can improve with the Go OpenMetrics text creator, as it has that notion whereas the Prometheus text format is all floats.

beorn7:
Then go ahead with this, if it has to happen.

I still remember how I wrote into the code how badly suited the text format is for summaries and histograms.

The frequency of "told you so" moments is increasing recently...

PR #166, opened 2019-01-07
Add public setter for JSONFormatter

toddg-remitly:
Exposing a public setter for the JSONFormatter enables setting custom FieldMap values.


brian-brazil:
May I ask why you want this? This code is also deprecated in favour of promlog.

brian-brazil:
There's no need to use our internal logging library if you're just using Prometheus instrumentation. There may be a case if you're writing an exporter or some deeper integration with our libraries.

brian-brazil:
As above, this doesn't make sense and is for deprecated code.

PR #168, opened 2019-01-23
Revert "Ensure there's always a . or e in floats. (#162)"

beorn7:
This reverts commit 67670fe90761d7ff18ec1d640135e53b9198328f.

The format change wreaks havoc with histograms if you have a mixed set
of monitored targets. `histogram_quantile` spits out weird values,
which are difficult to diagnose.

A developer team here at SoundCloud pulled in the current version of
prometheus/common via a different indirect dependency than
prometheus/client_golang. (client_golang alone would not pull it in
yet if you are using Go modules.) Canary instances then had the newer
prometheus/common, while the normal production instances had not. The
calculated quantiles for the complete service jumped up dramatically,
so the team rolled back and started to look for a performance
regression. (Just looking at the canary alone still worked, but since
nobody suspected this kind of monitoring failure case, the
investigation went totally the wrong way.)

Given the wide distribution of prometheus/client_golang and the way Go
modules work, we will see many of those innocuous upgrades that
suddenly change the `le` values on new deployments. Since this change
is buried behind dependencies, users will run into this problem
without suspecting it, even if we announce it very loud and clearly in
prometheus/common or even prometheus/client_golang.

For now, we have to revert the change and then think about a way to
mitigate it. I'm thinking of sanitizing `le` values in Prometheus
2.x. But let's think of it without haste.


brian-brazil:
This is going to break either now, or when client_golang starts negotiating and serving OpenMetrics. This an expected outcome (as is any change to buckets) and I don't see any sane options to mitigate this, so I think we should leave this as-is to reduce the overall impact.

beorn7:
Neither can this be compared to a controlled change of the bucketing. In that case, the developer changes the bucketing, and then the bucket change.
Here, however, the buckets change because of seemingly unrelated changes that do not touch the bucketing scheme.

beorn7:
This would not have broken with protobuf. And even with the text format, this would not have broken Prometheus 1.x. There is clearly a series of ill-advised design decisions at work here. We should rather try to fix it while we still can instead of just going on with business as usual.

matthiasr:
It did not specify it, but in practice there has always been exactly one way that it was done. To me this means there was an oversight in the specification, not that we are free to change it however we want whenever we want.
As @beorn7 stated, the Prometheus server could collapse these values. If we don't want to do it magically, we can make a new exposition format – we have all the mechanisms for version negotiation that we need. We can simply _not change it_ and wait for OpenMetrics until we solve this. We can go through with the change, but come up with a communications and rollout plan, rather than letting everyone run into this whenever they run into it. There are many options, but to work out which one is the best we need time.

brian-brazil:
> It did not specify it, but in practice there has always been exactly one way that it was done.

That's not true, the Python and Java clients at least did/do different things.

> To me this means there was an oversight in the specification

Agreed, which this is trying to rectify.

> As @beorn7 stated, the Prometheus server could collapse these values.

This would break non-Go users, as how their histograms are ingested would change.

> If we don't want to do it magically, we can make a new exposition format – we have all the mechanisms for version negotiation that we need. 

I'm not seeing how that could help, that would result in the exact same problem.

> We can simply not change it and wait for OpenMetrics until we solve this.

This is intended to handle this before OpenMetrics rolls out, as this makes OpenMetrics a noop in this regard.

> come up with a communications and rollout plan, rather than letting everyone run into this whenever they run into it.

That sounds good, it has already been checked in for a month.

brian-brazil:
I should mention too that exactly how this is handled is still not 100% defined on the OpenMetrics side, we're keeping an eye out to see more if having the minimal representation is sane in enough languages to continue to have a MUST. Our next meeting is next week, and I will bring this up.

beorn7:
When I grudgingly agreed to this change in https://github.com/prometheus/common/pull/162#issuecomment-444948291 , I assumed we can add a big and fat BREAKING CHANGE warning to the next release of client_golang. What I didn't think of is how easily people might vendor in a newer version of prometheus/common (because they just build from head always or because they use Go modules and happen to include another dependency that declared a more recent version of prometheus/common for some reason – FTR: Go modules would have handled this fine if we had major versions in prometheus/common and bumped that with that change).

I now realize that this breaks user in very non-obvious ways, which is why I think we should revert this change for as long as it has not been referred to by a client_golang release. Then we should think about a transition plan. This could be that Prometheus 2 sanitizes the `le` label value to the value that Python and Java uses (so that nothing changes for Python and Java, where no change is planned). This is manageable for users because it's a controlled change where you can upgrade your Prometheus server with the knowledge that graphs and alerts will break for a while. You can silence alerts and all that.

With the way this has been done so far, users will accidentally put this change into clients and break graphs and trigger alerts without any warning. Also, with the wide spread of Go instrumentation, in any larger organizations, both formats will be mixed for a while, especially in infrastructure components, where you might want to aggregate across monitored targets of different versions for a while.

If Prometheus itself sanitized the `le` labels to the desired future format, this would not matter, and Go targets could be gradually upgraded without friction.

The above shows that a smooth transition is possible. We can flesh out details later, but I strongly recommend to merge this PR immediately to avoid more surprise damage

brian-brazil:
> This could be that Prometheus 2 sanitizes the le label value to the value that Python and Java uses (so that nothing breaks for Python and Java users, where no change is planned). 

Python (pre OM), Java, Go, and OpenMetrics all have slightly different outputs unfortunately.

matthiasr:
> I'm not seeing how that could help, that would result in the exact same problem.

It would not affect them, until the non-Go libraries increment to the new version with the defined expectation.

> it has already been checked in for a month

Can you point me at the plan? I think I missed it.

matthiasr:
> all have slightly different outputs unfortunately

for me, that speaks in favor of _first_ settling on one format, _then_ planning the transition, and _then_ implementing it.

juliusv:
To me it seems obvious that we cannot just (intentionally) break this in such an unexpected manner.

It's one thing for client libraries in different languages to have had different bucket formats since forever, it's a totally different thing to break it within one client library and in a way that is very likely to go unnoticed until it's in production (and even then, you run into the situation that at first you think it's not a Prometheus problem, but that your service is acting up).

brian-brazil:
> I don't understand how? Wouldn't we be coercing the Go behavior to match theirs?

No, the only other client that currently produces this is the recent Python one that's the OM reference implementation. The other clients (Java anyway, haven't dug too much into the rest) produce different output.

beorn7:
To not let your vision get clouded by certain statements here:
1. Coercing the value of the `le` label into something consistent is what Prom1 did. If Prom2 did it, too, it would solve the problem, and the change could be applied in a controlled and centralized fashion.
2. Item 1 doesn't even matter because this PR is about limiting the damage of a breaking change. I explicitly said that this has to be merged immediately. We can then think about the best further course of action once the damage is contained. Item 1 was mostly intended to prove that there are options.
3. I explicitly called out the aspects I did not take into account when I (grudgingly) agreed to the change that this PR is reverting. Thus, my agreement back then is void and does not block the revert in any way.

juliusv:
@brian-brazil It seems there are more voices here for reverting than against. However, you are the maintainer. Are you ok with reverting this for now, until a more careful way forward is agreed upon? Otherwise it may have to go to a formal team vote.

brian-brazil:
I'm currently waiting on input from other OpenMetrics people. It's already been in a month, we can wait a little bit to see if there's better options rather than making a breaking change into three when this gets rolled back and then rolled forward again.

beorn7:
Note that this change hasn't been released as part of `client_golang` yet. This change seeps in gradually as people update dependencies, following whatever dependency management strategy they are applying. Thus, as long as we don't revert, we put the huge number of users at risk that haven't pulled this in yet, while a role back now only affects the very few that have already gone there.
Brian's concern about "three breaking changes" applies the more the longer we wait. Thus, I pledge to rollback immediately.
Sadly, I have to stop talking now and return to my day job. At SC, we know what to look out for now. We are throwing everybody else under a bus right now. Please make a call.

brian-brazil:
:+1: 

Had a brief chat with OM people, noone bit on the notion of changing the spec or proposed anything that'd get in the way of an ultimate solution. The proposal is to revert this now, and reintroduce it when OM support is added to client_golang.

beorn7:
I'll write a mail to prometheus-users@ in case somebody got hit by this.

brian-brazil:
A mitigation option just occurred to me, we could have `histogram_quantile` coalesce buckets with equivalent `le` labels. Doesn't help with people directly accessing bucket values via a selector (e.g. apdex), but is non-breaking and help things generally.

juliusv:
@matthiasr It does, mostly. It ignores the `le` label for the hash-grouping and then parses the `le` label as a float, where the formatting doesn't matter. The problem is that the function will see two incoming time series which need to be treated as equivalent and added up:

```
foo{le="1"} 5
foo{le="1.0"} 7
```

brian-brazil:
https://github.com/prometheus/prometheus/pull/5158 for that idea.

PR #169, opened 2019-02-01
Return CloseIdleConnections with client to allow cleanup ability

mknapphrt:
This will return the CloseIdleConnections function along with the http client so that connections can be cleaned up later on when they are no longer needed.

Depended on by: https://github.com/prometheus/prometheus/pull/5144

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -111,26 +111,46 @@ func newClient(rt http.RoundTripper) *http.Client {
 	return &http.Client{Transport: rt}
 }
 
+type ClientWithCloser struct {
+	*http.Client
+	closer func()
+}
+
 // NewClientFromConfig returns a new HTTP client configured for the
 // given config.HTTPClientConfig. The name is used as go-conntrack metric label.
-func NewClientFromConfig(cfg HTTPClientConfig, name string) (*http.Client, error) {
+func NewClientFromConfig(cfg HTTPClientConfig, name string) (*ClientWithCloser, error) {
 	rt, err := NewRoundTripperFromConfig(cfg, name)
 	if err != nil {
 		return nil, err
 	}
-	return newClient(rt), nil
+	client := &ClientWithCloser{newClient(rt), rt.CloseIdleConnections}
+	return client, nil
+}
+
+type roundTripperWithCloser interface {
+	http.RoundTripper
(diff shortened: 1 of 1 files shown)


config/http_config.go
```diff
@@ -113,24 +113,24 @@ func newClient(rt http.RoundTripper) *http.Client {
 
 // NewClientFromConfig returns a new HTTP client configured for the
 // given config.HTTPClientConfig. The name is used as go-conntrack metric label.
-func NewClientFromConfig(cfg HTTPClientConfig, name string) (*http.Client, error) {
-	rt, err := NewRoundTripperFromConfig(cfg, name)
+func NewClientFromConfig(cfg HTTPClientConfig, name string) (*http.Client, func(), error) {
```
brian-brazil: I was thinking rather than changing the return values, to instead have an interface that includes http.Client and also the new method. That way it's there for those who need it, and doesn't pollute the API usage for others.

mknapphrt:
I tried to make it as unobtrusive as I could. The methods only return extended interfaces or structs so there should be no loss of functionality.

mknapphrt:
Didn't realize just `make` didn't run the tests :/ So I'm not sure what the best way to handle this is, because the test cases for `basicAuthRoundTripper` try to assert the type of `client.Transport` as a `basicAuthRoundTripper` but the round tripper is embedded in `rtWithCloser` and that assertion fails.

I could modify `basicAuthRoundTripper`, `bearerAuthRoundTripper`, and `bearerAuthFileRoundTripper` so that they implemented `roundTripperWithCloser`, but that would require modifying the signature of their "constructors", which I was trying to avoid.

Unless it's ok for `NewRoundTripperFromConfig` to return the `CloseIdleConnections` method as a third return value and only the `NewClientFromConfig` method needed to be changed to include the http.Client and new method.

config/http_config.go
```diff
@@ -111,26 +111,46 @@ func newClient(rt http.RoundTripper) *http.Client {
 	return &http.Client{Transport: rt}
 }
 
+type ClientWithCloser struct {
+	*http.Client
+	closer func()
```
brian-brazil: I'd expect this to be public

brian-brazil:
I believe this is obsolete.

PR #172, opened 2019-02-17
Use fixed length millisecond timestamp format for logs

sylr:
Port https://github.com/prometheus/prometheus/pull/5225 over to common

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -18,10 +18,21 @@ package promlog
 
 import (
 	"os"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
+)
+
+var (
+	// This timestamp format differs from RFC3339Nano by using .000 instead
+	// of .999999999 which changes the timestamp from 9 variable to 3 fixed
+	// decimals (.130 instead of .130987456).
+	timestampFormat = log.TimestampFormat(
+		func() time.Time { return time.Now().UTC() },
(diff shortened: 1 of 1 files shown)


promlog/log.go
```diff
@@ -18,12 +18,23 @@ package promlog
...
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 )
 
+var (
+	timestampFormat = "2006-01-02T15:04:05.000Z07:00"
+)
+
+func init() {
```
brian-brazil: You're still affecting all users of the library here. Can you make it so it only affects our loggers?

sylr: I don't think so, the global var is the only way to specify the timestamp format in the go-kit/log lib.

brian-brazil: You can change it 3 lines from the bottom of this file

sylr: Indeed

promlog/log.go
```diff
@@ -18,12 +18,20 @@ package promlog
...
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 )
 
+var (
+	timestampFormat = log.TimestampFormat(
+		func() time.Time { return time.Now().UTC() },
+		"2006-01-02T15:04:05.000Z07:00",
```
brian-brazil: You should probably add a comment as to how this is different to the usual

sylr: Done

brian-brazil:
Thanks!

PR #173, opened 2019-02-28
config: reload cert files from disk automatically

simonpasquier:
Fix for https://github.com/prometheus/prometheus/issues/4155.

cc @brian-brazil @brancz


brian-brazil:
This would only apply on reload, it shouldn't require that to be in line with how the other auth files work.

brancz:
@brian-brazil Are you saying the other file references need to be handled as well, or in the same way? As far as I can tell the only other one in the http config is the `bearerTokenFile`, which is already handled by re-reading on every scrape. Not objecting, just trying to understand what you meant.

brian-brazil:
The other ones are already handled this way.

simonpasquier:
@brian-brazil @brancz can you have a look again? FWIW I've tested it locally.
This PR is using [`tls.Config.GetClientCertificate`](https://golang.org/pkg/crypto/tls/#Config) which is only available since Go 1.8. I've assumed that the change doesn't have to be made for earlier versions of Go so I've added `// +build go1.8` directives. If this is wrong, I could amend the PR to remove this "limitation".

brian-brazil:
> I've assumed that the change doesn't have to be made for earlier versions of Go so I've added // +build go1.8 directives. If this is wrong, I could amend the PR to remove this "limitation".

This won't be used by client_golang, so that seems a safe assumption. FYI @beorn7

config/http_config.go
```diff
@@ -124,42 +129,51 @@ func NewClientFromConfig(cfg HTTPClientConfig, name string) (*http.Client, error
...
-	// Authorization header correctly on each request.
-	if len(cfg.BearerToken) > 0 {
-		rt = NewBearerAuthRoundTripper(cfg.BearerToken, rt)
-	} else if len(cfg.BearerTokenFile) > 0 {
-		rt = NewBearerAuthFileRoundTripper(cfg.BearerTokenFile, rt)
-	}
 
-	if cfg.BasicAuth != nil {
-		rt = NewBasicAuthRoundTripper(cfg.BasicAuth.Username, cfg.BasicAuth.Password, cfg.BasicAuth.PasswordFile, rt)
+	if len(cfg.TLSConfig.CAFile) == 0 {
```
brian-brazil: I'm not getting what you're trying to do here.

config/http_config.go
```diff
@@ -308,6 +320,116 @@ func (c *TLSConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	t.mtx.RLock()
+	equal := bytes.Equal(h[:], t.hashCAFile)
+	rt := t.rt
+	t.mtx.RUnlock()
+	if equal {
+		// The CA cert hasn't changed, use the existing RoundTripper.
+		return rt.RoundTrip(req)
+	}
+
+	// Create a new RoundTripper.
```
brian-brazil: We should probably close idle connections here.

config/http_config.go
simonpasquier: If no CA file is provided, we don't need a round tripper that reloads the CA. So we return a "normal" round-tripper.

brian-brazil: What if the root changes?

beorn7:
> This won't be used by client_golang, so that seems a safe assumption.

And even if we did at some point in the future, I see little problems raising the min requirement for client_golang to 1.8. Those are ancient versions, and while I think we should support them as long as it doesn't cause any overhead, we should not commit to more than that.

config/http_config.go
simonpasquier: Which root?

config/http_config.go
```diff
@@ -195,6 +199,12 @@ func (rt *bearerAuthRoundTripper) RoundTrip(req *http.Request) (*http.Response,
 	return rt.rt.RoundTrip(req)
 }
 
+func (rt *bearerAuthRoundTripper) CloseIdleConnections() {
+	if ci, ok := rt.rt.(closeIdler); ok {
```
brian-brazil: This is cleaner in Go 1.12 as it exposes this directly, without having to reach into the transport

config/http_config.go
brian-brazil: I meant the CA (as that's usually the root), but that would miss the client ssl auth changing.

simonpasquier: Client cert and key files are handled differently: we leverage [`tls.Config.GetClientCertificate`](https://golang.org/pkg/crypto/tls/#Config) so they are always read from disk.

https://github.com/prometheus/common/pull/173/files#diff-a2356a3b837239d300d6a0326a452aafR319

https://github.com/prometheus/common/pull/173/files#diff-a2356a3b837239d300d6a0326a452aafR345

config/http_config.go
simonpasquier: You refer to `client.CloseIdleConnections()`? I don't know how we can avoid going through the nested round-trippers until we call `Transport.CloseIdleConnections()`.

https://golang.org/src/net/http/client.go?s=27593:27632#L841

brancz: Have we even switched over to go 1.12 everywhere? I feel like it might be a little early to use that. I think I'd prefer this approach at this point.

brancz:
lgtm :+1:

roidelapluie:
@simonpasquier this is so awesome. I will be able to drop my https://github.com/roidelapluie/sslproxy :)

config/http_config.go
brian-brazil: Can we add a TODO at least to clean this up?

simonpasquier: I'm still not clear about what the TODO should be about... AFAICT consumers of the library using Go 1.12 will be able to do this and it will just work:

```go
client, err := config.NewClient(...)
...
client.CloseIdleConnections()
```

But again I don't see how we could avoid having `CloseIdleConnections()` for all the custom round-trippers if we want `client.CloseIdleConnections()` to be effective.

config/http_config.go
```diff
@@ -308,6 +342,125 @@ func (c *TLSConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	t.mtx.Lock()
+	t.rt = rt
+	t.hashCAFile = h[:]
+	t.mtx.Unlock()
+
+	return rt.RoundTrip(req)
+}
+
+func (t *tlsRoundTripper) CloseIdleConnections() {
+	t.mtx.Lock()
```
brian-brazil: Wouldn't this be a read lock?

simonpasquier: Correct.

PR #175, opened 2019-03-08
promlint: initial copy from github.com/prometheus/prometheus/util/pro…

mdlayher:
…mlint

For discussion in https://github.com/prometheus/prometheus/issues/5317.

Updates https://github.com/prometheus/prometheus/issues/5317.

### DIFF
--- a/promlint/promlint.go
+++ b/promlint/promlint.go
@@ -1,0 +1,268 @@
+// Copyright 2017 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package promlint provides a linter for Prometheus metrics.
+package promlint
+
+import (
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+// A Linter is a Prometheus metrics linter.  It identifies issues with metric
+// names, types, and metadata, and reports them to the caller.
+type Linter struct {
+	r io.Reader
+}
+
+// A Problem is an issue detected by a Linter.
+type Problem struct {
+	// The name of the metric indicated by this Problem.
+	Metric string
+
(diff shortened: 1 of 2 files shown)


promlint/promlint.go
```diff
@@ -0,0 +1,268 @@
+// Copyright 2017 The Prometheus Authors
```
mdlayher: Do we ever update the copyright year when moving code around?

mdlayher:
Is there a reason this is still testing with Go 1.7? We're up to 1.12 now, and IMHO supporting anything older than _maybe_ 1.10 is not worth it.

beorn7:
> Is there a reason this is still testing with Go 1.7? We're up to 1.12 now, and IMHO supporting anything older than maybe 1.10 is not worth it.

That's probably stemming from the 1.7 support of prometheus/client_golang. The idea is to support it there for as long as it doesn't cause any overhead (besides having a line in our CI config). I think we should always support three or perhaps four versions back in client_golang, but we can support more as long as we aren't using any features that were only added in recent versions.

mdlayher:
Four versions back would be 1.8, and all I need is 1.8 for `sort.SliceStable`. :)

But if it's really necessary to keep 1.7 support, I could add a temporary shim in here for 1.7 that calls `sort.Slice` instead, but then the sort output is not stable.

https://golang.org/pkg/sort/#Slice

Food for thought: the Go project's policy on version support: https://golang.org/doc/devel/release.html#policy.

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.

Since I don't maintain these repositories, I will respect your choices, but TBQH I'm not sure we're doing our users a favor by supporting older versions of Go that are not supported by the upstream project.

mdlayher:
Oops, just realized `sort.Slice` was also added in 1.8. I can do it the old-school way and use `sort.Stable` from 1.2 if you'd prefer that.

beorn7:
As said, that kind of effort would totally a reason to drop 1.7 support in client_golang.

mdlayher:
Right on, I wasn't sure if my workaround here qualified! I'll submit a couple of PRs for here and there.

mdlayher:
https://github.com/prometheus/client_golang/pull/546 and https://github.com/prometheus/common/pull/176.

brian-brazil:
As noted in https://github.com/prometheus/prometheus/issues/5317, there's no reason to move this code here currently.

mdlayher:
I wrote this as a general purpose library without any dependency on the Prometheus repository itself, and for lack of a better place, stuffed it into util/.  Although prometheus/prometheus is the only org repository user of it, keeping it here doesn't really feel like the right place since it can be and is imported as a library in other projects as well.

brian-brazil:
Being in common doesn't denote that code is a public API, common is for sharing among our repos.

PR #177, opened 2019-03-09
Split proxy username and password as secrets

roidelapluie:

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -72,6 +72,10 @@ type HTTPClientConfig struct {
 	BearerTokenFile string `yaml:"bearer_token_file,omitempty"`
 	// HTTP proxy server to use to connect to the targets.
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
+	// Username to use to connect to the targets using proxy.
+	ProxyUsername Secret `yaml:"proxy_username,omitempty"`
+	// Password to use to connect to the targets using proxy.
+	ProxyPassword Secret `yaml:"proxy_password,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
 }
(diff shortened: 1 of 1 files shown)


config/http_config.go
```diff
@@ -88,6 +92,12 @@ func (c *HTTPClientConfig) Validate() error {
 	if c.BasicAuth != nil && (string(c.BasicAuth.Password) != "" && c.BasicAuth.PasswordFile != "") {
 		return fmt.Errorf("at most one of basic_auth password & password_file must be configured")
 	}
+	if c.ProxyURL.String() == "" && (string(c.ProxyUsername) != "" && string(c.ProxyPassword) != "") {
+		return fmt.Errorf("if proxy_username is set, proxy_url must be set")
+	}
+	if string(c.ProxyPassword) != "" && string(c.ProxyUsername) == "" {
+		return fmt.Errorf("if proxy_password is set, proxy_username must be set")
```
brian-brazil: Is that what the spec says?

config/http_config.go
```diff
@@ -72,6 +72,10 @@ type HTTPClientConfig struct {
 	BearerTokenFile string `yaml:"bearer_token_file,omitempty"`
 	// HTTP proxy server to use to connect to the targets.
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
+	// Username to use to connect to the targets using proxy.
```
brian-brazil: Are there any other auth methods possible?

simonpasquier:
How about crerating a `ProxyURL` struct embedding `URL` (eg `type ProxyURL struct {URL}`) and implement its `MarshalYAML()` method so it replaces the [`URL.User`](https://golang.org/pkg/net/url/#Userinfo) field with an obfuscated value if it isn't empty?

brian-brazil:
That's not how we deal with secrets elsewhere, and is getting a tad magic.

PR #179, opened 2019-03-21
config: set TLSHandshakeTimeout in HTTP transport

BenoitKnecht:
If omitted (i.e. zero), TLS handshakes never timeout, which means TCP
connections (and their associated goroutines) can stay alive forever.

Also set `ExpectContinueTimeout`, so that both have the same value as in
`http.DefaultTransport`.

See prometheus/prometheus#5394.


brian-brazil:
Thanks!

PR #184, opened 2019-04-04
Adding CA Cert file data as config option

NerdisDev:
I´ve added a field in the tls_config, so that you can put your CA Cert file data in the tls_config. 
I think that´s very useful for who with no option to save each ca cert file onto a drive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,6 @@
 
 include Makefile.common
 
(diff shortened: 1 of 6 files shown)


brian-brazil:
Can you not put the file in place the same way you do the config file?

PR #185, opened 2019-04-04
Add the ability to insert the CA certificate data inline

Timwun:
I´ve added a field in the tls_config, so that you can put your CA Cert file data in the tls_config, called ca_data.
I personally have the use case, that I´m having a lot of endpoints with CA certs and one prometheus-server on a GKE Cluster. WIthout ca_data I need to set for each CA cert a new configmap witch is quiet a lot of work.


brian-brazil:
We try to keep with one way to do things. If something should be possible with configuration management, then we avoid duplicating that. It'd also be a bit unusual to have many many CAs, and that they'd not be installed on a system level.

PR #186, opened 2019-04-05
config: extend validity of testdata certs

simonpasquier:
Some test certificates had an end of validity date set to `Jul 13 2019`. I've regenerated them all to be valid for 40 years. FTR I've partially followed the instructions from https://pki-tutorial.readthedocs.io/en/latest/.


brian-brazil:
:+1:

PR #192, opened 2019-05-07
Add json convert function for Duration type

MIBc:
I extended alertmanager api for list and edit its config. The api received a json request and returned a json result.
I found that Duration type didn't support json Unmarshal/Marshal. As a result, it would raise error "json: cannot unmarshal string into Go struct field Route.group_interval of type model.Duration " for edit api. The list api returned 10000 but not "10s" if the config had "group_wait: 10s" or other Duration type config.
I hope this commit could be helpful, if others have the same requirement.

Signed-off-by: lvjiawei <lvjiawei@cmss.chinamobile.com>

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"regexp"
@@ -261,4 +262,19 @@ func (d *Duration) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 	*d = dur
 	return nil
+}
+
+func (d *Duration) UnmarshalJSON(data []byte) error {
+	var s string
+	json.Unmarshal(data, &s)
+	dur, err := ParseDuration(s)
+	if err != nil {
+		return err
+	}
+	*d = dur
+	return nil
+}
+
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
 }


brian-brazil:
Is this actually used in the alertmanager anywhere?

MIBc:
I add these api in alertmanager api.go. The api.reveive and api.respond use json format to receive and response result separately. 
Furthermore, some duration type define like that:
GroupWait      *model.Duration `yaml:"group_wait,omitempty" json:"group_wait,omitempty"`
GroupInterval  *model.Duration `yaml:"group_interval,omitempty" json:"group_interval,omitempty"`
RepeatInterval *model.Duration `yaml:"repeat_interval,omitempty" json:"repeat_interval,omitempty"`
So it make sense to  add json Unmarshal/Marshal for duration type.

brian-brazil:
We're using OpenAPI in the alertmanager now, so that stuff shouldn't be relevant.

In general we only accept changes that make sense for our own codebase, so until alertmanager needs this we should keep things simple here.

MIBc:
OK. But this commit has no relation with alertmanager . I just met this problem in my job. I think other projects also need this, if they use duration type.

brian-brazil:
This is an internal library of the Prometheus project, to keep things sane we only accept changes that make sense for Prometheus.

MIBc:
Thanks a lot . I will close it.

PR #193, opened 2019-05-16
Fix unmarshal of negative times

jacksontj:
While adding tests in https://github.com/prometheus/client_golang/pull/570 I noticed that this `model.Time` doesn't properly unmarshal negative times.


brian-brazil:
Could you add a unittest for this?

jacksontj:
There weren't any tests for time marshal/Unmarshal, so I added a couple.

brian-brazil:
Thanks!

PR #194, opened 2019-06-13
Allow TLS renegotiation

meyergru:
This patch is to address an interop issue against Microsoft IIS. If, for example, one tries to use client certificates against such a server in blackbox-exporter, there will always be a TLS renegotiation error. The problem has been addressed a while ago in Go by adding the 'Renegotiation' parameter.

As an alternative, this could be made configurable, but I am no Go expert.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -291,7 +291,7 @@ func cloneRequest(r *http.Request) *http.Request {
 
 // NewTLSConfig creates a new tls.Config from the given TLSConfig.
 func NewTLSConfig(cfg *TLSConfig) (*tls.Config, error) {
-	tlsConfig := &tls.Config{InsecureSkipVerify: cfg.InsecureSkipVerify}
+	tlsConfig := &tls.Config{InsecureSkipVerify: cfg.InsecureSkipVerify, Renegotiation: tls.RenegotiateFreelyAsClient}
 
 	// If a CA cert is provided then let's read it in so we can validate the
 	// scrape target's certificate properly.


brian-brazil:
There may be security issues from enabling this, so it should be disabled by default and configurable. You can see how the other options are done.

meyergru:
I do not think there are security issues by using this as a default: Around the time Go was invented, there actually was a CVS w/r to "insecure" TLS renegotiation (I think in 2011), which may have been the reason for Go disabling this per default. This has long since been fixed in OpenSSL and the likes and was only "fixed" in Go around 2016.

Matter-of-fact, older versions of Windows could disable TLS renegotiation at the server side, but since Windows server 2012, it cannot be disabled any more (we tried!). Thus, this is a real problem for monitoring IIS. Matter-of-fact, almost all tools that I know of use this default and cannot disable it, including most browsers, curl and Postman.

I was very surprised that Go does not allow this. It seems to have some shortcomings in the TLS area. I found also another problem, because the certificates are not sent if the CA they are signed with is not in the trusted issuer list presented by the server.

Since blackbox_exporter can only be given one client certificate, it should send that no matter what and not act like a browser that has access to multiple certificates to choose from.

As I said, I am no Go expert, so I lack the skill (and time) to make that feature configurable.
For the time being, we have to patch out local blackbox_exporter installation, but I thought it might be a good idea to propose this change anyway.

brian-brazil:
There've been a few vulnerabilities around this over the years, as the vast majority of users don't need it it's safest to keep it disabled by default to reduce attack surface.

meyergru:
Following that logic you might as well disable client certificates altogether, since the feature is non-functional at this time anyway.

PR #195, opened 2019-06-17
Update prometheus/client_golang to v1.0.0

beorn7:
Signed-off-by: beorn7 <beorn@grafana.com>


beorn7:
I also plan to tag a v0.5.0 so that code depending on prometheus/common will pick up prometheus/client_golang v1 in their dependency tree.

PR #196, opened 2019-07-02
Bump github.com/sirupsen/logrus

simonpasquier:
See https://github.com/prometheus/consul_exporter/pull/120#issuecomment-506650462

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,9 +10,9 @@ require (
 	github.com/pkg/errors v0.8.0
 	github.com/prometheus/client_golang v1.0.0
 	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
-	github.com/sirupsen/logrus v1.2.0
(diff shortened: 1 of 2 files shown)


brian-brazil:
Shouldn't we be moving stuff to go kit?

simonpasquier:
I agree. I can create the corresponding issues in the repositories that still use logrus.

@SuperQ WDYT?

SuperQ:
Are we moving all logging from logrus to go kit? I'm not really a fan of it. While it's nice that it does structured logs, it made some stuff more difficult. @juliusv and I ran into some issues.

brian-brazil:
That move was decided some time ago.

juliusv:
Yeah, I didn't know we still had logrus around :)

tariq1890:
I think this PR can be closed. Didn't know of this earlier, my bad.

PR #198, opened 2019-07-30
update dependencies of prometheus/common

tariq1890:


brian-brazil:
Thanks!

axw:
Was the inclusion of `vendor` intentional? This has broken prometheus/client_golang for non-module-aware versions of Go.

beorn7:
I'm pretty sure we should not include the `vendor` directory in libraries.

See https://github.com/prometheus/client_golang/issues/625 for how this breaks users.

brian-brazil:
#199 is out to fix

tariq1890:
Whoops. Completely missed the vendor add. Why don't we include the gitignore for this though?

PR #201, opened 2019-07-30
fix all issues reported by the default golangci linters

tariq1890:
This fixes all the issues reported by the default linter tools used by golangci.

### DIFF
--- a/config/tls_config_test.go
+++ b/config/tls_config_test.go
@@ -58,7 +58,7 @@ func TestValidTLSConfig(t *testing.T) {
 		}
 		// non-nil functions are never equal.
 		got.GetClientCertificate = nil
-		if !reflect.DeepEqual(*got, *cfg.config) {
+		if !reflect.DeepEqual(got, cfg.config) {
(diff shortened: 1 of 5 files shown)


model/silence_test.go
```diff
@@ -76,7 +76,7 @@ func TestMatcherValidate(t *testing.T) {
 			t.Errorf("%d. Expected error %q but got none", i, c.err)
 			continue
 		}
-		if c.err == "" && err != nil {
```
tariq1890: This is a redundant tautology.

config/tls_config_test.go
```diff
@@ -58,7 +58,7 @@ func TestValidTLSConfig(t *testing.T) {
 		}
 		// non-nil functions are never equal.
 		got.GetClientCertificate = nil
-		if !reflect.DeepEqual(*got, *cfg.config) {
+		if !reflect.DeepEqual(got, cfg.config) {
```
tariq1890: We shouldn't compare actual structs which contain mutexes. Compare using the pointers instead.

tariq1890:
@brian-brazil @simonpasquier FYI

brian-brazil:
Thanks!

PR #203, opened 2019-08-23
Reduce sync pool byte growth with bufio Writers

afking:
Avoids storing unbounded buffers in sync pool.

### DIFF
--- a/expfmt/text_create.go
+++ b/expfmt/text_create.go
@@ -14,9 +14,10 @@
 package expfmt
 
 import (
-	"bytes"
+	"bufio"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
 	"strconv"
 	"strings"
@@ -27,7 +28,7 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
-// enhancedWriter has all the enhanced write functions needed here. bytes.Buffer
+// enhancedWriter has all the enhanced write functions needed here. bufio.Writer
 // implements it.
 type enhancedWriter interface {
 	io.Writer
@@ -44,7 +45,7 @@ const (
 var (
 	bufPool = sync.Pool{
 		New: func() interface{} {
-			return bytes.NewBuffer(make([]byte, 0, initialBufSize))
+			return bufio.NewWriterSize(ioutil.Discard, initialBufSize)
 		},
 	}
 	numBufPool = sync.Pool{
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR, I'm not quite understanding what the problem is here. Can you expand?

afking:
Sure, so we had some heap growth after upgrading prometheus libraries. Profiling showed bytes buffer growing a lot:
 IMG_7096

The service has a large number of metrics. Could be related to https://github.com/golang/go/issues/23199 ?
This fix uses bufio.Writer which won't increase the buffer size.

afking:
Benchmarking shows very similar performance +- 100ns per test.
```
go test -bench=Create -run=NONE -benchmem
goos: darwin
goarch: amd64
BenchmarkCreate-8                 200000              6865 ns/op               0 B/op          0 allocs/op
BenchmarkCreateBuildInfo-8       5000000               368 ns/op               0 B/op          0 allocs/op
```

brian-brazil:
How many requests are you doing per second to this, and roughly how big is the /metrics?

afking:
Need to have a look for the requests but approximately 11k lines of metrics output.

brian-brazil:
I'd expect then you to have 1-2 buffers (depending on concurrency) of size ~1MB for 100 byte lines, and as the output is always the same size there shouldn't be an issue here. I'm wondering if the usage might be misattributed here.

afking:
Yes potentially, I’ll need to look more into it. I didn’t see a need to store the buffers so this “fix” is a small optimisation.

beorn7:
For context (@brian-brazil asked about it on IRC): The buffer is mostly used to provide the various more efficient `Write...` methods needed here (`WriteRune`, `WriteString`, `WriteByte`), avoiding allocating and copying around tiny byte slices to write a single rune or byte or a short string. If the `io.Writer` passed to `MetricFamilyToText` already implements those methods, the code performs an interface upgrade and does _not_ use the additional buffering at all.

@afking: Based on the above, you could check if the buffer is even used at all (by adding a debug `Println` into the code creating the `bytes.Buffer`). In vanilla cases, no interface upgrade is needed. That's mostly happening if some middleware is used for HTTP exposition.

Having said all that, https://github.com/golang/go/issues/23199 seems relevant to this code. This PR essentially limits the size of that additional buffer we only need if the interface upgrade fails. That seems sane to me. Perhaps have it a bit larger than the 512 bytes we have now? The default buffer size in the `bufio` package is 4k. That's probably a sweet spot for us here too. (Remember: We only need this buffer to convert many small `WriteByte` etc. calls to only a few call of the generic `Write` method. Turning 4000 calls into one will give us most of the benefit.) Also, `initialBufSize` should be renamed into `bufSize` as it will never change with the changes in this PR.

expfmt/text_create.go
```diff
@@ -44,7 +45,7 @@ const (
 var (
 	bufPool = sync.Pool{
 		New: func() interface{} {
-			return bytes.NewBuffer(make([]byte, 0, initialBufSize))
+			return bufio.NewWriterSize(ioutil.Discard, initialBufSize)
```
beorn7: Following up to my initial comment: It's probably fine to simply use `bufio.NewWriter` here and get rid of `initialBufSize` altogether. Let's trust the authors of `bufio` to pick a generally sweet spot for the buf size. If we ever feel we need a different size, we can still introduce it.

afking: Changed to the default bufio size.

afking:
Great thanks for the reviews. I've been on holiday this week so wasn't able to debug my original issue further. Will post here when I'm able to address it.

brian-brazil:
Thanks!

PR #207, opened 2019-09-04
Check for various number formats we don't support.

brian-brazil:
Go 1.13 adds support for parsing a number of different number
formats, which could lead to inadventanly expanding what is
considered valid Prometheus text format. Test for this and
add safety checks for future potential changes.

All tests passed on Go 1.12, before code changes.

@beorn7 Presuming this looks good, I'll handle the other uses next.


brian-brazil:
Yes, I wanted to do this first as it's the "official" parser.

PR #210, opened 2019-09-13
Enable Go 1.13

simonpasquier:


go.mod
```diff
@@ -18,3 +18,5 @@ require (
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+go 1.11
```
brian-brazil: We support go 1.9, as that's what client_golang supports.

simonpasquier: IMO it doesn't make sense to have a version number less than 1.11 since Go modules aren't available for previous versions.

PR #212, opened 2019-10-16
Upgrade gopkg.in/yaml.v2 to v2.2.4 for security

dolmen:
Because v2.2.4 includes better protections against some malicious YAML documents.
https://github.com/go-yaml/yaml/releases/tag/v2.2.4

PR #213, opened 2019-10-17
Bump gopkg.in/yaml.v2

simonpasquier:
This brings in https://github.com/go-yaml/yaml/pull/515.


beorn7:
Note that #212 is doing the same, but missing the DCO. Merging this one.

PR #215, opened 2019-12-09
Make `version.NewCollector` more idiomatic

beorn7:
Discussing a documentation request in client_golang (
https://github.com/prometheus/client_golang/issues/693 ), I realized
that the usage here in the `version` package is not as idiomatic as it
could be. Returning a `GaugeVec`, the metric value could easily be
changed.

To improve things, this could be done with a `NewConstMetric`, but I think the
currently best way for this is a `GaugeFunc`. It is a single statement now, 
it reads more nicely, and it
avoids misaligning the label values. Plus, it returns just a
`Collector`, so no misunderstanding what to do with it (like adding
more elements to the `GaugeVec` or changing the value).

I have checked all usages in the Prometheus GH org. They are all compatible with this change.

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -34,8 +34,8 @@ var (
 )
 
 // NewCollector returns a collector which exports metrics about current version information.
-func NewCollector(program string) *prometheus.GaugeVec {
-	buildInfo := prometheus.NewGaugeVec(
+func NewCollector(program string) prometheus.Collector {
+	return prometheus.NewGaugeFunc(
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1: 

This will be easier once client_golang supports OM's Info metrics, which I don't think should pose any issues.

beorn7:
Which has reminded me to file https://github.com/prometheus/client_golang/issues/696 .

PR #217, opened 2020-01-10
route.go: allow to chain instrumentations

roidelapluie:

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -53,6 +53,12 @@ func New() *Router {
 
 // WithInstrumentation returns a router with instrumentation support.
 func (r *Router) WithInstrumentation(instrh func(handlerName string, handler http.HandlerFunc) http.HandlerFunc) *Router {
+	if r.instrh != nil {
+		newInstrh := instrh
+		instrh = func(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
+			return newInstrh(handlerName, r.instrh(handlerName, handler))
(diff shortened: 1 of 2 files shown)


roidelapluie:
The purpose is to clean https://github.com/prometheus/prometheus/blob/64194f7d45cbb7fa2f41c8f219706308f65d511a/web/web.go#L1121-L1129

route/route.go
```diff
@@ -53,6 +53,12 @@ func New() *Router {
 
 // WithInstrumentation returns a router with instrumentation support.
 func (r *Router) WithInstrumentation(instrh func(handlerName string, handler http.HandlerFunc) http.HandlerFunc) *Router {
+	if r.instrh != nil {
```
brian-brazil: Why would it be nil?

roidelapluie: It is the zero value for a function

brian-brazil: That doesn't answer my question. Where are we calling this with nil, and why?

roidelapluie:
We check the existing instrumentation (r.ins) not the new one.

Le sam. 11 janv. 2020 à 10:10, Brian Brazil <notifications@github.com> a
écrit :

route/route.go
roidelapluie: When we call router.New(). This is the existing instrumentation.

brian-brazil:
Thanks!

PR #219, opened 2020-01-18
Very preliminary OpenMetrics creation

beorn7:
This is mostly to enable exemplar rendering in client_golang.

The blacklisting feature is introduced here so that OpenMetrics can be configured as opt-in in client_golang.

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -30,79 +30,133 @@ type Encoder interface {
 	Encode(*dto.MetricFamily) error
 }
 
-type encoder func(*dto.MetricFamily) error
-
-func (e encoder) Encode(v *dto.MetricFamily) error {
-	return e(v)
+// Closer is implemented by Encoders that need to be closed to finalize
+// encoding. (For example, OpenMetrics needs a final `# EOF` line.)
+//
+// Note that all Encoder implementations returned from this package implement
+// Closer, too, even if the Close call is a no-op. This happens in preparation
+// for adding a Close method to the Encoder interface directly in a (mildly
+// breaking) release in the future.
+type Closer interface {
+	Close() error
 }
 
-// Negotiate returns the Content-Type based on the given Accept header.  If no
-// appropriate accepted type is found, FmtText is returned (which is the
-// Prometheus text format). Formats can be optionally blacklisted by providing
-// them as blacklisted parameters.
-func Negotiate(h http.Header, blacklisted ...Format) Format {
-	blacklist := map[Format]struct{}{}
-	for _, format := range blacklisted {
-		blacklist[format] = struct{}{}
-	}
+type encoderCloser struct {
+	encode func(*dto.MetricFamily) error
+	close  func() error
+}
 
+func (ec encoderCloser) Encode(v *dto.MetricFamily) error {
+	return ec.encode(v)
+}
+
+func (ec encoderCloser) Close() error {
+	return ec.close()
+}
+
+// Negotiate returns the Content-Type based on the given Accept header. If no
+// appropriate accepted type is found, FmtText is returned (which is the
+// Prometheus text format). This function will never negotiate FmtOpenMetrics,
+// as the support is still experimental. To include the option to negotiate
+// FmtOpenMetrics, use NegotiateOpenMetrics.
+func Negotiate(h http.Header) Format {
 	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
-		var format Format
 		ver := ac.Params["version"]
 		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
 			switch ac.Params["encoding"] {
 			case "delimited":
-				format = FmtProtoDelim
+				return FmtProtoDelim
 			case "text":
-				format = FmtProtoText
+				return FmtProtoText
 			case "compact-text":
-				format = FmtProtoCompact
-			default:
-				continue
+				return FmtProtoCompact
 			}
-		} else if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
-			format = FmtText
-		} else if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion || ver == "") {
-			format = FmtOpenMetrics
-		} else {
-			continue
 		}
-		if _, listed := blacklist[format]; listed {
-			continue
+		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
+			return FmtText
 		}
-		return format
 	}
 	return FmtText
 }
 
-// NewEncoder returns a new encoder based on content type negotiation.
+// NegotiateIncludingOpenMetrics works like Negotiate but includes
+// FmtOpenMetrics as an option for the result. Note that this function is
+// temporary and will disappear once FmtOpenMetrics is fully supported and as
+// such may be negotiated by the normal Negotiate function.
+func NegotiateIncludingOpenMetrics(h http.Header) Format {
+	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
+		ver := ac.Params["version"]
+		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
+			switch ac.Params["encoding"] {
+			case "delimited":
+				return FmtProtoDelim
+			case "text":
+				return FmtProtoText
+			case "compact-text":
+				return FmtProtoCompact
+			}
+		}
+		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
+			return FmtText
+		}
+		if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion || ver == "") {
+			return FmtOpenMetrics
+		}
+	}
+	return FmtText
+}
+
+// NewEncoder returns a new encoder based on content type negotiation. All
+// Encoder implementations returned by NewEncoder also implement Closer, and
+// callers should always call the Close method. It is currently only required
+// for FmtOpenMetrics, but a future (breaking) release will add the Close method
+// to the Encoder interface directly. The current version of the Encoder
+// interface is kept for backwards compatibility.
 func NewEncoder(w io.Writer, format Format) Encoder {
 	switch format {
 	case FmtProtoDelim:
-		return encoder(func(v *dto.MetricFamily) error {
-			_, err := pbutil.WriteDelimited(w, v)
-			return err
-		})
+		return encoderCloser{
+			encode: func(v *dto.MetricFamily) error {
+				_, err := pbutil.WriteDelimited(w, v)
+				return err
+			},
+			close: func() error { return nil },
+		}
 	case FmtProtoCompact:
-		return encoder(func(v *dto.MetricFamily) error {
-			_, err := fmt.Fprintln(w, v.String())
-			return err
-		})
+		return encoderCloser{
+			encode: func(v *dto.MetricFamily) error {
+				_, err := fmt.Fprintln(w, v.String())
+				return err
+			},
+			close: func() error { return nil },
+		}
 	case FmtProtoText:
-		return encoder(func(v *dto.MetricFamily) error {
-			_, err := fmt.Fprintln(w, proto.MarshalTextString(v))
-			return err
-		})
+		return encoderCloser{
+			encode: func(v *dto.MetricFamily) error {
+				_, err := fmt.Fprintln(w, proto.MarshalTextString(v))
+				return err
+			},
+			close: func() error { return nil },
+		}
 	case FmtText:
-		return encoder(func(v *dto.MetricFamily) error {
-			_, err := MetricFamilyToText(w, v)
-			return err
-		})
+		return encoderCloser{
+			encode: func(v *dto.MetricFamily) error {
+				_, err := MetricFamilyToText(w, v)
+				return err
+			},
+			close: func() error { return nil },
+		}
 	case FmtOpenMetrics:
-		return encoder(func(v *dto.MetricFamily) error {
-			_, err := MetricFamilyToOpenMetrics(w, v)
-			return err
-		})
+		return encoderCloser{
+			encode: func(v *dto.MetricFamily) error {
+				_, err := MetricFamilyToOpenMetrics(w, v)
+				return err
+			},
+			close: func() error {
+				_, err := FinalizeOpenMetrics(w)
+				return err
+			},
+		}
 	}
 	panic(fmt.Errorf("expfmt.NewEncoder: unknown format %q", format))
 }
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -36,26 +36,41 @@ func (e encoder) Encode(v *dto.MetricFamily) error {
 	return e(v)
 }
 
-// Negotiate returns the Content-Type based on the given Accept header.
-// If no appropriate accepted type is found, FmtText is returned.
-func Negotiate(h http.Header) Format {
+// Negotiate returns the Content-Type based on the given Accept header.  If no
+// appropriate accepted type is found, FmtText is returned (which is the
+// Prometheus text format). Formats can be optionally blacklisted by providing
+// them as blacklisted parameters.
+func Negotiate(h http.Header, blacklisted ...Format) Format {
+	blacklist := map[Format]struct{}{}
+	for _, format := range blacklisted {
+		blacklist[format] = struct{}{}
+	}
+
 	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
-		// Check for protocol buffer
+		var format Format
+		ver := ac.Params["version"]
 		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
 			switch ac.Params["encoding"] {
 			case "delimited":
-				return FmtProtoDelim
+				format = FmtProtoDelim
 			case "text":
-				return FmtProtoText
+				format = FmtProtoText
 			case "compact-text":
-				return FmtProtoCompact
+				format = FmtProtoCompact
+			default:
+				continue
 			}
+		} else if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
+			format = FmtText
+		} else if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion || ver == "") {
+			format = FmtOpenMetrics
+		} else {
+			continue
 		}
-		// Check for text format.
-		ver := ac.Params["version"]
-		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
-			return FmtText
+		if _, listed := blacklist[format]; listed {
+			continue
 		}
+		return format
 	}
 	return FmtText
 }
@@ -83,6 +98,11 @@ func NewEncoder(w io.Writer, format Format) Encoder {
 			_, err := MetricFamilyToText(w, v)
 			return err
 		})
+	case FmtOpenMetrics:
+		return encoder(func(v *dto.MetricFamily) error {
+			_, err := MetricFamilyToOpenMetrics(w, v)
+			return err
+		})
 	}
-	panic("expfmt.NewEncoder: unknown format")
+	panic(fmt.Errorf("expfmt.NewEncoder: unknown format %q", format))
 }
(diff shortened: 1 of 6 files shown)


expfmt/encode.go
```diff
@@ -36,26 +36,40 @@ func (e encoder) Encode(v *dto.MetricFamily) error {
 	return e(v)
 }
 
-// Negotiate returns the Content-Type based on the given Accept header.
-// If no appropriate accepted type is found, FmtText is returned.
-func Negotiate(h http.Header) Format {
+// Negotiate returns the Content-Type based on the given Accept header.  If no
+// appropriate accepted type is found, FmtText is returned.  Formats can be
```
brian-brazil: Might want to clarify that FmtText is Prometheus format here, so it's not confused with OM text format. Possibly also rename the constants to include Prom too.

Don't we use a single space after a period usually?

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+// This method fulfills the type 'prometheus.encoder'.
+//
+// Note that OpenMetrics requires a final `# EOF` line. Since this function acts
+// on individual metric families, it is the responsibility of the caller to
+// append this line to 'out' once all metric families have been written.
+//
+// TODO(beorn7): The output is not yet fully OpenMetrics compliant. It should,
+// however, work with Prometheus. Known deficiencies are:
+//
+// - A `_total` suffix for counters is not enforced, but if a counter has it, it
```
brian-brazil: You can use `unknown` as the type rather than producing invalid output in that case.

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+//   is truncated from the `# TYPE` and `# HELP` line.
+//
+// - No support for `# UNIT` lines, info type, enum type, `_created` lines.
```
brian-brazil: None of these are mandatory. s/enum/stateset/, and gaugehistogram is missing too.

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily) (written int, err error) {
+	// Fail-fast checks.
+	if len(in.Metric) == 0 {
+		return 0, fmt.Errorf("MetricFamily has no metrics: %s", in)
```
brian-brazil: This is valid, I don't think this should cause a failure.

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+	switch metricType {
+	case dto.MetricType_COUNTER:
+		n, err = w.WriteString(" counter\n")
+	case dto.MetricType_GAUGE:
+		n, err = w.WriteString(" gauge\n")
+	case dto.MetricType_SUMMARY:
+		n, err = w.WriteString(" summary\n")
+	case dto.MetricType_UNTYPED:
+		n, err = w.WriteString(" untyped\n")
```
brian-brazil: unknown

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+				metric.Histogram.GetSampleSum(), 0, false,
+				nil,
+			)
+			written += n
+			if err != nil {
+				return
+			}
+			n, err = writeOpenMetricsSample(
+				w, name, "_count", metric, "", 0,
+				0, metric.Histogram.GetSampleCount(), true,
```
brian-brazil: We require that the +inf is the same as the count, I believe there's races in client_golang that can make that not the case (and this code is also racy as it calls GetSampleCount at two different times). Take the value from the +Inf bucket to be safe.

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+	if err != nil {
+		return written, err
+	}
+	if metric.TimestampMs != nil {
+		err = w.WriteByte(' ')
+		written++
+		if err != nil {
+			return written, err
+		}
+		n, err = writeOpenMetricsFloat(w, float64(*metric.TimestampMs)/1000)
```
brian-brazil: I'd suggest using decimal with timestamps to avoid any potential precision weirdness. And it'll be a tiny bit faster.

As currently specified, only le and quantile need the specific special float rendering. Otherwise you can technically use any number format anywhere.

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+	for _, metric := range in.Metric {
+		switch metricType {
+		case dto.MetricType_COUNTER:
+			if metric.Counter == nil {
+				return written, fmt.Errorf(
+					"expected counter in metric %s %s", name, metric,
+				)
+			}
+			n, err = writeOpenMetricsSample(
+				w, name, "", metric, "", 0,
```
brian-brazil: I'd expect this to have _total here, like it does for _count etc.

expfmt/openmetrics_create_test.go
```diff
@@ -0,0 +1,598 @@
+// Copyright 2014 The Prometheus Authors
```
brian-brazil: 2020

expfmt/openmetrics_create.go
```diff
@@ -0,0 +1,503 @@
...
+// OpenMetrics text format and writes the resulting lines to 'out'. It returns
+// the number of bytes written and any error encountered. The output will have
+// the same order as the input, no further sorting is performed. Furthermore,
+// this function assumes the input is already sanitized and does not perform any
+// sanity checks. If the input contains duplicate metrics or invalid metric or
+// label names, the conversion will result in invalid text format output.
+//
+// This method fulfills the type 'prometheus.encoder'.
+//
+// Note that OpenMetrics requires a final `# EOF` line. Since this function acts
```
brian-brazil: I can't see where you're adding the EOF. I'd suggest adding a Close method to the Encoder interface to cover it.

expfmt/encode.go
beorn7: I clarified in the comment.

At this point, where the OM support is still work in progress, I'd recommend not to potentially break any users of this package if we can avoid it (even if we technically have the right to do so). I'd rather do one breaking change here when we finalize the support and want to have everything nice nice and tidy. So I'd suggest to keep FmtText for now.

The double space was a hiccup of my Emacs formatting settings.

expfmt/openmetrics_create.go
beorn7: Good idea.

beorn7: Will do.

expfmt/openmetrics_create.go
beorn7: Fixed.

expfmt/openmetrics_create.go
beorn7: I have changed the wording. With invalidly named counters changed to `unknown`, the output should now be strictly OM compliant. I changed the wording in this regard, too.

expfmt/openmetrics_create.go
beorn7: Fixed.

expfmt/openmetrics_create.go
beorn7: The problem is that the +Inf bucket is optional in the protobuf. We cannot rely on it to exist, and the current implementation in client_golang generally doesn't add it.

Luckily, the code in client_golang has been fixed not not have a race long ago (as [seen on YouTube](https://www.youtube.com/watch?v=VmrEG-3bWyM)). And the protobuf handed over to the encoder is immutable by contract. So no race can happen.

expfmt/openmetrics_create.go
beorn7: Do you have a neat standard library function in mind to do that formatting? I looked around for a while and only ended up with fairly involved fiddly code, where I'd think the probability of coding a bug is much higher than running into precision issues. But perhaps I am missing something obvious.

expfmt/openmetrics_create.go
beorn7: The metric name has to contain the `_total` (for now) so that current users of client_golang v1 can just switch back and forth between OM and Prometheus text format. Also see the amended doc comment above.

expfmt/openmetrics_create_test.go
beorn7: Fixed.

expfmt/openmetrics_create.go
beorn7: That's the point of the comment: The caller has to add the EOF. A `Close` method would be much better, but as said above, I recommend to not break users of `expfmt` right now, while we are still experimenting. When we wrap things up, we can do one bundled breaking change where we change the `Encoder` interface.

expfmt/openmetrics_create.go
brian-brazil: `sprintf("%d.%09d", t / 1e3,  t % 1e3)` should do it offhand, though watch for negative timestamps.

The exemplar timestamp is more the issue, as that's nanos.

expfmt/openmetrics_create.go
brian-brazil: How are you planning on doing that, considering that the caller doesn't know what has been negotiated?

expfmt/openmetrics_create.go
brian-brazil: My point is more that this seem fragile as it's not the pattern we have elsewhere, so I'd say either restructure the code above so there's _total here or add a comment.

expfmt/encode.go
```diff
@@ -37,8 +37,9 @@ func (e encoder) Encode(v *dto.MetricFamily) error {
 }
 
 // Negotiate returns the Content-Type based on the given Accept header.  If no
-// appropriate accepted type is found, FmtText is returned.  Formats can be
-// optionally blacklisted by providing them as blacklisted parameters.
+// appropriate accepted type is found, FmtText is returned (which is the old
```
brian-brazil: it's the current Prometheus text format, the older Prometheus text format is something else (was that before histograms?)

expfmt/openmetrics_create_test.go
```diff
@@ -36,7 +36,7 @@ func TestCreateOpenMetrics(t *testing.T) {
 		in  *dto.MetricFamily
 		out string
 	}{
-		// 0: Counter, NaN as value, timestamp given.
+		// 0: Counter, NaN as value, timestamp given, no _total suffix.
```
brian-brazil: A counter can't have a NaN value.

brian-brazil:
It occured to me that we're going to need to change the le/quantile values for the Prometheus format at this point too, so you get the same series no matter what is negotiated under the covers.

expfmt/openmetrics_create.go
beorn7: I'd like to avoid fmt.Sprintf as that creates allocations.

However, the `strconv` functions don't provide a similarly powerful formatting, so that negative timestamps or timestamps close to zero have to be treated very carefully.

That's why I think we are more likely to introduce a bug now than ever running into precision issues in practice.

My suggestion is to solve this properly at a later time with plenty of testing of cornercases but don't try to solve it now.
I'll add a TODO about that.

expfmt/openmetrics_create.go
beorn7: I'll add a comment.

expfmt/openmetrics_create.go
beorn7: The caller does know. The flow is to first call `expfmt.Negotiate` to get the content type, and then call `expfmt.NewEncoder` with the negotiated encoding.

It's dirty, as said, but we'll fix it later:
https://github.com/prometheus/client_golang/blob/beorn7/exemplars/prometheus/promhttp/http.go#L189-L207

expfmt/encode.go
beorn7: Removed the "old".

expfmt/openmetrics_create_test.go
beorn7: Is that part of the OpenMetrics spec or generally true?

beorn7:
> It occured to me that we're going to need to change the le/quantile values for the Prometheus format at this point too, so you get the same series no matter what is negotiated under the covers.

I'd say the opposite is required: We don't want to change the le/quantile labels as long as the Prometheus format is negotiated. It would be a breaking change of client_golang if we did. (We mitigated the impact by changing the `histogram_quantile` function, but it will still break many use cases.)

Only changing the le/quantile labels when negotiating OpenMetrics is at least a change that will only happen if a config is changed (in client_golang, OpenMetrics negatiation has to be activated).

It's still not ideal but better than changing the format by a minor version upgrade in client_golang.

Also, I'm still hopeful that https://github.com/OpenObservability/OpenMetrics/issues/129 makes it into the spec. Then we don't need to worry in OpenMetrics.

(And if https://github.com/OpenObservability/OpenMetrics/issues/129 didn't make it into the spec, we wouldn't know an algorithm to reproducibly create correct OM, so we don't have to worry, either.)

expfmt/openmetrics_create.go
brian-brazil: Wouldn't that break anyone using this code directly, as OM will be negotiated, and then rejected by Prometheus? godoc isn't liking our org right now, so can't get an idea of impact.

I don't mind a temporary hack on our end, but let's not break people.

expfmt/openmetrics_create_test.go
brian-brazil: OpenMetrics spec, also generally true (but not enforced).

There's also limits on the size of the exemplar labels, which should probably also be implemented somewhere client side at some point so the whole scrape doesn't get dropped.

expfmt/openmetrics_create.go
beorn7: Yes, that's indeed an issue. With the way the Go Modules dependency resolution works, this could even sneak in without upgrading prometheus/client_golang.

On the other hand, adding a `Close` method would not help, either. (The user that upgrades prometheus/common in the expectation of no breaking changes won't add a call of the `Close` method.) And it would break even more people (those that depend on an unchanged `Encoder` interface, for some reason).

The only way I can think of right now is to make negotiating OM an opt-in, even here on the level of prometheus/common.

Or we just keep things in a branch for now and only merge once this is finalized.

beorn7: Should we decide to leave this in a branch, I would indeed go for the `Close` method approach right now.

expfmt/openmetrics_create_test.go
beorn7: OK, so I have done the following now:
- Changed the test.
- Added to the doc comment of `MetricFamilyToOpenMetrics` that size of exemplar labels and the value of a counter isn't checked yet.

expfmt/openmetrics_create.go
brian-brazil: > The only way I can think of right now is to make negotiating OM an opt-in, even here on the level of prometheus/common.

The plan has always been to enable negotiation by default, and ultimately make it the default if negotiation fails.

> On the other hand, adding a Close method would not help, either. 

We could change the Encode method so things would break obviously. I can't imagine there's that many direct users of this (I'd expect most would be for parsing, not generation), almost everyone should be using client_golang.

beorn7:
> At most we'll make it a SHOULD rather than MUST to produce the right output

But nobody knows how to create the right output.

> I'd hope we'd be enabling it be default, what are your plans there?

At the moment, I cannot see a way how to do it without breaking users in more or less subtle ways. Given the widespread use of the library, I don't think we should make it the default without a major version bump. Thus, I'd like to keep it an opt-in in v1.

> Such algorithms are well known, and implemented in many languages including Go.

My understanding is, as [explained before](https://github.com/OpenObservability/OpenMetrics/issues/129#issuecomment-503154885), that all algorithms we know are concerned with reproducibly creating the same float64 from a given string rather then creating the same string from a given float64. But for OM we need the latter. Happy to be proven wrong, but that hasn't happened yet.

beorn7:
To summarize: My goal here was to let us experiment with exemplars and OM without breaking existing users of this code (and neither client_golang). As the discussion above lead to the insight that an uninformed upgrade of prometheus/common could trigger OM negotiation without the user being aware that the `# EOF` line has to be added (one way or the other), that seems impossible to accomplish.

To avoid working from a branch, I'll add another commit to this PR that uses a separate `Negotiate` function for OM. We can then decide if we want to merge that or keep working from a branch for the experiments.

brian-brazil:
Godoc is still unhappy, but poking around github there seems to be at least a handful of actively maintained projects using this code (though most would be better off using promhttp), so breakage isn't going to be free and this is a kinda-public API.

> We can then decide if we want to merge that or keep working from a branch for the experiments.

I'd prefer not to have two APIs for this. To what extent are you planning on doing experiments, just local or getting it into a Prometheus release?

beorn7:
> I'd prefer not to have two APIs for this. To what extent are you planning on doing experiments, just local or getting it into a Prometheus release?

I want this to instrument targets with exemplars. prometheus/prometheus doesn't need to be released with any of these changes, but if we do this without breaking changes, there is also no harm in doing so.

beorn7:
OK, PHAL.

Now we have a `Closer`, but Go stdlib style as an interface upgrade to not break anybody.

OM will only be an option for content negotiation with a dedicated function to negotiate. It is marked in the doc string as going away once the experiment is over.

expfmt/openmetrics_create.go
beorn7: I have tried another approach now.

I really don't want to break existing uses of this code in any way. Rather I'd keep the experiment in a branch. But I think it should work with what I have pushed now.

expfmt/encode.go
```diff
@@ -51,16 +73,44 @@ func Negotiate(h http.Header) Format {
...
-		// Check for text format.
+		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
+			return FmtText
+		}
+	}
+	return FmtText
+}
+
+// NegotiateOpenMetrics works like Negotiate but includes FmtOpenMetrics as an
+// option for the result. Note that this function is experimental and might
```
brian-brazil: is temporary and will disappear

expfmt/encode.go
```diff
@@ -83,6 +133,17 @@ func NewEncoder(w io.Writer, format Format) Encoder {
 			_, err := MetricFamilyToText(w, v)
 			return err
 		})
+	case FmtOpenMetrics:
+		return encoderCloser{
```
brian-brazil: I'd have all of the above return an encoderCloser too, it'll make it easier to switch in future for users. None of the API usage I saw had an explicit type anywhere.

expfmt/encode.go
```diff
@@ -51,16 +73,44 @@ func Negotiate(h http.Header) Format {
...
+		}
+	}
+	return FmtText
+}
+
+// NegotiateOpenMetrics works like Negotiate but includes FmtOpenMetrics as an
+// option for the result. Note that this function is experimental and might
+// disappear once FmtOpenMetrics is fully supported and as such may be
+// negotiated by the normal Negotiate function.
+func NegotiateOpenMetrics(h http.Header) Format {
```
brian-brazil: Can we avoid putting OpenMetrics in the name? NegotiateWithCloser or something maybe.
As-is, you'd expect it to prefer/only-negotiate OM.

expfmt/encode.go
beorn7: Done.

expfmt/encode.go
beorn7: Done.

expfmt/encode.go
beorn7: `NegotiateWithCloser` wouldn't really convey the meaning, either. It would arguably distract from the actual purpose of this function.

`NegotiateIncludingOpenMetrics` perhaps?

I just tried to

expfmt/encode.go
```diff
@@ -51,38 +72,89 @@ func Negotiate(h http.Header) Format {
...
+		if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion || ver == "") {
+			return FmtOpenMetrics
+		}
 	}
 	return FmtText
 }
 
-// NewEncoder returns a new encoder based on content type negotiation.
+// NewEncoder returns a new encoder based on content type negotiation. The
+// caller is responsible to check if the returned Encoder also implements
```
brian-brazil: We always implement Closer, we're only not returning it for backwards compatibility. I'd suggest saying to always close.

beorn7: I changed the doc comment.

expfmt/encode.go
brian-brazil: NegotiateIncludingOpenMetrics sounds good.

beorn7: Done.

brian-brazil:
:+1:

beorn7:
Thanks. After all this careful consideration, I'll squash this into one commit as I don't think we will take those steps apart again.

PR #225, opened 2020-03-03
promlog: Avoid panic for new logger with empty config

BenoitKnecht:
Make sure that

```golang
promlog.New(&promlog.Config{})
```

doesn't panic. Instead, create a logger with reasonable defaults.

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>


brian-brazil:
Thanks!

BenoitKnecht:
This has to be some sort of record :smile:

PR #226, opened 2020-03-23
Support 0 for model.Duration.

cyriltovena:
This allows to pass the string "0" to parse a model.Duration. This is because `time.Duration` also support this zero value and so it is easier to interchange model.Duration with time.Duration.

time.Duration does parses 0 using this:

```
	// Special case: if all that is left is "0", this is zero.
	if s == "0" {
		return 0, nil
	}
```

see https://golang.org/src/time/format.go?s=40541:40587#L1389

Signed-off-by: Cyril Tovena <cyril.tovena@gmail.com>

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -186,7 +186,7 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
 func ParseDuration(durationStr string) (Duration, error) {
-	// shortcut if the value is 0
+	// Special case: if we receive a "0" string we can directly return a zero value without further processing.
 	if durationStr == "0" {
 		return 0, nil
 	}
--- a/model/time.go
+++ b/model/time.go
@@ -186,7 +186,7 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
 func ParseDuration(durationStr string) (Duration, error) {
-	// Special case: if we receive a "0" string we can directly return a zero value without further processing.
+	// Allow 0 without a unit.
 	if durationStr == "0" {
 		return 0, nil
 	}
--- a/model/time.go
+++ b/model/time.go
@@ -186,6 +186,10 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
 func ParseDuration(durationStr string) (Duration, error) {
+	// shortcut if the value is 0
+	if durationStr == "0" {
+		return 0, nil
+	}
 	matches := durationRE.FindStringSubmatch(durationStr)
 	if len(matches) != 3 {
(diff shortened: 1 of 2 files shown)


brian-brazil:
We always require a non-zero value when this is used in Prometheus, so I think requiring a unit is reasonable.

cyriltovena:
Ok !

gouthamve:
> We always require a non-zero value when this is used in Prometheus

This is not true, we can now set the retention duration as `0s`. I'd say this is a very obscure message and can be made better with this PR:

```
➜  prometheus git:(master) ./prometheus --storage.tsdb.retention.time=0 --storage.tsdb.retention.size=1GB 
Error parsing commandline arguments: not a valid duration string: "0"
usage: prometheus [<flags>]
```

brian-brazil:
> This is not true, we can now set the retention duration as 0s.

I'd argue that that's not a sane setup, as you'll still have 2 hours in the head.

gouthamve:
Nope, this is when I want only size based retention, and not time based. The way to disable time based is to set it to 0, if I'm not wrong?

brian-brazil:
That's true.

model/time.go
```diff
@@ -186,6 +186,10 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
 func ParseDuration(durationStr string) (Duration, error) {
+	// shortcut if the value is 0
```
brian-brazil: This implies that this is a fast path, which it isn't.

beorn7:
Note: https://golang.org/pkg/time/#ParseDuration also doesn't need a unit iff the value is "0".

model/time.go
beorn7: Why isn't it a fast path? If I don't have to spin up the regexp machinery, I'd totally call it a fast path.

In different news: Should we accept `"-0"`? https://golang.org/pkg/time/#ParseDuration does, BTW…

In yet different news, comment should start capitalized and end with a period.

beorn7: I guess what you might refer to is that "0" would not pass the code below, and thus it's not only fast, but _also_ handling a special case. (In general, explaining your thoughts a tad more detailed might help…)

cyriltovena:
> Note: https://golang.org/pkg/time/#ParseDuration also doesn't need a unit iff the value is "0".

Yeah that's what I was trying to explain in the PR description.

model/time.go
cyriltovena: I'll improve this comment should we also support "-0" ? You tell me.

beorn7:
You did @cyriltovena . I have to apologize for my limited attention span. (Pretending to be a millennial… :o)

model/time.go
beorn7: Thinking about it: We do not support negative durations in general, so we should not support `"-0"` either.

model/time.go
```diff
@@ -186,7 +186,7 @@ var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
 func ParseDuration(durationStr string) (Duration, error) {
-	// shortcut if the value is 0
+	// Special case: if we receive a "0" string we can directly return a zero value without further processing.
```
brian-brazil: This is a bit extreme, just "Allow 0 without a unit." would do.

brian-brazil:
Can you add the DCO?

beorn7:
Everyone seems happy now. Merging…

Issue #227, opened 2020-04-06
JSON marshal and unmarshal functions for model.Duration


sid-agrawal:
Ran into the following error when I tried Unmarshalling a JSON output of `github.com/prometheus/alertmanager/config.Config.`

```
json: cannot unmarshal string into Go struct field 
alertmanagerConfig.value.route.routes.group_interval of type model.Duration
```

Looks like there are no UnmarshalJSON and MarshalJSON methods for model.Duration. And a simple fix would be to add those methods. I can make PR if folks agree that, that is the right fix.

roidelapluie:
Fixed in https://github.com/prometheus/common/pull/280

PR #229, opened 2020-04-06
Add JSON markshal and unmarshal support for model.Duration

sid-agrawal:
@brian-brazil this PR addresses https://github.com/prometheus/common/issues/227

Signed-off-by: Sid Agrawal <sid@arista.com>

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"regexp"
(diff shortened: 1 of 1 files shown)


sid-agrawal:
After applying this diff I was able to successfully Unmarshal github.com/prometheus/alertmanager/config.Config

roidelapluie:
The prometheus project is not using json anywhere, why do you need this?

roidelapluie:
Thanks for your contribution.

The prometheus project is not using json anywhere, why do you need this?

brian-brazil:
This is an internal library of Prometheus, you shouldn't be depending on it.

sid-agrawal:
> This is an internal library of Prometheus, you shouldn't be depending on it.
In that case I will close this PR.



> Thanks for your contribution.
> 
> The prometheus project is not using json anywhere, why do you need this?
I was looking for a Go Struct to save the config: https://prometheus.io/docs/alerting/configuration/.

This was my app could programmatically manipulate (i.e. add/delete) receivers, routes etc.

PR #230, opened 2020-04-09
http_config: flag set method for URL

gutorc92:
Signed-off-by: Gustavo Coelho <gutorc@hotmail.com>


brian-brazil:
This is an internal library of Prometheus, can you explain which part of Prometheus you intend to use this for and how?

roidelapluie:
Without answer, let's close this.

PR #231, opened 2020-04-09
Clarify that this repo is considered internal to Prometheus

beorn7:
I have seen this statement multiple times in review comments. It might
help to have it here on the front page.

Signed-off-by: beorn7 <beorn@grafana.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 [![Build Status](https://travis-ci.org/prometheus/common.svg)](https://travis-ci.org/prometheus/common)
 
 This repository contains Go libraries that are shared across Prometheus
-components and libraries.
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1: 

We might want some wording exempting expfmt, as that is the official way to parse our formats in Go.

beorn7:
Is it really? I'd see it more as a reference implementation, at most, i.e. I'd not suggest that we give any stability guarantees (for the package usage, in contrast to the exposition format).

I'll merge this now. We can still iterate on special-casing expfmt if needed.

brian-brazil:
We have at least been pointing people to it for some years, unlike the rest of this repo for which we never did so.

PR #233, opened 2020-05-10
Support HTTP2

roidelapluie:
According to my readings, as we provide a custom transport, we have to
configure it to use HTTP2.

https://github.com/golang/go/issues/20645

I have tested the behaviour of this and it is correct.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/mwitkow/go-conntrack"
+	"golang.org/x/net/http2"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
Maybe it's just ForceAttemptHTTP2 we need to set?

roidelapluie:
https://github.com/golang/net/blob/master/http2/transport.go#L136

it is doing way more than that, and errors if it is already configured

brian-brazil:
Are you sure they're doing something different? That'd seem like a bug to me.

roidelapluie:
I will try this out, indeed this more recent than what I have read

roidelapluie:
Breaks go < 1.13

config/http_config.go
```diff
@@ -153,6 +154,12 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAli
 				conntrack.DialWithName(name),
 			),
 		}
+		// TODO: use ForceAttemptHTTP2 when we move to go 1.13+
```
brian-brazil, line 157: Go

And don't forget your full stop.

PR #235, opened 2020-05-28
Allow using custom Encoders

mwarzynski:
Changes are backwards compatible.
**Goal**: allow using custom encoders in Prometheus.

If one would like to implement custom formatting for metrics, it wasn't possible. There was no way, because both evaluation of which format to use and list of encoders were hardcoded in the code.

Solution is to create a (configurable) instance that would accept the additional encoder implementations (see `EncoderImplementation` struct).

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -41,6 +41,31 @@ type Closer interface {
 	Close() error
 }
 
+// EncoderCreator is able to create Encoder based on provided options.
+// EncoderCreator can negotiate the Format if HTTP Header is provided.
+// Then, using the Format caller may use EncoderCreator to create
+// particular Encoder.
+type EncoderCreator struct {
+	encoders []EncoderImplementation
+}
+
+// EncoderImplementation has all of the information required to define Encoder.
+type EncoderImplementation struct {
+	// HeaderAcceptType defines for what 'Accept' header value this encoder should be used.
+	// Example: value 'application/pdf' would correspond to Header "Accept: application/pdf".
+	HeaderAcceptType string
+	// HeaderAcceptVersion defines for what version of 'Accept' header value this encoder should be used.
+	// Example: value '1.0.0' would correspond to Header "Accept: type; version=1.0.0".
+	HeaderAcceptVersion string
+	// EncodeFormat name of the encoder format. It's actually used as Encoder identifier.
+	// Negotiate returns the Format which can be used by NewEncoder to create an Encoder.
+	EncodeFormat Format
+	// EncoderWriterFunc allows to define the encoding process.
+	EncodeWriterFunc func(w io.Writer) func(v *dto.MetricFamily) error
+	// CloseFunc allows to define how to close the encoding process.
+	CloseFunc func(w io.Writer) func() error
+}
+
 type encoderCloser struct {
 	encode func(*dto.MetricFamily) error
 	close  func() error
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR. This is the code for the exposition format of Prometheus, it is not a generic formatter nora generic content-type negotiation library. Nor are there any plans for Prometheus to produce arbitrary formats.

I'd suggest achieving whatever you goal is by building on top of the existing code rather than by trying to change it.

mwarzynski:
> This is the code for the exposition format of Prometheus

Exactly, my goal is to use Prometheus, but with custom format for `/metrics` endpoint. I wouldn't say not having plans to add arbitrary format could be a valid reason to reject the extendability.

My next step was to allow using the custom encoders in `client_golang`. In fact it requires just a few lines of changes:

```
From 9692d12f87c8d615430a68ee49a0d8098631ff73 Mon Sep 17 00:00:00 2001
From: Mateusz Warzynski
Date: Thu, 28 May 2020 18:19:51 +0200
Subject: [PATCH] Implementation in `client_golang`.

---
 prometheus/promhttp/http.go | 10 +++++++---
 1 file changed, 7 insertions(+), 3 deletions(-)

diff --git a/prometheus/promhttp/http.go b/prometheus/promhttp/http.go
index 5e1c454..c268585 100644
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -111,6 +111,8 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 		}
 	}
 
+	encoders := expfmt.NewEncoderCreator(opts.AdditionalEncoders...)
+
 	h := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
 		if inFlightSem != nil {
 			select {
@@ -146,9 +148,9 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 
 		var contentType expfmt.Format
 		if opts.EnableOpenMetrics {
-			contentType = expfmt.NegotiateIncludingOpenMetrics(req.Header)
+			contentType = encoders.NegotiateIncludingOpenMetrics(req.Header)
 		} else {
-			contentType = expfmt.Negotiate(req.Header)
+			contentType = encoders.Negotiate(req.Header)
 		}
 		header := rsp.Header()
 		header.Set(contentTypeHeader, string(contentType))
@@ -165,7 +167,7 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 			w = gz
 		}
 
-		enc := expfmt.NewEncoder(w, contentType)
+		enc := encoders.NewEncoder(w, contentType)
 
 		// handleError handles the error according to opts.ErrorHandling
 		// and returns true if we have to abort after the handling.
@@ -350,6 +352,8 @@ type HandlerOpts struct {
 	// (which changes the identity of the resulting series on the Prometheus
 	// server).
 	EnableOpenMetrics bool
+
+	AdditionalEncoders []expfmt.EncoderImplementation
 }
 
 // gzipAccepted returns whether the client will accept gzip-encoded content.
-- 
2.26.2
```

I don't see an option to achieve custom format for `/metrics` endpoint without copying the whole code. Sharing my work with community is probably a better way.

brian-brazil:
I'm not aware of any plans to support arbitrary formats for being added to client_golang by @beorn7 either.

> I wouldn't say not having plans to add arbitrary format could be a valid reason to reject the extendability.

I don't see an issue with rejecting features that are out of scope, and which we have no plans to use ourselves.

If you wish to propose wider ranging changes like this, it's generally best to start on the prometheus-developers mailing list.

roidelapluie:
Thank you for your PR.

As brian said,  there is no plans on supporting such a flexibility. We aim at keeping things simple. The text exposition format and OpenMetrics should be sufficient in the Prometheus context.

I suggest you hop on the [developer mailing list](https://groups.google.com/forum/#!forum/prometheus-developers) but it is unlikely that you would get a different answer.

mwarzynski:
Okay.

Thank you for being responsive!

PR #237, opened 2020-06-13
Move modtimevfs to common

roidelapluie:
I'd like to reuse that for the exporter toolkit


roidelapluie:
Draft until more work is put in the toolkit

PR #241, opened 2020-06-29
promlog kit log wrapper: adding Debug/Info/Warn/Error high lvl functions

ahuret:
Provide ready to use leveled log functions:

* Debug
* Info
* Warn
* Error

embedding `log.Logger` keep it BC.

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -35,6 +35,30 @@ var (
 	)
 )
 
+type logger struct {
+	log.Logger
+	debug log.Logger
+	info  log.Logger
+	warn  log.Logger
+	err   log.Logger
+}
+
+func (l *logger) Debug(keyvals ...interface{}) error {
+	return l.debug.Log(keyvals...)
+}
+
+func (l *logger) Info(keyvals ...interface{}) error {
+	return l.info.Log(keyvals...)
+}
+
+func (l *logger) Warn(keyvals ...interface{}) error {
+	return l.warn.Log(keyvals...)
+}
+
+func (l *logger) Err(keyvals ...interface{}) error {
+	return l.err.Log(keyvals...)
+}
+
 // AllowedLevel is a settable identifier for the minimum level a log entry
 // must be have.
 type AllowedLevel struct {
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is an internal library of Prometheus, so any changes to this need to make sense in that context. Why do you want to make this change?

ahuret:
There are multiple benefits by using theses functions:

* clarity of code => `l.Info(...)` vs `level.Info(logger).Log(...)`
* do not recreate logger for each log (see https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L343, each log brings a new logger)

I'll contribute to couchbase-exporter (prometheus) which historically use prometheus logger (the one deprecated), and contribution would be easier by reusing your library.

brian-brazil:
That would be a change in all of our Go repos, not just in Prometheus. It'd also mean we are no longer using a plain go-kit logger, which would be problematic for those importing our code.

ahuret:
> That would be a change in all of our Go repos, not just in Prometheus.

Then we can create mrs, changes will be quick by using sed. I can help.

>  It'd also mean we are no longer using a plain go-kit logger, which would be problematic for those importing our code.

I don't understand, `*logger` here still implements `go-kit logger`

ahuret:
`ag --go "(level.[a-zA-Z])" -l | xargs sed -i 's/level.\([a-zA-Z]\+\)(\(.*\)).Log/\2.\1/g'` covers most cases

brian-brazil:
> I don't understand, *logger here still implements go-kit.Logger interface

Yes, but our code would now require a different interface to be passed around when it started using these. I don't think a bespoke logging interface is desired.

ahuret:
True. Imo it's not really "bespoke", it's a standard way of how loggers works. 
But if you don't want this to change (at least now), I'll implement this at exporter level. Thx for ur time!

brian-brazil:
We've had problems in the past with libraries we depend on using their own log interfaces, and would prefer that we don't go down that route ourselves - plus all the disruption it'd cause to change this.

ahuret:
I understanding. Closing it, then :)

PR #242, opened 2020-06-30
Update dependencies

roidelapluie:


beorn7:
Eeeenteresting. Go 1.11 and go 1.12 create a slightly different `go.mod`, and that trips our tests. I guess that's happening because a dependency has different Go requirements for different versions? Not sure how to proceed from here. There are no build tags for `go.mod` files, I presume...

brian-brazil:
I thought we'd excluded some tests from go-specifc things. Was that just for linting?

Odd that it'd fail on 11&12 and not the surrounding versions.

beorn7:
We have a test that makes sure that `go mod tidy` will not create any changes in the `go.mod` file.

However, as described above, it creates a diff for go 1.11 and go 1.12 because the checked in `go.mod` file is created with go 1.14. Go 1.10 and go 1.9 do not suffer because they have no module support at all.

I guess we could run the `go mod tidy` check only for the current go version. Just not sure if we should actually have different `go.mod` files for different go versions. That would be bad…

roidelapluie:
I have added the indirect dependency to the go.mod. Tests succeed now.

simonpasquier:
> I thought we'd excluded some tests from go-specifc things. Was that just for linting?

[Yes](https://github.com/prometheus/common/blob/546f1fd8d7df61d94633b254641f9f8f48248ada/Makefile#L19) and for [gofmt](https://github.com/prometheus/common/blob/546f1fd8d7df61d94633b254641f9f8f48248ada/.circleci/config.yml#L66-L67) too.

roidelapluie:
Can we merge this?

brian-brazil:
:+1:

PR #246, opened 2020-07-31
Add support for more user-friendly duration input

roidelapluie:
With this pull request, it is possible to write:

6d23h

instead of

167h

Which is a lot more easy to read.

** This is a breaking change, unless we remove the re-marshalling for now. **

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -196,6 +196,8 @@ func ParseDuration(durationStr string) (Duration, error) {
 	}
 	var dur time.Duration
 
+	// Parse the match at pos `pos` in the regex and use `mult` to turn that
+	// into ms, then add that value to the total parsed duration.
 	m := func(pos int, mult time.Duration) {
 		if matches[pos] == "" {
 			return
@@ -205,13 +207,13 @@ func ParseDuration(durationStr string) (Duration, error) {
 		dur += d * mult
 	}
 
-	m(2, 1000*60*60*24*365)
-	m(4, 1000*60*60*24*7)
-	m(6, 1000*60*60*24)
-	m(8, 1000*60*60)
-	m(10, 1000*60)
-	m(12, 1000)
-	m(14, 1)
+	m(2, 1000*60*60*24*365) // y
+	m(4, 1000*60*60*24*7)   // w
+	m(6, 1000*60*60*24)     // d
+	m(8, 1000*60*60)        // h
+	m(10, 1000*60)          // m
+	m(12, 1000)             // s
+	m(14, 1)                // ms
 
 	return Duration(dur), nil
 }
--- a/model/time.go
+++ b/model/time.go
@@ -191,8 +191,7 @@ func ParseDuration(durationStr string) (Duration, error) {
 		return 0, nil
 	}
 	matches := durationRE.FindStringSubmatch(durationStr)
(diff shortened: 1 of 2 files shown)


roidelapluie:
user request: https://github.com/prometheus/prometheus/issues/7681

model/time.go
```diff
@@ -191,67 +191,57 @@ func ParseDuration(durationStr string) (Duration, error) {
 		return 0, nil
 	}
 	matches := durationRE.FindStringSubmatch(durationStr)
-	if len(matches) != 3 {
+	fmt.Printf("%v", len(matches))
```
brian-brazil, line 194: Debug

model/time.go
```diff
@@ -181,7 +181,7 @@ func (d *Duration) Type() string {
 	return "duration"
 }
 
-var durationRE = regexp.MustCompile("^([0-9]+)(y|w|d|h|m|s|ms)$")
+var durationRE = regexp.MustCompile("^(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?$")
```
brian-brazil, line 184: This doesn't match `0`

model/time.go
```diff
@@ -191,67 +191,57 @@ func ParseDuration(durationStr string) (Duration, error) {
 		return 0, nil
 	}
 	matches := durationRE.FindStringSubmatch(durationStr)
-	if len(matches) != 3 {
+	fmt.Printf("%v", len(matches))
+	if len(matches) != 15 {
```
brian-brazil, line 195: Checking for nil is the usual way

roidelapluie:
This would break the people consuming e.g. the rules API.

roidelapluie:
Mmh in fact no. We return the duration as an integer. So that's not breaking.

model/time.go
roidelapluie: Neither did the previous implementation. "0" is treated differently. Note that this is an unexported regex.

roidelapluie: PS: there is already a testcase for "0".

roidelapluie:
I have addressed the comments and added more (relevant) tests

roidelapluie:
I have arbitrarily decided when we marshal a duration to only marshal weeks and years if it is the exact count.

It means that 14d will be changed to 2w but 92d will stay 92d and not 13w1d.

model/time.go
```diff
@@ -191,67 +191,62 @@ func ParseDuration(durationStr string) (Duration, error) {
...
+	m := func(pos int, mult time.Duration) {
+		if matches[pos] == "" {
+			return
+		}
+		n, _ := strconv.Atoi(matches[pos])
+		d := time.Duration(n) * time.Millisecond
+		dur += d * mult
 	}
+
+	m(2, 1000*60*60*24*365)
```
brian-brazil, line 208: A quick comment here listing the y etc. would make things clearer, especially for anyone glancing through the code trying to find the syntax.

roidelapluie: done

brian-brazil:
:+1:

juliusv:
I implemented the frontend side for the graph range input for this here: https://github.com/prometheus/prometheus/pull/7833

PR #248, opened 2020-08-03
config: add SetDirectory methods

abursavich:
Part of cleanup for https://github.com/prometheus/prometheus/pull/7701

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Why does this need to change?

This also doesn't look right, as it's mutating the config the user passed in.

abursavich:
It's a step in parsing a config file to allow for relative paths.

brian-brazil:
Parsing should not involve changing the config file, as we need to render it back to the user. The code already does relative paths.

abursavich:
Maybe I should've said it's a step in "loading" a config file.

https://github.com/prometheus/prometheus/blob/f482c7bdd734ba012f6cedd68ddc24b1416efb35/config/config.go#L55-L67
https://github.com/prometheus/prometheus/blob/f482c7bdd734ba012f6cedd68ddc24b1416efb35/config/config.go#L145-L215

I think it would be cleaner to have the code in this package, but it can live elsewhere and be more clunky:

`func SetBasicAuthDirectory(c *config.BasicAuth, dir string)`
`func SetTLSConfigDirectory(c *config.TLSConfig, dir string)`
`func SetHTTPClientConfigDirectory(c *config.HTTPClientConfig, dir string)`

PR #249, opened 2020-08-03
Disable HTTP/2

roidelapluie:
HTTP/2 support is golang has many problematic cornercases where dead
connections would be kept.

https://github.com/golang/go/issues/32388
https://github.com/golang/go/issues/39337
https://github.com/golang/go/issues/39750

I suggest we disable HTTP/2 for now and enable it manually on the
blackbox exporter.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -155,6 +155,12 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAli
 			),
 		}
 		if enableHTTP2 {
+			// HTTP/2 support is golang has many problematic cornercases where
+			// dead connections would be kept and used in connection pools.
+			// https://github.com/golang/go/issues/32388
+			// https://github.com/golang/go/issues/39337
+			// https://github.com/golang/go/issues/39750
+			// TODO: Re-Enable HTTP/2 once upstream issue is fixed.
 			// TODO: use ForceAttemptHTTP2 when we move to Go 1.13+.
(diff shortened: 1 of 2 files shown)


brian-brazil:
How will the blackbox exporter turn it back on via this code? It doesn't do persistent connections, so those issues likely don't affect it.

roidelapluie:
What about this?

brian-brazil:
Let's not start hiding things in the YAML, we currently use function arguments for this - and the one that's there is already only used by blackbox.

roidelapluie:
@brian-brazil I have adapted it.

config/http_config.go
```diff
@@ -154,10 +154,12 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAli
 				conntrack.DialWithName(name),
 			),
 		}
-		// TODO: use ForceAttemptHTTP2 when we move to Go 1.13+.
-		err := http2.ConfigureTransport(rt.(*http.Transport))
-		if err != nil {
-			return nil, err
+		if enableHTTP2 {
```
brian-brazil, line 157: Can we add a comment that this should go away once Go is fixed?

roidelapluie: Actually  if we add a comment, could we tight the disabling to the disableKeepAlives arg?

brian-brazil: In theory yes, but it feels pretty hacky.

roidelapluie:
I have put the golang issues in comment.

Issue #255, opened 2020-09-10
regression: massive dependency tree on import


howardjohn:
Importing `github.com/prometheus/common` now causes import of 7 million lines of code since #242. Prior to that, only 1.5 MLOC were imported.

This includes:
* aws-sdk-go and aws-sdk-go-v2, and google.golang.org/api at 1M each
* the entireity of Envoy, Consul, Etcd, and Nats
* many many more

The root of this is github.com/go-kit/kit ultimately. It imports a *ton* of stuff recently. Note that client_golang has not yet updated, so likely the ecosystem is not yet fully impacted unless they update prometheus/common directly.

We use it only for ~100 lines of logging code, so we can *almost* just replace it entirely. However, github.com/prometheus/common actually depends on an older version(s) of github.com/prometheus/common due to circular imports. The circular import can be removed if github.com/mwitkow/go-conntrack dependency is dropped. go-conntrack offers tracing and monitoring (using prometheus). Our usage is only the tracing. Therefor, we could remove this circular import with a fork of go-conntrack if desired

brian-brazil:
This repo is an internal library, so dependencies we need for Prometheus anyway aren't a problem per-se nor are issues that arise from 3rd parties importing it directly as such usage isn't supported.

However nothing in this repo should be depending on all of that stuff, and for example the entirety of Nats should be not getting compiled into Prometheus. We only depend on the logging bits of go-kit. Can you explain how this is happening?

Also, it is not generally appropriate to fork our direct dependencies over issues like this. It there's an issue let's tackle it more directly - and I believe you'll find the circular import (which isn't actually circular - as Go forbids those) is between client_golang and common, which we are aware of and which is much trickier to resolve.

SuperQ:
I think this assessment is incorrect. The difference between what Go downloads and what is actually vendored is different.

If I check prior to #242 and run `go mod vendor` this is the result:

```
[prometheus/common/vendor] $ cloc .
     608 text files.
     588 unique files.
      64 files ignored.

3 errors:
Line count, exceeded timeout:  ./golang.org/x/net/idna/tables10.0.0.go
Line count, exceeded timeout:  ./golang.org/x/net/idna/tables11.0.0.go
Line count, exceeded timeout:  ./golang.org/x/net/idna/tables9.0.0.go

github.com/AlDanial/cloc v 1.82  T=12.88 s (42.2 files/s, 25743.1 lines/s)
--------------------------------------------------------------------------------
Language                      files          blank        comment           code
[...]

PR #256, opened 2020-09-10
Reduce size of dependencies by 5x

howardjohn:
See additional info in https://github.com/prometheus/common/issues/255

This is essentially a partial revert of #242


howardjohn:
I don't think I can get the tests passing - `go mod tidy` behavior has changed since go 1.11/1.12 but its expecting the result to be identical across all versions

brian-brazil:
This needs discussion in #255 first, a change like this will generally only be considered if it's causing a problem for something in the Prometheus repo as these are internal libraries which we do not support depending on directly.

roidelapluie:
It sounds like an issue in go-kit, not in common.

howardjohn:
Closing due to conversation in https://github.com/prometheus/common/issues/255 and since it doesn't seem like this is moving forward

PR #258, opened 2020-09-22
Extend version.Print with platform info

dstdfx:
For https://github.com/prometheus/prometheus/issues/7962

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -61,6 +61,7 @@ var versionInfoTmpl = `
   build user:       {{.buildUser}}
   build date:       {{.buildDate}}
   go version:       {{.goVersion}}
+  platform:         {{.platform}}
 `
 
 // Print returns version information.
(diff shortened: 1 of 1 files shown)


brian-brazil:
From context, I think @roidelapluie wants this in the logs so we can easily see it when users post issues. Having it only show on an extra command run doesn't really help.

dstdfx:
> From context, I think @roidelapluie wants this in the logs so we can easily see it when users post issues. Having it only show on an extra command run doesn't really help.

In the referenced issue the output looks exactly like from ```prometheus --version``` command. Moreover the output of this command is mentioned in the [issue template](https://github.com/prometheus/prometheus/blob/master/.github/ISSUE_TEMPLATE/bug_report.md).

roidelapluie:
Yes it is in prometheus --version as that is what we ask users to provide

roidelapluie (approved):

roidelapluie:
We often only get the tail of the logs for users.

brian-brazil:
Thanks!

PR #260, opened 2020-10-21
Router: support HTTP basic auth

rickypai:
adds HTTP basic auth support for Router so we support HTTP basic auth in Prometheus (https://github.com/prometheus/prometheus/pull/8088)


roidelapluie:
Thank you but we would use the node_exporter tls library for this, which still need some minor tweaks. That should hopefully happen in the coming weeks.

brian-brazil:
Closing per above.

PR #261, opened 2020-10-25
Enable HTTP/2 again

roidelapluie:
Given that https://github.com/golang/go/issues/39337 and
https://github.com/golang/go/issues/40201 are fixed:

- We now have a function to configure HTTP/2 transport,
  therefore we have a "ping" every 5 minutes on the line
  if not traffic is doing on
- Dead HTTP/2 connections will be removed from the pool


brian-brazil:
https://github.com/golang/go/issues/39750 and https://github.com/golang/go/issues/32388 aren't closed yet, is this really ready to go?

Do we need to wait for a Go release? We should also have the original reporter try all of this out to be sure.

roidelapluie:
I was thinking golang/go#39337  is the same as golang/go#39750 , and golang/go#32388  is solved by setting the timeouts?

brian-brazil:
I'm not sure looking at those, @JensErat can you help us figure out if this is all sufficiently fixed on Go's end?

PR #263, opened 2020-11-06
Add unique label validation

ipstatic:
Originally created as a PR in the
[client_golang](https://github.com/prometheus/client_golang/pull/812)
repo however was suggested to move it here.

Signed-off-by: Jarod Watkins <jarod@42lines.net>

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -299,6 +299,25 @@ func (p *TextParser) startLabelName() stateFn {
 		p.parseError(fmt.Sprintf("expected '=' after label name, found %q", p.currentByte))
 		return nil
 	}
+	// Check for duplicate label names.
+	if len(p.currentMetric.Label) > 1 {
+		labelCount := make(map[string]int)
+		for _, l := range p.currentMetric.Label {
+			lName := l.GetName()
+			_, exist := labelCount[lName]
+			if !exist {
+				labelCount[lName] = 1
+			} else {
+				labelCount[lName] += 1
+			}
+		}
+		for _, count := range labelCount {
+			if count > 1 {
+				p.parseError(fmt.Sprintf("duplicate label names for metric %q", p.currentMF.GetName()))
+				return nil
+			}
+		}
+	}
 	return p.startLabelValue
 }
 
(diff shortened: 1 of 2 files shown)


expfmt/text_parse.go
```diff
@@ -299,6 +299,25 @@ func (p *TextParser) startLabelName() stateFn {
 		p.parseError(fmt.Sprintf("expected '=' after label name, found %q", p.currentByte))
 		return nil
 	}
+	// Check for duplicate label names.
+	if len(p.currentMetric.Label) > 1 {
```
brian-brazil, line 303: You don't need this check.

expfmt/text_parse.go
```diff
@@ -299,6 +299,25 @@ func (p *TextParser) startLabelName() stateFn {
...
+		labelCount := make(map[string]int)
```
brian-brazil, line 304: You can do this more simply, as we don't need a count - only to detect a duplicate.

expfmt/text_parse.go
ipstatic: I was trying to prevent the code below from executing if there is only one label. But I can remove it.

expfmt/text_parse.go
ipstatic: I might be missing something, but how can I detect duplicates without keeping a count of the labels I have seen?

brian-brazil: If there is an entry in the map, then you know you have seen it. You only need a set, `map[string]struct{}`.

expfmt/text_parse.go
brian-brazil: The for loop already handles that correctly.

expfmt/text_parse.go
```diff
@@ -300,22 +300,15 @@ func (p *TextParser) startLabelName() stateFn {
...
-		}
-		for _, count := range labelCount {
-			if count > 1 {
-				p.parseError(fmt.Sprintf("duplicate label names for metric %q", p.currentMF.GetName()))
-				return nil
-			}
+	labels := make(map[string]struct{})
+	for _, l := range p.currentMetric.Label {
+		lName := l.GetName()
+		_, exist := labels[lName]
```
brian-brazil, line 306: exists

and you can combine this into the if

brian-brazil:
Thanks!

PR #266, opened 2020-12-07
AddingPowerSupport_CI/Testing

santosh653:
Adding Power support & updating go versions to >=1.13 as lower versions are not supported.,

Adding power support ppc64le with Continues Integration/testing so that code remains architecture independent.

This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3.

The build is successful on both arch: amd64/ppc64le, please find the Travis Link below.
https://travis-ci.com/github/santosh653/common

Please let me know if you need any further details.

Thanks You !!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
+arch:
+  - amd64
+  - ppc64le
 language: go
 # Supported Go versions are synced with github.com/prometheus/client_golang.
 go:
-  - 1.9.x
-  - 1.10.x
-  - 1.11.x
-  - 1.12.x
   - 1.13.x
   - 1.14.x
   - 1.15.x
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,24 @@ arch:
 language: go
 # Supported Go versions are synced with github.com/prometheus/client_golang.
 go:
+  - 1.9.x
+  - 1.10.x
+  - 1.11.x
+  - 1.12.x
   - 1.13.x
   - 1.14.x
   - 1.15.x
+jobs:
+ exclude:
+  - go: 1.9.x
+    arch: ppc64le
+  - go: 1.10.x
+    arch: ppc64le
+  - go: 1.11.x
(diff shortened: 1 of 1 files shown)


brian-brazil:
This library support 1.9 or higher as client_golang does, so we shouldn't remove tests for those versions.

We're also looking to move off of Travis.

santosh653:
@brian-brazil some of dependent packs are not available for go 1.9/1.10 hence excluding them and running the code for rest all other go versions.,
Hope this change answers/addresses your query.
The build is successful for rest all.

Thanks you !!

brian-brazil:
We're looking to move off Travis and delete this file, so if you're going to propose any changes it should be over in the CircleCI config.

gerrith3:
@brian-brazil Circle CI does not currently have any ppc64le hardware or access to it at the moment. We are looking into adding them to our CI set but it may be a while.

brian-brazil:
I'm afraid there's not much to do then, as we're moving off Travis.

santosh653:
Are there any queries open on this PR?

gerrith3:
@brian-brazil have you considered running travis for ppc64le/s390x/arm/etc. and CircleCI for intel?  The reason I ask is that prometheus is pretty heavily used on ppc64le and we wind up having to do side ports and validation fairly often.   It would be nice if the top of tree were tested on ppc64le and probably s390x as well.

brian-brazil:
To be clear, we are planning on completely getting rid of Travis. If you care about Prometheus then the place to test would be over in the Prometheus repo - any endian stuff is over there.

santosh653:
Do we have any further queries on this PR? & Is this PR getting merged?

brian-brazil:
This PR will not be merged, as I said we're getting rid of Travis and thus there's not much point in making changes to something we're about to delete.

SuperQ:
We don't use Travis CI anymore.

PR #270, opened 2021-02-08
Add MarshalText and UnmarshalText for Duration

metalmatze:
This is mostly needed for https://github.com/alecthomas/kong.
Furthermore it should be useful for others too, as it implements stdlib interfaces.

### DIFF
--- a/model/time_test.go
+++ b/model/time_test.go
@@ -157,6 +157,79 @@ func TestParseDuration(t *testing.T) {
 	}
 }
 
+func TestDuration_UnmarshalText(t *testing.T) {
+	var cases = []struct {
+		in  string
+		out time.Duration
+
+		expectedString string
+	}{
+		{
+			in:             "0",
+			out:            0,
+			expectedString: "0s",
+		}, {
+			in:             "0w",
+			out:            0,
+			expectedString: "0s",
+		}, {
+			in:  "0s",
+			out: 0,
+		}, {
+			in:  "324ms",
+			out: 324 * time.Millisecond,
+		}, {
+			in:  "3s",
+			out: 3 * time.Second,
+		}, {
+			in:  "5m",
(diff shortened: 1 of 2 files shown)


roidelapluie:
model.Duration is an internal format, generally kong should use time.Duration. What would be the usecase?

metalmatze:
time.Duration doesn't support anything above hours. Therefore it would need a custom wrapper to support `1w`, `1y` and more. 
We want to add flags that support these time ranges too.
As there's no direct dependency on anything and this change implements two stdlib interfaces I don't see this hurting.

roidelapluie:
I am not sure that I'd like to advise large projects like kong to rely on prometheus common given all the direct and indirect dependencies we have. We have complains like https://github.com/prometheus/common/issues/255 already, and I do not want to amplify these. I agree that the code change is small, but it is also using exported interfaces only, such as this could as well be done downstream in kong with the same amount of code:

`type promDuration model.Duration`

Would that be an option?

roidelapluie:
Okay, can you add a comment about which interface you are implementing?

model/time_test.go
```diff
@@ -157,6 +157,79 @@ func TestParseDuration(t *testing.T) {
...
+			t.Errorf("Unexpected error on input %q", c.in)
+		}
+		if time.Duration(d) != c.out {
+			t.Errorf("Expected %v but got %v", c.out, d)
+		}
+		expectedString := c.expectedString
+		if c.expectedString == "" {
+			expectedString = c.in
+		}
+		text, _ := d.MarshalText() // MarshalText returns hardcoded nil
```
roidelapluie, line 226: For the future it would have been great to check for error, even if we only return nil.

roidelapluie:
Thanks!!

Issue #271, opened 2021-02-10
Scrape - configure auth header scheme


dazinator:
## Proposal
Allow Authorization header 'scheme' name and value to be configured.

Right now, I can either configure `Basic` authentication or `Bearer` authentication. Depending on which I configure, dictates the scheme name prefix in the auth header on the scrape request. For example prometheus can be configured to send either of these headers:

- Authorization: BASIC hwhhejfjjjejdnwje
- Authorization: BEARER hwhhejfjjjejdnwje

Each option essentialy places a constraint on the 'scheme name' portion of the authorization header, whilst allowing a static value / token to be passed as the credentials portion. This means I can configure a static value to be passed for the credentials only limited to "Basic" or "Bearer" schemes.

I have an application that uses a custom authentication scheme called API key authentication. This is also a static value that must be passed in the Authorization header. The difference is it requires a custom scheme name 'ApiKey`. I'd like prometheus to allow me to specify the scheme name, so that I am not limited to Basic or Bearer schemes and can use a custom scheme name.

Sure I could set up a custom proxy but this introduces another moving part and more complexity in the landscape so I am thinking this should be weighed against alternatives. Being able to configure the scheme name (as opposed to being restricted to Basic and Bearer) would provide for a more general purpose implementation that makes prometheus more flexible and covers more use cases without much additional complexity that I can tell..

For example right now the following two auth headers are configurable but the third is not.

- Authorization: BASIC hwhhejfjjjejdnwje
- Authorization: BEARER hwhhejfjjjejdnwje
- Authorization: API-KEY hwhhejfjjjejdnwje

I beleive that allowing an authorization header to be configured - by specifying its components:

- scheme name
- value

.. would make for a more flexibile option to cover more use cases.

*“Nice to have” is not a good use case. :)*

We have a .net core web api we'd like to scrape that doesn't support basic auth and only supports short lived access tokens using the BEARER authentication scheme. It does support a custom Authorization scheme for api key authentication (which is a long lived value) however prometheus is currently not able to be configured to leverage that.

roidelapluie:
I have opened the following pull request: https://github.com/prometheus/common/pull/272 to gather feedback from other team members.

dazinator:
I see the PR has now been merged - thank you for implementing this @roidelapluie

roidelapluie:
#274 will still be needed before I can push it down to prometheus

dazinator:
@roidelapluie looks like #274 was merged - can this now be considered closed?
[...]

PR #272, opened 2021-02-14
Support authorization_credentials

roidelapluie:
This backward-compatible patch enables authorization header type to be
set.

For consistency, bearer_token is renamed to authorization_credentials
and a new authorization_type is introduced.

The terminology is taken from
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

refs:
https://github.com/prometheus/common/issues/271
https://github.com/prometheus/prometheus/issues/5107

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -52,6 +52,21 @@ func (a *BasicAuth) SetDirectory(dir string) {
 	a.PasswordFile = JoinDir(dir, a.PasswordFile)
 }
 
+// Authorization contains HTTP authorization credentials.
+type Authorization struct {
+	Type            string `yaml:"type,omitempty"`
+	Credentials     Secret `yaml:"credentials,omitempty"`
+	CredentialsFile string `yaml:"credentials_file,omitempty"`
+}
+
+// SetDirectory joins any relative file paths with dir.
+func (a *Authorization) SetDirectory(dir string) {
+	if a == nil {
+		return
+	}
+	a.CredentialsFile = JoinDir(dir, a.CredentialsFile)
+}
+
 // URL is a custom URL type that allows validation at configuration load time.
 type URL struct {
 	*url.URL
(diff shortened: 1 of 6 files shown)


roidelapluie:
Questions:
Do we want this?


Do we want the form:


```
authorization_credentials: superlongbearer
authorization_type: APIKEY
```
or
```
authorization:
   credentials: superlongbearer
   type: APIKEY
```

Note: the latter would mean always 2 lines (in pretty yaml) and therefore you can't easily `sed` your existing configuration file.

```
authorization_credentials: superlongbearer
```
or
```
authorization:
   credentials: superlongbearer
```

dazinator:
My vote would be for the latter format

```
authorization:
   credentials: superlongbearer
```

roidelapluie:
I have implemented the latter. cc @beorn7

beorn7:
I'll review properly ASAP.

In the meantime, you can fix the CI failures. (o:

config/http_config.go
```diff
@@ -116,6 +140,37 @@ func (c *HTTPClientConfig) Validate() error {
...
+			return fmt.Errorf("authorization is not compatible with bearer_token & bearer_token_file")
+		}
+		if string(c.Authorization.Credentials) != "" && c.Authorization.CredentialsFile != "" {
+			return fmt.Errorf("at most one of authorization credentials & credentials_file must be configured")
+		}
+		c.Authorization.Type = strings.TrimSpace(c.Authorization.Type)
+		if len(c.Authorization.Type) == 0 {
+			c.Authorization.Type = "Bearer"
+		}
+		if strings.ToLower(c.Authorization.Type) == "basic" {
```
dazinator, line 154: What's the benefit for preventing "basic" as the auth type here? Just thinking the two config mechanisms are not strictly equivalent - as with one the user must save the username and password in the config in plain text. In the other the user can save the credentials pre encoded as they are transmitted over the wire. It may be preferable to the user to use the second mechanism?

roidelapluie: This is to reduce support requests. That prevents users to put clear text password here and expect it to work. We have a well defined way to do Basic auth, let's not add a second.

roidelapluie: I would also note that sounds users might have the fans sense of security that this method would use an encrypted password while it's not the case.

beorn7 (approved):

PR #275, opened 2021-02-19
Rename Authorization -> Authentication

roidelapluie:


beorn7 (commented):
Hmm, my comments were more referring to documentation. As the HTTP header is called `Authorization`, for better or worth, it might be the best to go with that in this code, which is closely related to that header.

roidelapluie:
As far as I see everywhere in the documentation I used authorization unless when refering to the option name or the header.

PR #277, opened 2021-02-23
Use http.Handler, not http.HandlerFunc.

tomwilkie:
This is not a backwards incompatible change, even if all `http.HandlerFunc`s are `http.Handler`s - go will implictly convert a function of the right signature to a `HandlerFunc`, but you need to explicitly convert it to a `Handler` via a `HandlerFunc`.

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -41,7 +41,7 @@ func WithParam(ctx context.Context, p, v string) context.Context {
 type Router struct {
 	rtr    *httprouter.Router
 	prefix string
-	instrh func(handlerName string, handler http.HandlerFunc) http.HandlerFunc
+	instrh func(handlerName string, handler http.Handler) http.Handler
 }
 
 // New returns a new Router.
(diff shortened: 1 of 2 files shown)


tomwilkie:
(Do not merge until we've figured out if this is something we even want to do)

beorn7:
I believe the most generally useful way is to accept `http.Handler` but return `http.HandlerFunc`.

tomwilkie:
> I believe the most generally useful way is to accept http.Handler but return http.HandlerFunc.

This sounds eminently reasonable - let me apply this and see what breaks :-)

tomwilkie:
(Whilst I think this is "correct", it doesn't seem worth pursuing, so going to close)

beorn7:
Did it break anything?

tomwilkie:
Nah, it worked well.  Its just more cumbersome to wraps everything in `http.HandlerFunc(...)` - most calls pass in a function right now.

PR #278, opened 2021-02-24
http: Add the ability to not follow redirects

roidelapluie:
Currently, when we unmarshal, we get the default value, which is to
follow redirects.

Related to https://github.com/prometheus/prometheus/issues/5108

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -116,7 +116,7 @@ type HTTPClientConfig struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
-	// FollowRedirects specifies wheter the client should follow HTTP 3xx redirects.
+	// FollowRedirects specifies whether the client should follow HTTP 3xx redirects.
 	// The omitempty flag is not set, because it would be hidden from the
 	// marshalled configuration when set to false.
 	FollowRedirects bool `yaml:"follow_redirects"`
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -111,6 +111,10 @@ type HTTPClientConfig struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
+	// FollowRedirects specifies wheter the client should follow the HTTP 3xx redirects.
+	// The omitempty flag is not set, because it would be hidden from the
+	// marshalled configuration when set to false.
+	FollowRedirects bool `yaml:"follow_redirects"`
 }
 
 // SetDirectory joins any relative file paths with dir.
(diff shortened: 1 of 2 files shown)


config/http_config.go
```diff
@@ -172,6 +176,9 @@ func (c *HTTPClientConfig) Validate() error {
 // UnmarshalYAML implements the yaml.Unmarshaler interface
 func (c *HTTPClientConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type plain HTTPClientConfig
+	*c = HTTPClientConfig{
```
beorn7, line 179: This overrides anything that might be pre-populated in `*c`. Not sure if we ever do that, but wouldn't it be safer (and easier) to do `c.FollowRedirects = true`?

beorn7 (changes requested):
Just a few remarks.

Disclaimer: My HTTP foo is not strong enough to really say if this appropriately deals with the issues discussed in #5108.

config/http_config.go
roidelapluie: No because C is nil at this point.

We do this in Prometheus:

https://github.com/prometheus/prometheus/blob/537c0aff49f14f9eed067c6a54a8edac862273b5/config/config.go#L191

roidelapluie: I have switched to the exact same pattern as prometheus (expose default struct)

beorn7: Just FTR: If `c` is `nil`, then L179, as above, panics.

config/http_config.go
```diff
@@ -111,6 +116,10 @@ type HTTPClientConfig struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
+	// FollowRedirects specifies wheter the client should follow HTTP 3xx redirects.
```
beorn7, line 119: Still "wheter" → "whether"

config/http_config_test.go
```diff
@@ -317,7 +357,7 @@ func TestNewClientFromConfig(t *testing.T) {
 		}
 		response, err := client.Get(testServer.URL)
 		if err != nil {
-			t.Errorf("Can't connect to the test server using this config: %+v", validConfig.clientConfig)
+			t.Errorf("Can't connect to the test server using this config: %+v %v", validConfig.clientConfig, err)
```
beorn7, line 360: It's kind of hard to see that there is an error following after the config. How about:

"Can't connect to the test server using config %+v: %v"

config/http_config.go
roidelapluie: mmmmh yes indeed. but still, this is the pattern we use everywhere, let's stick to it?

config/http_config.go
```diff
@@ -172,6 +181,7 @@ func (c *HTTPClientConfig) Validate() error {
 // UnmarshalYAML implements the yaml.Unmarshaler interface
 func (c *HTTPClientConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type plain HTTPClientConfig
+	*c = DefaultHTTPClientConfig
```
beorn7, line 184: As said, this wouldn't work if `c` were `nil`. (Which implies that `c` is not `nil` rather than that we are doing anything wrong here.)

beorn7 (approved):
Just silly comments and a typo left.

roidelapluie:
Thanks for your review, I think this is good to go.

beorn7:
Yes, sure, that's why I had approved it already. (o:

PR #280, opened 2021-03-05
Add methods to model.Duration for JSON marshalling and unmarshalling

56quarters:
Add methods to model.Duration structs to allow them to be serialized
and unserialized to/from JSON. This helps downstream consumers of the
structs that attempt to use them in objects that are expected to be
able to be represented as JSON (such as Cortex).

An example in Cortex where this would help: [`validation.Limits`](https://github.com/cortexproject/cortex/blob/master/pkg/util/validation/limits.go#L73)

Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"regexp"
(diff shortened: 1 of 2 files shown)


56quarters:
@roidelapluie I'm not sure I understand your question.

roidelapluie:
Prometheus has an API page where we expose rules, it can also unmarshal this as yaml.

However, we do not implement yamlUnmarshaller here .. so I wonder why json marshalling would be mandatory and not yaml marshalling

56quarters:
YAML marshalling (as well as text marshalling) is implemented for this type: https://github.com/prometheus/common/blob/47ee79a16821e03c31ca741659d4f7919aed91c1/model/time.go#L270

roidelapluie:
Oh sorry I did not see it 👀 

LGTM except one thing: can you make all the unmarshal functions look the same (look at how we manage the errors)? Just pick one patterm.

56quarters:
> LGTM except one thing: can you make all the unmarshal functions look the same (look at how we manage the errors)? Just pick one patterm.

Sure, will do

56quarters:
@roidelapluie Changes made but the CI seems to have failed for unrelated reasons. Thanks for all your help on this!

roidelapluie:
Thanks!

roidelapluie:
@56quarters do you need a release for this?

56quarters:
@roidelapluie That'd be great if it's not too much trouble, thanks!

roidelapluie:
v0.19.0 is out :)

PR #286, opened 2021-03-17
Add an environment variable to enable HTTP2

roidelapluie:
HTTP2 had many issues in Prometheus. Before enabling it again, because
some of the underlying issues have been fixed, I would like to introduce
an environment variable to enable it.

That would enable us to contact some "power users" to test this in the
conditions where they had issues to see if this can be re-enabled in
Prometheus generally.

refs #249 #261

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -27,6 +27,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


LeviHarrison:
I don't think an environment variable is the best way to do this. If HTTP/2 has these nasty issues, we should avoid using it whenever possible. For example, in #9068 we think that HTTP/2 is required for scraping, but it isn't needed anywhere else so we should only have it enabled in the scraping engine. An environment variable would turn it on or off everywhere, not giving us the control we need.

roidelapluie:
We do not want to give control to the user. This is to be able to contact the people who have met issues with HTTP2 at scale and ask them to test with the newer go versions to see if we can re-enable it prometheus-wide.

LeviHarrison:
I agree that an environment variable could be effective for testing the re-adoption of HTTP/2 as a whole, but as you pointed out there are still seemingly unresolved bugs and I'm not sure how soon the project will be switching it back on. Ideally, it works great everywhere, but until we know that's the case there will still be users who need HTTP/2 in a certain component and will take assurance that although it could have bugs, it won't affect the rest of their instance.

LeviHarrison:
Should we label this expiremental?

roidelapluie:
The plan is to have the env variable, let the people who discovered the issues test this, and re-enable this project-wide.

roidelapluie:
I have rebased

LeviHarrison (approved):
Tested and works great!

LeviHarrison:
We should put this in the changelog of the new Prometheus release so people now to test

roidelapluie:
I prefer to contact the people who had the issue initially , otherwise it becomes a config option. the goal here is really canary testing.

PR #287, opened 2021-04-11
OAuth 2.0 Client

LeviHarrison:
This is a client implementation of the OAuth 2.0 protocol, specifically for the `client_credentials` grant type. It's intended for use as an authentication method for Prometheus remote write.


roidelapluie:
Thanks.

Why didn't you use the upstream oauth  library? It seems better and also have roundtrippers we can integrate in our http client.

LeviHarrison:
I'm so sorry, I though you meant I was to implement this using _only_ the standard library.

>>> Additionally, I'd like to have this with the golang stdlib and in the prometheus/common repository.

roidelapluie:
https://pkg.go.dev/golang.org/x/oauth2 is what I call stdlib, sorry for the confusion :(

LeviHarrison:
I don't think we can use the `oauth2` RoundTrippers because they don't support `next`-chaining, so we'll have to implement our own. Besides that, I think this package should provide a configuration with YAML aliases, and an initialization function that accepts that configuration and returns a RoundTripper. Does that sound good?

roidelapluie:
I am not following, oauth2 allow setting a `base` roundtripper.

LeviHarrison:
Interesting, I didn't see that. I guess, when added to Prometheus, would could just use the RoundTripper chained together here as the base:

```go
	t := httpClient.Transport

	if conf.SigV4Config != nil {
		t, err = newSigV4RoundTripper(conf.SigV4Config, httpClient.Transport)
		if err != nil {
			return nil, err
		}
	}

	if len(conf.Headers) > 0 {
		t = newInjectHeadersRoundTripper(conf.Headers, t)
	}

	httpClient.Transport = &nethttp.Transport{
		RoundTripper: t,
	}
```

So, in that case, I think all this package would need to have is the YAML configuration and the initialization function. Right?

roidelapluie:
I think so yes

LeviHarrison:
Ok switched over to the `oauth2` package :)

oauth2/oauth2.go
```diff
@@ -0,0 +1,68 @@
...
+	ClientID       string            `yaml:"client_id"`
+	ClientSecret   string            `yaml:"client_secret"`
+	Scopes         []string          `yaml:"scopes,omitempty"`
+	TokenURL       string            `yaml:"token_url"`
+	EndpointParams map[string]string `yaml:"endpoint_params,omitempty"`
+}
+
+// NewOAuth2RoundTripper returns a new http.RoundTripper that authenticates the request
+// with a token fetched using the provided configuration.
+func (c *Config) NewOAuth2RoundTripper(next http.RoundTripper) (http.RoundTripper, error) {
```
roidelapluie, line 37: Can we take the context here?

oauth2/oauth2.go
```diff
@@ -0,0 +1,68 @@
...
+	config := &clientcredentials.Config{
+		ClientID:       c.ClientID,
+		ClientSecret:   c.ClientSecret,
+		Scopes:         c.Scopes,
+		TokenURL:       c.TokenURL,
+		EndpointParams: mapToValues(c.EndpointParams),
+	}
+
+	tokenSource := config.TokenSource(context.Background())
```
roidelapluie, line 46: Pass the context

oauth2/oauth2.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Fetch the token now to see if the config is valid
+	// so we don't run into any errors later.
+	_, err := tokenSource.Token()
```
roidelapluie, line 50: I expect this to make an HTTP call, which I would avoid at this point.

roidelapluie:
Can you please integrate this in the http client?

Thanks!

LeviHarrison:
Do you mean returning an HTTP client?

roidelapluie:
I mean using your new oauth package in https://github.com/prometheus/common/blob/main/config/http_config.go

LeviHarrison:
Done!

config/http_config.go
```diff
@@ -270,6 +277,10 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAli
 		if cfg.BasicAuth != nil {
 			rt = NewBasicAuthRoundTripper(cfg.BasicAuth.Username, cfg.BasicAuth.Password, cfg.BasicAuth.PasswordFile, rt)
 		}
+
+		if cfg.OAuth2 != nil {
+			rt = cfg.OAuth2.NewOAuth2RoundTripper(context.Background(), rt)
```
roidelapluie, line 282: we sould take the context of the request

roidelapluie: mmh I need to re think this

config/http_config.go
```diff
@@ -140,8 +144,8 @@ func (c *HTTPClientConfig) Validate() error {
 	if len(c.BearerToken) > 0 && len(c.BearerTokenFile) > 0 {
 		return fmt.Errorf("at most one of bearer_token & bearer_token_file must be configured")
 	}
-	if c.BasicAuth != nil && (len(c.BearerToken) > 0 || len(c.BearerTokenFile) > 0) {
-		return fmt.Errorf("at most one of basic_auth, bearer_token & bearer_token_file must be configured")
+	if (c.BasicAuth != nil || c.OAuth2 != nil) && (len(c.BearerToken) > 0 || len(c.BearerTokenFile) > 0) {
```
roidelapluie, line 147: oauth2 should be at the same level

config/http_config.go
```diff
@@ -106,6 +108,8 @@ type HTTPClientConfig struct {
 	BasicAuth *BasicAuth `yaml:"basic_auth,omitempty"`
 	// The HTTP authorization credentials for the targets.
 	Authorization *Authorization `yaml:"authorization,omitempty"`
+	// The OAuth2 client credentials used to fetch a token for the targets.
+	OAuth2 *oauth2.Config
```
roidelapluie, line 112: Can we have that in a oauth2 field?

oauth2/oauth2.go
```diff
@@ -0,0 +1,60 @@
...
+	"net/url"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// Config is the client configuration.
+type Config struct {
+	ClientID       string            `yaml:"client_id"`
+	ClientSecret   string            `yaml:"client_secret"`
```
roidelapluie, line 28: This should be a secret.

config/http_config.go
LeviHarrison: Yeah I think the only way to do it would be taking the context as a parameter, but looking at where it's called in Prometheus (in at least one instance), the function their doesn't have the request context either. It might be good to do anyway though.

oauth2/oauth2.go
LeviHarrison: How would I go about making that distinction?

config/http_config.go
LeviHarrison: If you mean
```go
if c.BasicAuth != nil || c.OAuth2 != nil && ... 
```
that causes the tests to fail.

oauth2/oauth2.go
roidelapluie: by using our secret type. it might mean that we move the code under config to avoid cross dependency

LeviHarrison: Thanks!

roidelapluie (approved):
Thank you!

It took me a few days but I have tested this with Keycloak and auth2-proxy and it works as expected.

LeviHarrison:
Thanks so much for your time! Does any more work need to be done in Prometheus itself or does this cover it all?

Issue #288, opened 2021-04-14
Implicit proxy configuration is not supported in http_config


c16a:
When a new `http.Client` is being created from `HTTPClientConfig`, only explicit proxy configuration is supported. Implicit proxy configurations such as HTTP_PROXY, HTTPS_PROXY, and NO_PROXY are ignored.

https://github.com/prometheus/common/blob/4240322d2138024beddb57544851d77a08afe19c/config/http_config.go#L220-L240

If there is no proxy URL provided by application, relevant proxies should be picked up from the environment.

roidelapluie:
closing as per https://github.com/prometheus/prometheus/issues/8720#issuecomment-831154231

PR #289, opened 2021-04-14
Added support for implicit proxy configurations

c16a:
When a new http.Client is being created from HTTPClientConfig,
implicit proxy configurations are not supported.

This commit adds support to pick up proxy configurations
from the environment when the consuming application does not
specify any.

Signed-off-by: Chaitanya Munukutla <chaitanya.m61292@gmail.com>

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"golang.org/x/net/http/httpproxy"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


c16a:
Fixes #288

roidelapluie:
Thank you. We prefer to have proxies explicitly set in the configuration. Additionally, this is a huge breaking change as any people might have proxies to go to the internet and not expect prometheus to use them for scraping etc.

c16a:
@roidelapluie then can you suggest how to exclude proxies for Prometheus remote reads and remote writes? Is it better to have application level exclusion policies instead of having it in common?

roidelapluie:
You should only set a proxy_url: in the prometheus configuration where you need it.

c16a:
Our weird case is that even if we don't need it, it's picking it up from the environment variables. In fact, we have not set the proxy configuration in Prometheus YML at all, but it's picking up proxy for EC2 SD (intended) and remote writes (unintended). Any clue why this might happen?

roidelapluie:
I can not reproduce this locally. Which env variables are you using exactly?

roidelapluie:
closing as per https://github.com/prometheus/prometheus/issues/8720#issuecomment-831154231

PR #290, opened 2021-04-15
Allow to customize DialContext in HTTPClientConfig

pracucci:
In Cortex we have a use case where we would need to customize the dialer used by the HTTP client in the upstream Prometheus alertmanager. Unfortunately we can't do it without changing it upstream so, in this PR, I'm proposing to add a `DialContext` option to `HTTPClientConfig`. Is it something Prometheus is willing to accept?

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -17,11 +17,13 @@ package config
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -120,6 +122,9 @@ type HTTPClientConfig struct {
 	// The omitempty flag is not set, because it would be hidden from the
 	// marshalled configuration when set to false.
 	FollowRedirects bool `yaml:"follow_redirects"`
+	// DialContext allows downstream projects to customize the dialer used by the
+	// HTTP client to open the network connection to the server.
+	DialContext func(ctx context.Context, network, addr string) (net.Conn, error) `yaml:"-"`
 }
 
 // SetDirectory joins any relative file paths with dir.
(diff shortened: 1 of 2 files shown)


config/http_config.go
```diff
@@ -120,6 +122,9 @@ type HTTPClientConfig struct {
 	// The omitempty flag is not set, because it would be hidden from the
 	// marshalled configuration when set to false.
 	FollowRedirects bool `yaml:"follow_redirects"`
+	// DialContext allows downstream projects to customize the dialer used by the
+	// HTTP client to open the network connection to the server.
+	DialContext func(ctx context.Context, network, addr string) (net.Conn, error) `yaml:"-"`
```
beorn7, line 127: I'm not sure if we want a config field here that doesn't parse from (or into) Yaml.

Perhaps it's better to add another parameter to `NewRoundTripperFromConfig` below?

I don't have a lot of context (no pun intended) on how this is used throughout the project.

pracucci: Let me give you a bit more context. We're trying to customise the dialer used by HTTP client used by Alertmanager notifiers (eg. webhook). If we move it away from the config we would have to propagate back the changes across notifiers using it so that we can customize their creation in Cortex. Having the `DialContext` in the config would make it transparent and reduce the change surface, but if you prefer that way I can work on it.

beorn7: Yeah, that's why we need the context how the `HTTPClientConfig` is used across the project. My gut feeling is that it is essentially used as a definition for how the config should look like in Yaml config files. Having a field in there that doesn't render as or parse from Yaml would go against that semantics. But my gut feeling might be wrong.

pracucci: Ok. Let's wait for @roidelapluie confirmation and then I can start working on refactoring the common HTTP client and alertmanager receiver integrations to be able to inject it everywhere.

pracucci: An alternative approach:
https://github.com/prometheus/common/pull/291

roidelapluie: I am wondering what you need extra from the dialer, are we missing options in our http client? I also think we should indeed not adding this to this struct.

pracucci: We're trying to do this https://github.com/cortexproject/cortex/pull/4085. Anyway, I proposed a different approach in #291.

beorn7:
The semantics of the config struct seems to be intended as I initially assumed. Since #291 looks promising, I'm closing this for now. Further discussion see #291.

PR #291, opened 2021-04-16
Added functional option to allow to customize DialContext() in HTTP client

pracucci:
In Cortex we have a use case where we would need to customize the dialer used by the HTTP client in the upstream Prometheus alertmanager. Specifically, we're trying to customise the dialer used by HTTP client used by Alertmanager notifiers (eg. webhook). Unfortunately we can't do it without changing it upstream so, in this PR, I'm proposing an alternative option to #290.

This PR is the base for the Alertmanager changes proposed in https://github.com/prometheus/alertmanager/pull/2547.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -17,11 +17,13 @@ package config
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -194,6 +196,24 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return unmarshal((*plain)(a))
 }
 
+// DialContextFunc defines the signature of the DialContext() function implemented
+// by net.Dialer.
+type DialContextFunc func(context.Context, string, string) (net.Conn, error)
+
+type httpClientOptions struct {
+	dialContextFunc DialContextFunc
+}
+
+// HTTPClientOption defines an option that can be applied to the HTTP client.
+type HTTPClientOption func(options *httpClientOptions)
+
+// WithDialContextFunc allows you to override func gets used for the actual dialing. The default is `net.Dialer.DialContext`.
+func WithDialContextFunc(fn DialContextFunc) HTTPClientOption {
+	return func(opts *httpClientOptions) {
+		opts.dialContextFunc = fn
+	}
+}
+
 // NewClient returns a http.Client using the specified http.RoundTripper.
 func newClient(rt http.RoundTripper) *http.Client {
 	return &http.Client{Transport: rt}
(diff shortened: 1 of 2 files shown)


pracucci:
@beorn7 WDYT about this alternative approach?

config/http_config.go
```diff
@@ -194,15 +196,33 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 // NewClient returns a http.Client using the specified http.RoundTripper.
 func newClient(rt http.RoundTripper) *http.Client {
 	return &http.Client{Transport: rt}
 }
 
 // NewClientFromConfig returns a new HTTP client configured for the
 // given config.HTTPClientConfig. The name is used as go-conntrack metric label.
-func NewClientFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives, enableHTTP2 bool) (*http.Client, error) {
-	rt, err := NewRoundTripperFromConfig(cfg, name, disableKeepAlives, enableHTTP2)
+func NewClientFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives, enableHTTP2 bool, optFuncs ...HTTPClientOption) (*http.Client, error) {
```
pracucci, line 224: If this design makes sense to you, I would move `disableKeepAlives` and `enableHTTP2` to functional options too.

config/http_config.go
```diff
@@ -217,7 +237,25 @@ func NewClientFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives, e
...
-func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives, enableHTTP2 bool) (http.RoundTripper, error) {
+func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives, enableHTTP2 bool, optFuncs ...HTTPClientOption) (http.RoundTripper, error) {
+	opts := &httpClientOptions{}
+	for _, f := range optFuncs {
+		f(opts)
+	}
+
+	var dialContext func(ctx context.Context, network, addr string) (net.Conn, error)
+
+	if opts.dialContextFunc != nil {
```
roidelapluie, line 248: Can we build up a list and append to the list when needed, to avoid the repetition?

config/http_config.go
roidelapluie: Yes it makes sense to me.

config/http_config.go
pracucci: I tried that in first place but the data type is not exported. Not sure how I could do it.

roidelapluie: Build the list here

pracucci: Which data type should have the slice?

config/http_config.go
pracucci: @roidelapluie I moved them to functional options. The default I've picked is both keepalive and HTTP2 enabled. To make it more explicit, the function to change these options allow only to disable it. WDYT?

roidelapluie:
See https://github.com/prometheus/prometheus/blob/a9a5f04ff9d67e48c07337518f94f34a8f76b6a3/discovery/moby/dockerswarm.go#L133

pracucci:
> See https://github.com/prometheus/prometheus/blob/a9a5f04ff9d67e48c07337518f94f34a8f76b6a3/discovery/moby/dockerswarm.go#L133

@roidelapluie I understand this, but the `conntrack.dialerOpt` is not exported.

roidelapluie:
It is an interface we can replicate I think, it should work

pracucci:
> It is an interface we can replicate I think, it should work

@roidelapluie I'm sorry to bother you but I can't understand how we can replicate it. The interface `dialerOpt` is based on a data type `dialerOpts` which is not exported. Am I missing anything?

```
type dialerOpts struct {
	name                  string
	monitoring            bool
	tracing               bool
	parentDialContextFunc dialerContextFunc
}

type dialerOpt func(*dialerOpts)
```

roidelapluie:
Oh sorry, you are right. I will give this a final review later on.

pracucci:
> Oh sorry, you are right. I will give this a final review later on.

Thanks. If this change will be accepted, I will then open a PR to conntrack to propose to expose `DialerOpt`. But I would prefer to not block on it, if it's not a big deal for you.

roidelapluie (approved):

PR #293, opened 2021-04-26
Read oauth2 secret from file

roidelapluie:
cc @LeviHarrison

I still need to add tests.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -120,6 +120,14 @@ type OAuth2 struct {
 	EndpointParams   map[string]string `yaml:"endpoint_params,omitempty"`
 }
 
+// SetDirectory joins any relative file paths with dir.
+func (a *OAuth2) SetDirectory(dir string) {
+	if a == nil {
+		return
+	}
+	a.ClientSecretFile = JoinDir(dir, a.ClientSecretFile)
+}
+
 // HTTPClientConfig configures an HTTP client.
 type HTTPClientConfig struct {
 	// The HTTP basic authentication credentials for the targets.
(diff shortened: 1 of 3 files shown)


roidelapluie:
I also need to add a check that a user does not pass the client_secret and the client_secret_file

config/http_config.go
```diff
@@ -462,20 +463,70 @@ func (rt *basicAuthRoundTripper) CloseIdleConnections() {
...
+		changed bool
+	)
+
+	if rt.config.ClientSecretFile != "" {
+		data, err := ioutil.ReadFile(rt.config.ClientSecretFile)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read oauth2 client secret file %s: %s", rt.config.ClientSecretFile, err)
+		}
+		secret = string(data)
+		changed = secret != rt.secret
```
roidelapluie, line 493: todo: add rlock here

roidelapluie:
Todo: strip extra blank chars

LeviHarrison:
Everything so far looks good!

config/http_config.go
```diff
@@ -489,8 +494,10 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		if err != nil {
```
LeviHarrison, line 494: Perhaps should we lock over here instead so we don't get multiple threads reading the file at once and assessing the same thing?

LeviHarrison (commented):
One more thing that GitHub won't let me add a comment to. Do we need to add OAuth2 here now that it can read from file?

https://github.com/roidelapluie/common/blob/9361f0ddb7fc9336498d0569e635add8153cf786/config/http_config.go#L148-L156

config/http_config.go
LeviHarrison: Just above that line, once again GitHub not letting me comment

LeviHarrison: Actually wait never mind that's pointless.

roidelapluie:
Yes thanks!!!

LeviHarrison (approved):
Looks great!

PR #294, opened 2021-04-27
Make OAuth 2.0 client_id and client_secret mandatory

LeviHarrison:
These should be mandatory as per the spec.

cc @roidelapluie

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -210,8 +210,11 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
-		if len(c.OAuth2.ClientID) < 1 || (len(c.OAuth2.ClientSecret) < 1 || len(c.OAuth2.ClientSecretFile) < 1) {
+		if len(c.OAuth2.ClientID) < 1 || (len(c.OAuth2.ClientSecret) < 1 && len(c.OAuth2.ClientSecretFile) < 1) {
 			return fmt.Errorf("the oauth2 client_id and either the client_secret or client_secret_file must be configured")
+		}
+		if len(c.OAuth2.TokenURL) < 1 {
+			return fmt.Errorf("the oauth2 token_url must be configured")
 		}
 		if len(c.OAuth2.ClientSecret) > 0 && len(c.OAuth2.ClientSecretFile) > 0 {
 			return fmt.Errorf("at most one of oauth2 client_secret & client_secret_file must be configured")
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -210,10 +210,13 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
-		if len(c.OAuth2.ClientID) < 1 || (len(c.OAuth2.ClientSecret) < 1 && len(c.OAuth2.ClientSecretFile) < 1) {
-			return fmt.Errorf("the oauth2 client_id and either the client_secret or client_secret_file must be configured")
+		if len(c.OAuth2.ClientID) == 0 {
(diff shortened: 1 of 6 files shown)


config/http_config.go
```diff
@@ -210,6 +210,9 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
+		if len(c.OAuth2.ClientID) < 1 || (len(c.OAuth2.ClientSecret) < 1 || len(c.OAuth2.ClientSecretFile) < 1) {
```
roidelapluie, line 213: We generally use == 0 to check this.

Can we have 2 distincts errors? This would be more readable and more direct for the user.

roidelapluie: We also need the token url mandatory.

roidelapluie (commented):
Thanks,

token URL is also mandatory.

LeviHarrison:
Whoops, that too. Thanks!

config/http_config.go
```diff
@@ -210,6 +210,12 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
+		if len(c.OAuth2.ClientID) < 1 || (len(c.OAuth2.ClientSecret) < 1 && len(c.OAuth2.ClientSecretFile) < 1) {
+			return fmt.Errorf("the oauth2 client_id and either the client_secret or client_secret_file must be configured")
```
roidelapluie, line 214: I think that I would still prefer to use == 0 like elsewhere and have 2 different errors for better readability.

config/http_config.go
```diff
@@ -210,6 +210,12 @@ func (c *HTTPClientConfig) Validate() error {
...
+		}
+		if len(c.OAuth2.TokenURL) < 1 {
```
roidelapluie, line 216:
```suggestion
		if len(c.OAuth2.TokenURL) == 0 {
```

config/http_config.go
```diff
@@ -210,6 +210,15 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
+		if len(c.OAuth2.ClientID) == 0 {
+			return fmt.Errorf("the oauth2 client_id must be configured")
+		}
+		if len(c.OAuth2.ClientSecret) == 0 && len(c.OAuth2.ClientSecretFile) == 0 {
+			return fmt.Errorf("either the oauth2 client_secret or client_secret_file must be configured")
```
roidelapluie, line 217:
```suggestion
			return fmt.Errorf("either oauth2 client_secret or client_secret_file must be configured")
```

config/http_config.go
```diff
@@ -210,6 +210,15 @@ func (c *HTTPClientConfig) Validate() error {
...
+		}
+		if len(c.OAuth2.TokenURL) == 0 {
+			return fmt.Errorf("the oauth2 token_url must be configured")
```
roidelapluie, line 220:
```suggestion
			return fmt.Errorf("oauth2 token_url must be configured")
```

config/http_config.go
```diff
@@ -210,6 +210,15 @@ func (c *HTTPClientConfig) Validate() error {
 		if c.BasicAuth != nil {
 			return fmt.Errorf("at most one of basic_auth, oauth2 & authorization must be configured")
 		}
+		if len(c.OAuth2.ClientID) == 0 {
+			return fmt.Errorf("the oauth2 client_id must be configured")
```
roidelapluie, line 214:
```suggestion
			return fmt.Errorf("oauth2 client_id must be configured")
```

roidelapluie (commented):
Final nits to be consistent with the other error messages. After this is changed I will merge and release 0.23.0

roidelapluie (approved):
Thanks!!!

Issue #299, opened 2021-05-17
expfmt: Handle nil value for Help the same way an an empty string


arnarpall:
I've discovered some minor inconsistent behaviour when I was writing unit tests for my metric collector.
My unit tests consists of calling the `testutil.GatherAndCompare` with a metric with no help value.
```golang
 prometheus.NewDesc("some_metric_name", "", nil, nil)
```

Calling gather will encode the metrics correctly and the HELP is set to the metric name.
```
# HELP some_metric_name
# TYPE some_metric_name gauge
some_metric_name 10
```

However passing this output into the `TextParser.TextToMetricFamilies` method the
parser sets the metric family Help to `nil`  and thus the encoder will never write the help.
```
# TYPE some_metric_name gauge
some_metric_name 10
```

The encoder will never write the help out to the since the encoder has the following guard clause https://github.com/prometheus/common/blob/main/expfmt/text_create.go#L96

This can be fixed very easily by just moving the guard clause could be moved before the help pointer is dereferenced.
https://github.com/prometheus/common/blob/main/expfmt/text_create.go#L112

PR #300, opened 2021-05-17
expfmt: Handle nil value for Help the same way an an empty string

arnarpall:
MetricFamilyToText method updated to alway write the help text even if metric help is nil. 
This ensures that empty strings and nil for help are treated the same way.

The tests for creating text outputs was updated slightly to utilise sub-tests
where the scenario description has been moved into a test name.
This gives a very handy at a glance overview if a test scenario has been broken.

issue: #299

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -210,7 +210,6 @@ func (p *TextParser) startComment() stateFn {
 		p.parseError("invalid metric name in comment")
 		return nil
 	}
-
 	if p.skipBlankTab(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -35,8 +35,8 @@ import (
 type stateFn func() stateFn
 
 const (
-	KeywordHelp = "HELP"
-	KeywordType = "TYPE"
+	keywordHelp = "HELP"
+	keywordType = "TYPE"
 )
 
 // ParseError signals errors while parsing the simple and flat text-based
@@ -179,7 +179,7 @@ func (p *TextParser) startComment() stateFn {
 		return p.startOfLine
 	}
 	keyword := p.currentToken.String()
-	if keyword != KeywordHelp && keyword != KeywordType {
+	if keyword != keywordHelp && keyword != keywordType {
 		// Generic comment, ignore by fast forwarding to end of line.
 		for p.currentByte != '\n' {
 			if p.currentByte, p.err = p.buf.ReadByte(); p.err != nil {
@@ -196,19 +196,16 @@ func (p *TextParser) startComment() stateFn {
 		return nil // Unexpected end of input.
 	}
 	p.setOrCreateCurrentMF()
-
 	// In case we are parsing a HELP comment but no value has been
 	// supplied we make sure that we set the string to empty
-	if keyword == KeywordHelp && p.currentMF.Help == nil {
+	if keyword == keywordHelp && p.currentMF.Help == nil {
 		p.currentMF.Help = proto.String("")
 	}
-
 	if isNewline(p.currentByte) {
 		// At the end of the line already.
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
-
 	if !isBlankOrTab(p.currentByte) {
 		p.parseError("invalid metric name in comment")
 		return nil
@@ -222,11 +219,10 @@ func (p *TextParser) startComment() stateFn {
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
-
 	switch keyword {
-	case KeywordHelp:
+	case keywordHelp:
 		return p.readingHelp
-	case KeywordType:
+	case keywordType:
 		return p.readingType
 	}
 	panic(fmt.Sprintf("code error: unexpected keyword %q", keyword))
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -197,7 +197,7 @@ func (p *TextParser) startComment() stateFn {
 	}
 	p.setOrCreateCurrentMF()
 	// In case we are parsing a HELP comment but no value has been
-	// supplied we make sure that we set the string to empty
+	// supplied we make sure that we set the string to empty.
 	if keyword == keywordHelp && p.currentMF.Help == nil {
 		p.currentMF.Help = proto.String("")
 	}
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -34,6 +34,11 @@ import (
 // by nil.
 type stateFn func() stateFn
 
+const (
+	KeywordHelp = "HELP"
+	KeywordType = "TYPE"
+)
+
 // ParseError signals errors while parsing the simple and flat text-based
 // exchange format.
 type ParseError struct {
@@ -174,7 +179,7 @@ func (p *TextParser) startComment() stateFn {
 		return p.startOfLine
 	}
 	keyword := p.currentToken.String()
-	if keyword != "HELP" && keyword != "TYPE" {
+	if keyword != KeywordHelp && keyword != KeywordType {
 		// Generic comment, ignore by fast forwarding to end of line.
 		for p.currentByte != '\n' {
 			if p.currentByte, p.err = p.buf.ReadByte(); p.err != nil {
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
 	if p.readTokenAsMetricName(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	if p.currentByte == '\n' {
+	p.setOrCreateCurrentMF()
+
+	// In case we are parsing a HELP comment but no value has been
+	// supplied we make sure that we set the string to empty
+	if keyword == KeywordHelp && p.currentMF.Help == nil {
+		p.currentMF.Help = proto.String("")
+	}
+
+	if isNewline(p.currentByte) {
 		// At the end of the line already.
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
+
 	if !isBlankOrTab(p.currentByte) {
 		p.parseError("invalid metric name in comment")
 		return nil
 	}
-	p.setOrCreateCurrentMF()
+
 	if p.skipBlankTab(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	if p.currentByte == '\n' {
+	if isNewline(p.currentByte) {
 		// At the end of the line already.
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
+
 	switch keyword {
-	case "HELP":
+	case KeywordHelp:
 		return p.readingHelp
-	case "TYPE":
+	case KeywordType:
 		return p.readingType
 	}
 	panic(fmt.Sprintf("code error: unexpected keyword %q", keyword))
@@ -492,7 +507,7 @@ func (p *TextParser) startTimestamp() stateFn {
 // readingHelp represents the state where the last byte read (now in
 // p.currentByte) is the first byte of the docstring after 'HELP'.
 func (p *TextParser) readingHelp() stateFn {
-	if p.currentMF.Help != nil {
+	if p.currentMF.GetHelp() != "" {
 		p.parseError(fmt.Sprintf("second HELP line for metric name %q", p.currentMF.GetName()))
 		return nil
 	}
@@ -729,6 +744,10 @@ func isValidMetricNameContinuation(b byte) bool {
 
 func isBlankOrTab(b byte) bool {
 	return b == ' ' || b == '\t'
+}
+
+func isNewline(b byte) bool {
+	return b == '\n'
 }
 
 func isCount(name string) bool {
--- a/expfmt/text_parse_test.go
+++ b/expfmt/text_parse_test.go
@@ -24,8 +24,8 @@ import (
 
 func testTextParse(t testing.TB) {
 	var scenarios = []struct {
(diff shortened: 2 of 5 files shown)


roidelapluie:
Thank you for this.

I will have a look at this soon, but wouldn't this have a lot of side effects like creating unwanted empty help strings?

Also pinging @beorn7 who might know more about the history of this.

arnarpall:
This should not really have any unwanted side effects since it really only affects the scenario when you are actually consuming the metrics in text format for example using `testutil.GatherAndCompare` and expecting the two outputs to be the same.

In 99% of cases I would expect people usually always just set the help string to some value and everything works as expected.
I really only noticed this since I was dynamically creating descriptions for counters and gauges in a custom collector and I just left the help string as and empty value and noticed this wen running gather and compare.

Passing this in as execpted output into gather and compare
```
# HELP spark_appStatus_jobs_failedJobs
# TYPE spark_appStatus_jobs_failedJobs counter
spark_appStatus_jobs_failedJobs{application_id="some.id",application_name="some.name"} 10
```

Fails the test since the help gets set to ni during parsing.
```
    metrics_test.go:255: 
        	Error Trace:	metrics_test.go:255
        	Error:      	Received unexpected error:
        	            	
        	            	metric output does not match expectation; want:
        	            	
        	            	# TYPE test_counter counter
        	            	test_counter 10
        	            	
        	            	got:
        	            	
        	            	# HELP test_counter 
        	            	# TYPE test_counter counter
        	            	test_counter 10
```

beorn7:
The format is tolerant about having or not having the HELP string.

So fundamentally, what we create with or without this PR is both correct.

However, we _are_ changing the behavior here, And that might break some use cases, especially the test use case you are listing here to justify the change, i.e. those user that are passing in no `HELP` line so for would now get a test failure.

Why aren't you just passing in
```
# TYPE spark_appStatus_jobs_failedJobs counter
spark_appStatus_jobs_failedJobs{application_id="some.id",application_name="some.name"} 10
```
in the case you mentioned above?

arnarpall:
Yeah that was what I did originally and thought to myself "yeah this will work"

But actually that is not the behaviour because when we pass that in as the expected output it fails because the text formatter will add the `HELP` in there with the name of the counter.

So if I put in 
```
# TYPE spark_appStatus_jobs_failedJobs counter
spark_appStatus_jobs_failedJobs{application_id="some.id",application_name="some.name"} 10
```
As the thing to pass into `GatherAndCompare`

The result that comes back after going from Text -> Metric -> Text will actually contain the `HELP`
```
# HELP spark_appStatus_jobs_failedJobs
# TYPE spark_appStatus_jobs_failedJobs counter
spark_appStatus_jobs_failedJobs{application_id="some.id",application_name="some.name"} 10
```

beorn7:
Hmm, that's weird. I would expect the sanitation to not work that way. Let me investigate.

beorn7:
So I tried to reproduce your problem, but I couldn't.

If I put in no HELP string, no HELP string is present in what `GatherAndCompare` expects.

Could you share your code where you see this?

arnarpall:
I've created the following test cases to illustrate the parsing of the empty help string.

```golang
func TestEmptyHelpString(t *testing.T) {
	counter := prometheus.NewCounter(prometheus.CounterOpts{
		Name: "tester",
		Help: "",
	})

	registry := prometheus.NewRegistry()
	registry.MustRegister(counter)

	t.Run("WithHelpInExpectedOutput", func(tt *testing.T) {
		expectedOutput := `
			# HELP tester 
			# TYPE tester counter
			tester 0
`
		assert.NoError(tt, testutil.GatherAndCompare(registry, strings.NewReader(expectedOutput)))
	})
	t.Run("WithoutHelpInExpectedOutput", func(tt *testing.T) {
		expectedOutput := `
			# TYPE tester counter
			tester 0
`
		assert.NoError(tt, testutil.GatherAndCompare(registry, strings.NewReader(expectedOutput)))
	})
}

```

These should both fail
```
=== RUN   TestEmptyHelpString/WithHelpInExpectedOutput
        	Error:      	Received unexpected error:
        	            	
        	            	metric output does not match expectation; want:
        	            	
        	            	# TYPE tester counter
        	            	tester 0
        	            	
        	            	got:
        	            	
        	            	# HELP tester 
        	            	# TYPE tester counter
        	            	tester 0
        	Test:       	TestEmptyHelpString/WithHelpInExpectedOutput
=== RUN   TestEmptyHelpString/WithoutHelpInExpectedOutput
       	Error:      	Received unexpected error:
        	            	
        	            	metric output does not match expectation; want:
        	            	
        	            	# TYPE tester counter
        	            	tester 0
        	            	
        	            	got:
        	            	
        	            	# HELP tester 
        	            	# TYPE tester counter
        	            	tester 0
```

beorn7:
Now I get it. I thought the problem you described is the other way around.

So the problem is that an empty string in the reference is swallowed and results in no help string at all.

beorn7:
Thinking about it, wouldn't it be better to fix the text parser (text_parse.go) to create an empty string rather than `nil` in the protobuf if a line `# HELP my_metric_name` is encountered?

arnarpall:
Yeah I thought about that as well. I'll take a quick look.

arnarpall:
@beorn7  of course there is a simpler way to handle this and I'm completely overlooked it. We just need to use the generated Getter methods for the proto object.

So instead of `metricFamilty.Help != nil` we just use `metricFamily.GetHelp()` this will return an empty string for nil values.

arnarpall:
The change is done and since it's `nil` safe by using the `GetHelp()` accessor it makes the code cleaner by removing `if in.Help != nil` branch all together

arnarpall:
True.

I'll revert this change tomorrow morning and modify the parser instead

arnarpall:
@beorn7  I believe what I've done should be sufficient to get this help comment parsing done.

No help at all in the input will leave the `MetricFamily.Help` as `nil`
Help comment exists but empty will result in `MetricFamily.Help` as and empty string

expfmt/text_parse.go
```diff
@@ -34,6 +34,11 @@ import (
 // by nil.
 type stateFn func() stateFn
 
+const (
+	KeywordHelp = "HELP"
+	KeywordType = "TYPE"
```
beorn7, line 39: Could you lower-case those? It's fine to make "magic strings" constants, but I'd prefer if they don't show up in godoc (unless, of course, they are meant for external consumption).

expfmt/text_parse.go
```diff
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
 	if p.readTokenAsMetricName(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	if p.currentByte == '\n' {
+	p.setOrCreateCurrentMF()
+
```
beorn7, line 199: The general coding style in this file is to avoid vertical whitespace. Could you keep it that way?

expfmt/text_parse.go
```diff
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
...
+	// In case we are parsing a HELP comment but no value has been
+	// supplied we make sure that we set the string to empty
+	if keyword == KeywordHelp && p.currentMF.Help == nil {
+		p.currentMF.Help = proto.String("")
+	}
+
```
beorn7, line 205: Same here.

expfmt/text_parse.go
```diff
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
...
+	if isNewline(p.currentByte) {
 		// At the end of the line already.
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
+
```
beorn7, line 211: Here you are even introducing vertical whitespace in unaffected parts. Obviously, it's a matter of taste and different coders have different styles, but the golden rule is to adapt newly added code to the style around it (and not the other way around).

expfmt/text_parse.go
```diff
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
...
 	if !isBlankOrTab(p.currentByte) {
 		p.parseError("invalid metric name in comment")
 		return nil
 	}
-	p.setOrCreateCurrentMF()
+
```
beorn7, line 216: Same.

expfmt/text_parse.go
```diff
@@ -190,28 +195,38 @@ func (p *TextParser) startComment() stateFn {
...
 	if p.skipBlankTab(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	if p.currentByte == '\n' {
+	if isNewline(p.currentByte) {
 		// At the end of the line already.
 		// Again, this is not considered a syntax error.
 		return p.startOfLine
 	}
+
```
beorn7, line 225: And again. (o:

beorn7 (commented):
Looks great. Thank you very much.

Just a nit about the exported constants and about vertical whitespace.

arnarpall:
> Looks great. Thank you very much.
> 
> Just a nit about the exported constants and about vertical whitespace.

I absolutely agree on all points and these are really just oversights of mine. I'll get them fixed right now

arnarpall:
@beorn7  I've addressed your comments.

expfmt/text_parse.go
```diff
@@ -492,7 +502,7 @@ func (p *TextParser) startTimestamp() stateFn {
 // readingHelp represents the state where the last byte read (now in
 // p.currentByte) is the first byte of the docstring after 'HELP'.
 func (p *TextParser) readingHelp() stateFn {
-	if p.currentMF.Help != nil {
+	if p.currentMF.GetHelp() != "" {
```
beorn7, line 505: Sorry for seeing this only now, but wouldn't this allow duplicate HELP strings as long as the first is empty? E.g. the following would be accepted with this PR:

```
# HELP foo
# HELP foo A help string.
# TYPE foo counter
foo 42
```

Maybe the check here now has to happen in line 201? I.e. if the help string is already non-nil there, create the parse error there and return nil. Otherwise, set it to empty and move on.

expfmt/text_parse.go
```diff
@@ -190,7 +195,13 @@ func (p *TextParser) startComment() stateFn {
 	if p.readTokenAsMetricName(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	if p.currentByte == '\n' {
+	p.setOrCreateCurrentMF()
+	// In case we are parsing a HELP comment but no value has been
+	// supplied we make sure that we set the string to empty
```
beorn7, line 200: Another nit: Please end sentence on a period.

beorn7 (changes requested):
Thank you very much.

Now I have found an actual issue, perhaps. See comment.

expfmt/text_parse.go
arnarpall: I've been thinking about this case and I thought that perhaps we should only fail if the help string is none-empty.
So for your example 
```
# HELP foo
# HELP foo A help string.
# TYPE foo counter
foo 42
```

Would result in where the empty help string would be overwritten
```
# HELP foo A help string.
# TYPE foo counter
foo 42
```

I guess it's a matter of being strict about one and only one help entry.
If we decide that the parser should not override empty help values then we would need to move the check up to line 201 as you say.

beorn7: IIRC the parser is tolerant about _where_ in the file the line is (originally, that was meant to be valid exposition format). In turn, I think, that means we should be strict about duplicate lines. Otherwise, it would be confusing if you _want_ the help string to be empty. You add an explicit empty HELP line, but somewhere much further down, a rogue HELP line sets the help string to something else. It would appear like magic to the naive user.

expfmt/text_parse_test.go
```diff
@@ -384,6 +385,73 @@ request_duration_microseconds_count 2693
...
+				},
+			},
+		},
+		// 7: Two Help strings present but empty will overwritten
+		{
+			in: `
+# HELP should_not_be_empty_help_string
+# HELP should_not_be_empty_help_string not empty
+# TYPE should_not_be_empty_help_string counter
+should_not_be_empty_help_string 0
```
beorn7, line 438: As discussed in the other thread, I think this should not be allowed.

(As another side thought: If it were allowed, it would raise the question what happens if the non-empty HELP string comes first and _then_ you get an empty one. I think it would error out with the current code. And then we are in the situation where order suddenly decides between "totally OK" and "absolutely invalid". Again, I would prefer to just cleanly error out in either case.)

My whole point would be invalid if there _were_ a valid use case to have two HELP entries (of which the first is empty). But I doubt there is one…

expfmt/text_parse_test.go
```diff
@@ -384,6 +385,73 @@ request_duration_microseconds_count 2693
 				},
 			},
 		},
+
+		// 5: Help string present but empty will result in empty help for metric family
```
beorn7, line 389: Great to have these test cases. But please end the sentence with a period. (Same below.)

beorn7 (changes requested):

expfmt/text_parse.go
arnarpall: Ahhh I see what you mean. I did not think about this at all. I will implement your proposed changes.

beorn7:
Hmm, where were we with this PR? After almost a year, I have only a distant memory, but it looked like we were very close to resolving this.

@arnarpall any thoughts about why you closed this now? Has the original problem become irrelevant to you?

arnarpall:
Hey hey. 
Yeah I actually never got the time to completely finish this work and the project that I worked on when I discovered this has not really had any changes to it in a while so my workaround just works.

Also since this is something that people have not been reporting, my feeling is that making this change to the parser is probably not warranted since we are dealing with a complete edge case.

PR #302, opened 2021-05-28
[WIP] replaced go-kit/kit with go-kit/log

RinkiyaKeDad:
Signed-off-by: RinkiyaKeDad <arshsharma461@gmail.com>

`github.com/go-kit/kit` is only being used for the log packages and brings in a lot of dependencies as compared to `github.com/go-kit/log`. This would help determine how useful switching would be.

For reference see kubernetes/kubernetes#102144 and [this](https://github.com/prometheus/common/issues/255#issuecomment-844260932) discussion.

cc @dims @gautierdelorme


roidelapluie:
Thank you. We have no plans to switch to github.com/go-kit/log as long as it is alpha release. We are expecting Peter to ping us when it is ready.

dims:
@roidelapluie hence the WIP. We can't move k/k to newer prometheus common until this happens. So we are all in the wait mode. 

We do have some time to switch over ( see 1.22 release schedule https://www.kubernetes.dev/resources/release/ for code freeze ) but would like to get this into all our CI jobs etc way in advance to avoid any nasty surprises.

Thanks for all your help!

/cc @peterbourgon @gouthamve

peterbourgon:
I don't see how I could have been any clearer about this:

Screen Shot 2021-05-28 at 1 15 56 PM

Absolutely do not merge this PR.

dims:
@peterbourgon yes of course, this was open as a work-in-progress to check if there was anything else needed.

peterbourgon:
@dims OK, cool — put it in Draft state, then, probably?

dims:
@peterbourgon that's totally fair. good point.

sagikazarmark:
The library is now marked as stable.

dims:
@roidelapluie @peterbourgon let me know if you want me to file the PR here ( against https://github.com/go-kit/log/tree/v0.1.0 )

PR #303, opened 2021-06-02
Fix marshalling of URL wrapper with nil value.

pstibrany:
`MarshalJSON` must return valid JSON. This PR fixes marshalling of `config.URL` wrapper values with `nil` value. It now returns `null`.

~This PR also changes unmarshalling of `null` or `""` JSON strings into URL: in both cases, wrapper with `nil` value is returned.~ (removed from this PR)

Related to https://github.com/prometheus/alertmanager/pull/2607

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -130,7 +130,7 @@ func (u URL) MarshalJSON() ([]byte, error) {
 	if u.URL != nil {
 		return json.Marshal(u.URL.String())
 	}
-	return nil, nil
+	return []byte("null"), nil
 }
 
 // OAuth2 is the oauth2 client configuration.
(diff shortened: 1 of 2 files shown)


pstibrany:
(This is currently blocking https://github.com/cortexproject/cortex/pull/4237)

pstibrany:
Based on discussion in https://github.com/prometheus/alertmanager/pull/2607, I've removed changes to unmarshalling in this PR, but still added unit tests documenting the current behaviour.

config/http_config.go
```diff
@@ -130,7 +130,7 @@ func (u URL) MarshalJSON() ([]byte, error) {
 	if u.URL != nil {
 		return json.Marshal(u.URL.String())
 	}
-	return nil, nil
+	return []byte("null"), nil
```
stevesg, line 133: [uber-nit] This might seem like a crazy idea, but we could save hard-coding the null by marshaling nil:

```suggestion
	return json.Marshal(nil)
```

stevesg (approved):
LGTM

PR #304, opened 2021-06-03
Replace go-kit/kit with go-kit/log

dims:
`github.com/go-kit/kit` is only being used for the log packages and brings in a lot of dependencies as compared to `github.com/go-kit/log`. This would help determine how useful switching would be.

For reference see [kubernetes/kubernetes#102144](https://github.com/kubernetes/kubernetes/pull/102144) and [this](https://github.com/prometheus/common/issues/255#issuecomment-844260932) discussion.

Fix for https://github.com/prometheus/common/issues/255


lilic:
Thank you!! 💯 I wanted to bump client_golang and common in etcd, but couldn't due to the go-kit needing etcd :D 

cc @roidelapluie

roidelapluie:
Thanks!

PR #305, opened 2021-06-03
Update Go modules

SuperQ:
* Set Go minimum version to 1.13 to support `%w` in errors.
* Bump all modules to the latest versions.

Signed-off-by: SuperQ <superq@gmail.com>


SuperQ:
Looks like there are some deprecation issues with the protobuf libraries in the expfmt package. I held back that version for now.

roidelapluie (approved):

PR #308, opened 2021-06-08
Add WithIdleConnTimeout HTTP client option

austince:
This change enables HTTP long-polling SD mechanisms to configure the idle connection timeout so that they can use this config lib to create HTTP clients.

For example, the `consul_sd` manually creates a `http.Transport`: 
https://github.com/prometheus/prometheus/blob/7bc11dcb06640ce22b4e15eb52b2c065f97cf79a/discovery/consul/consul.go#L191-L198

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -47,6 +47,9 @@ var DefaultHTTPClientConfig = HTTPClientConfig{
 var defaultHTTPClientOptions = httpClientOptions{
 	keepAlivesEnabled: true,
 	http2Enabled:      true,
+	// 5 minutes is typically above the maximum sane scrape interval. So we can
+	// use keepalive for all configurations.
+	idleConnTimeout: 5 * time.Minute,
 }
 
 type closeIdler interface {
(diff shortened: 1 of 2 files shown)


austince:
@roidelapluie could you take a look at this when you get a chance?

roidelapluie:
To be clear, do you plan to update consul_sd after this is merged? :)

I am asking because while we can accept some small contributions, the goal is not that we start maintaining a fully generic http client with all the options possible.

austince:
> To be clear, do you plan to update consul_sd after this is merged? :)

I'd be happy to :) and I'll use it in the updated `kuma_sd` as well https://github.com/prometheus/prometheus/pull/8844

austince:
@roidelapluie do you think this is an acceptable addition, since it has an immediate use case that will not continue to grow?

On a higher note, do you imagine all SDs in Prometheus that need an HTTP client should use this library?

roidelapluie:
Yes, for kuma_sd ofc it's good

austince:
Great, thank you!

PR #309, opened 2021-06-16
Add sigv4 config to common

treid314:
As requested here: https://github.com/prometheus/alertmanager/issues/2559#issuecomment-837374230
Will throw error if a user doesn't supply username and secret key

Signed-off-by: Tyler Reid <tyler.reid@grafana.com>

### DIFF
--- a/config/sigv4_config.go
+++ b/config/sigv4_config.go
@@ -1,0 +1,42 @@
+// Copyright 2021 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


roidelapluie:
Hello, I do not think sigv4 belongs in config.

We should instead have the sigv4 config and the roundtripper in a github.com/prometheus/common/sigv4 package, together with its own go.mod. That way, we can reuse the code easily between Prometheus RW and alertmanager without ading AWS as dependency of common.

Thanks.

treid314:
Oh! I totally misunderstood, I can do that! I'm going to close this PR and make a new one with the sigv4 module.

PR #312, opened 2021-06-18
Parse text openmetrics format to MetricFamilies

ide-rea:
This PR is for https://github.com/prometheus/prometheus/issues/8932. My solution about this issue is divided into two steps.
1. Use the **openMetricsParser** located in  https://github.com/prometheus/prometheus/tree/main/pkg/textparse  to parse the text exchange format into the **MetricFamily map** . this step is implemented in this PR.
2. Use https://github.com/prometheus/client_golang/blob/master/prometheus/testutil/promlint/promlint.go  **promlint** to check the metric foramt. If the first step is correct, i'll fastly plan to implement this.

My concern is that, firstly as the **openMetricsParser** is located in _github.com/prometheus/prometheus/_ , there may be  some hidden circular import problem, although i have tested it works. Secondly, I may be totally wrong about the right direction for solving this issue. i'm very happy to hear from any advices:)!

Signed-off-by: Xiaoyu Zhang <ideoutrea@163.com>


roidelapluie:
The two parsers should not be identical. The one in prometheus is a very optimized parser, made for scrape. It does not collect all the edge cases.

The parser in this repo is for validation and should be less permissive and catch all the issues.

ide-rea:
Firstly, thanks @roidelapluie , but I have few more questions to ask.
> The parser in this repo is for validation and should be less permissive and catch all the issues.

Can i understand the parser shoud be permissive as same as this parser  ?https://github.com/prometheus/common/blob/main/expfmt/text_parse.go

I think the parser may only supported limited openmetrics feature like https://github.com/prometheus/common/blob/main/expfmt/openmetrics_create.go right now , how do you think that?

roidelapluie:
the parser should support:  https://github.com/prometheus/client_python/blob/master/tests/openmetrics/test_parser.py

PR #322, opened 2021-08-12
promlog: check the log level first

royeo:
The logger returned by `level.NewFilter` will check the log level first, this avoids meaningless `caller` and `ts` calls.


roidelapluie:
Can you please explain why we would need this change? Thanks!

royeo:
The logger returned by `log.With` will gets the value of the bound field such as `ts` and `caller`, `runtime.Caller` is expensive. So if it's the previous logic, logger will obtains the binding value before checking the log level, this has some impact on performance.

roidelapluie:
Thank you, this is correct.

PR #324, opened 2021-09-12
Add `WithDialTLSContextFunc` option

LeviHarrison:
This PR adds needed functionality to implement https://github.com/prometheus/prometheus/issues/4827 in the form of the `WithDialTLSContextFunc` option and `__server_name__` label.

Example implementation:

```go
fn := func(ctx context.Context, network, addr string) (net.Conn, error) {
	tls := cfg.ScrapeConfig.HTTPClientConfig.TLSConfig
	tls.ServerName = target.labels.Get(model.ServerNameLabel)

	dialer := &tls.Dialer{
		Config: tls,
	}
	return dialer.DialContext(ctx, network, addr)
}

NewClientFromConfig(cfg, "", WithDialTLSContextFunc(fn))
```

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -283,20 +283,32 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 // by net.Dialer.
 type DialContextFunc func(context.Context, string, string) (net.Conn, error)
 
+// DialTLSContextFunc defines the signature of the DialContext() function implemented
+// by tls.Dialer.
+type DialTLSContextFunc func(context.Context, string, string) (net.Conn, error)
+
 type httpClientOptions struct {
-	dialContextFunc   DialContextFunc
-	keepAlivesEnabled bool
-	http2Enabled      bool
-	idleConnTimeout   time.Duration
+	dialContextFunc    DialContextFunc
+	dialTLSContextFunc DialTLSContextFunc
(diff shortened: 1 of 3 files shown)


LeviHarrison:
Unfortunately, the `DialTLSContext` options was [only introduced in 1.14](https://golang.org/doc/go1.14#net/http), so this patch doesn't support 1.13.

LeviHarrison:
cc @roidelapluie

LeviHarrison:
Do we need to do the same with `DialTLSContextFunc` as we do with `DialContextFunc`?

https://github.com/prometheus/common/blob/d9fc5f48ec67a29d8ad182d5e695df46b713678d/config/http_config.go#L369-L378

roidelapluie:
No.

roidelapluie:
To remove go 1.13, we would need to see with client_golang.

Otherwise, I'd suggest that you go get this commit in your fork so we can assess of the feature would work.

LeviHarrison:
Build tags with version constraints could also be an option.

roidelapluie:
Can you please rebase this so we can continue working on this?

LeviHarrison:
Sure, but if I remember correctly what was holding this up was the usage of this new function: https://github.com/prometheus/prometheus/pull/9482#issuecomment-978047587.

PR #325, opened 2021-09-15
Re-enable HTTP/2

roidelapluie:
HTTP/2 has been tested and validated by some of our users. It is time to
bring it back.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -381,7 +381,7 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
 			DialContext:           dialContext,
 		}
 		if opts.http2Enabled && os.Getenv("PROMETHEUS_COMMON_DISABLE_HTTP2") == "" {
-			// HTTP/2 support is golang has many problematic cornercases where
+			// HTTP/2 support is golang had many problematic cornercases where
 			// dead connections would be kept and used in connection pools.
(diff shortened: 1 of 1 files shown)


beorn7 (approved):

roidelapluie:
Can you please review again? the comment was bad

config/http_config.go
```diff
@@ -380,17 +380,17 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
 			// https://github.com/golang/go/issues/39750
 
-			// Enable HTTP2 if the environment variable
-			// PROMETHEUS_COMMON_ENABLE_HTTP2 is set.
-			// This is a temporary workaround so that users can safely test this
-			// and validate that HTTP2 can be enabled Prometheus-Wide again.
+			// Do not enable HTTP2 if the environment variable
+			// PROMETHEUS_COMMON_DISABLE_HTTP2 is set to a non-empty value.
+			// This is a temporary workaround, than be enabled if we see issues
+			// again in the short future.
```
beorn7, line 393: This still doesn't sound quite right. Perhaps "This allows users to easily disable HTTP2 in case they run into issues again, but will be removed once we are confident that things work as expected."?

beorn7 (changes requested):

beorn7 (approved):

PR #334, opened 2021-10-13
promlog: Fix caller

roidelapluie:
DefaultCaller is hard-coded to select the 3rd stack frame up, but we
first filter by level, which means we need to take a 5th strack frame.

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -118,10 +118,11 @@ func New(config *Config) log.Logger {
 		l = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
 	}
 
-	l = log.With(l, "ts", timestampFormat, "caller", log.DefaultCaller)
-
 	if config.Level != nil {
+		l = log.With(l, "ts", timestampFormat, "caller", log.Caller(5))
 		l = level.NewFilter(l, config.Level.o)
+	} else {
+		l = log.With(l, "ts", timestampFormat, "caller", log.DefaultCaller)
 	}
 	return l
 }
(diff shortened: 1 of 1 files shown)


promlog/log.go
```diff
@@ -167,10 +169,15 @@ func (l *logger) SetLevel(lvl *AllowedLevel) {
 	l.mtx.Lock()
 	defer l.mtx.Unlock()
 	if lvl != nil {
-		if l.currentLevel != nil && l.currentLevel.s != lvl.s {
+		if (l.currentLevel != nil && l.currentLevel.s != lvl.s) || l.currentLevel == nil {
+			l.leveled = log.With(l.base, "ts", timestampFormat, "caller", log.Caller(5))
 			_ = l.base.Log("msg", "Log level changed", "prev", l.currentLevel, "current", lvl)
```
LeviHarrison, line 174: If I understand the logic correctly, this message will now also be logged when a new dynamic logger is created. I don't think that's expected.

promlog/log.go
```diff
@@ -166,11 +168,16 @@ func (l *logger) Log(keyvals ...interface{}) error {
...
+		return
+	}
+
+	l.leveled = log.With(l.base, "ts", timestampFormat, "caller", log.Caller(5))
+	if l.currentLevel != nil && l.currentLevel.s != lvl.s {
+		_ = l.base.Log("msg", "Log level changed", "prev", l.currentLevel, "current", lvl)
 	}
-	l.leveled = level.NewFilter(l.base, lvl.o)
+	l.currentLevel = lvl
+	l.leveled = level.NewFilter(l.leveled, lvl.o)
```
LeviHarrison, line 182: Could we just make the previous assignment to `leveled` the function parameter?

```suggestion
	l.leveled = level.NewFilter(log.With(l.base, "ts", timestampFormat, "caller", log.Caller(5)), lvl.o)
```

PR #345, opened 2021-11-24
Check if TLS certificate and key file have been modified

LeviHarrison:
This PR adds the capability to check the hashes of the client certificate and key files to see if they've been modified, as is already done with the CA file.

Fixes https://github.com/prometheus/prometheus/issues/9512

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -931,11 +931,14 @@ func (t *tlsRoundTripper) getTLSFilesWithHash() ([]byte, []byte, []byte, []byte,
 	}
 	h1 := sha256.Sum256(b1)
 
-	b2, b3, err := readCertAndKey(t.certFile, t.keyFile)
-	if err != nil {
-		return nil, nil, nil, nil, err
+	var h2, h3 [32]byte
+	if t.certFile != "" {
+		b2, b3, err := readCertAndKey(t.certFile, t.keyFile)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+		h2, h3 = sha256.Sum256(b2), sha256.Sum256(b3)
 	}
-	h2, h3 := sha256.Sum256(b2), sha256.Sum256(b3)
 
 	return b1, h1[:], h2[:], h3[:], nil
 }
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -833,12 +833,12 @@ func (c *TLSConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // readCertAndKey reads the cert and key files from the disk.
 func readCertAndKey(certFile, keyFile string) ([]byte, []byte, error) {
-	certData, err := os.ReadFile(certFile)
+	certData, err := ioutil.ReadFile(certFile)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	keyData, err := os.ReadFile(keyFile)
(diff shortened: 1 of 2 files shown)


roidelapluie:
I think we should only store the cert, you can not change the key without changing the cert.

LeviHarrison:
The issue with that would be if someone were to change only the key, yes, the TLS configuration would be invalid, but instead of getting (what I think would be) a `private key does not match public key` error, they would continue to get whatever error they had before, or nothing, because the configuration wouldn't be updated.

config/http_config.go
```diff
@@ -785,54 +815,69 @@ func NewTLSRoundTripper(
...
-	tlsConfig := t.tlsConfig.Clone()
-	if !updateRootCA(tlsConfig, b) {
-		return nil, fmt.Errorf("unable to use specified CA cert %s", t.caFile)
-	}
-	rt, err = t.newRT(tlsConfig)
-	if err != nil {
-		return nil, err
+	// The cert and key files are read separately by the client
+	// using GetClientCertificate, therefore the RoundTripper
+	// doesn't need to be updated if only they are changed.
```
roidelapluie, line 861: It actually does, otherwise it would not recreate new connections. t.CloseIdleConnections() is useless if other scrapes use the same http client and are in progress.

config/http_config.go
```diff
@@ -785,33 +815,41 @@ func NewTLSRoundTripper(
...
-	b, err := readCAFile(t.caFile)
+func (t *tlsRoundTripper) getTLSFilesWithHash() ([]byte, []byte, []byte, []byte, error) {
+	b1, err := readCAFile(t.caFile)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, nil, err
+	}
+	h1 := sha256.Sum256(b1)
+
+	b2, b3, err := readCertAndKey(t.certFile, t.keyFile)
```
roidelapluie, line 833: Where do we deal with unset certFile and t.keyFile?

LeviHarrison: Could you please elaborate on what you mean by unset?

roidelapluie: I mean empty strings

LeviHarrison: Good point. A round tripper just isn't created if the `CAFile` is unset, but we can't do the same with the `certFile` and `keyFile`.

https://github.com/prometheus/common/blob/102c12a0a1cca70d642421c7cb151eaa701295bd/config/http_config.go#L452-L455

fpetkovski:
We are also hitting this problem, anything we can help to move the PR forward?

LeviHarrison:
Apologies, I just need to find time to work on this.

config/http_config.go
LeviHarrison: Should be fixed.

simonpasquier:
@LeviHarrison do you want me to pick up the work? This bug is becoming a significant issue for us right now.

simonpasquier:
Rebased on top of main, I didn't change anything to the PR except for resolving the conflicts. Thanks a lot @LeviHarrison!

simonpasquier:
@roidelapluie PTAL

simonpasquier:
@roidelapluie friendly ping :)

roidelapluie:
Thanks!

PR #350, opened 2021-12-21
Set HTTP hostname based on TLS server name

fpetkovski:
The net/http library uses the Host field from the Request object in
order to determine the value of the Host header [1]. In order for the
Prometheus client to support SNI, it needs to set this field to the
value provided in the TLS server name.

[1] https://github.com/golang/go/issues/29865

Signed-off-by: fpetkovski <filip.petkovsky@gmail.com>

Fixes https://github.com/prometheus/prometheus/issues/9403

### DIFF
--- a/config/http_config_test.go
+++ b/config/http_config_test.go
@@ -37,7 +37,7 @@ import (
 	"testing"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


roidelapluie:
Hello,

Thank you for your contribution.
I am surprised that SNI does not work out of the box.

I am not willing to have this repository depend on client_golang. We have an exception for an interface in version, and the tests in sigv4, but that's it.

fpetkovski:
Thanks for the feedback. Could you point me to where the dependency to client_golang is introduced in this PR? Do you mean because we are referencing the TLS config?

bwplotka (approved):
Hm, I don't see any client_golang dep @roidelapluie  🤔 

Looks good to me, but let's wait for @roidelapluie voice, maybe I am missing something.

config/http_config_test.go
```diff
@@ -179,6 +179,27 @@ func TestNewClientFromConfig(t *testing.T) {
...
 				fmt.Fprint(w, ExpectedMessage)
 			},
+		}, {
+			clientConfig: HTTPClientConfig{
+				TLSConfig: TLSConfig{
+					CAFile:             TLSCAChainPath,
+					CertFile:           ClientCertificatePath,
+					KeyFile:            ClientKeyNoPassPath,
+					ServerName:         "test-domain.com",
+					InsecureSkipVerify: true},
```
bwplotka, line 189:
```suggestion
					InsecureSkipVerify: true,
				},
```

roidelapluie:
The initial commit introduced a dependency. Then it was force pushed.


I can not merge this as is, I need to check that 1. SNI works even with our connection pooling and 2. This does not break blackbox exporter, which has similar logic.

fpetkovski:
Ah yes, that seems to be correct: https://github.com/prometheus/common/commit/038724fffa024b190650f1cf95adaf7fd57eaa96#diff-cdb28b391884a82ef481177749e35ba9c3c54bda95d93ffe3d5f3ba1c67c2c5eR35

roidelapluie:
after thinking twice, the TLS servername should come from the header, not the other way around.

fpetkovski:
What do you mean by this? Isn't the TLS server name a user provided config?

SuperQ:

PR #364, opened 2022-03-18
Deduplicate slashes for sigv4 signature

importhuman:
Signed-off-by: Ujjwal Goyal <importujjwal@gmail.com>

For issue [10200 on the prometheus repo](https://github.com/prometheus/prometheus/issues/10200).
Not sure if this is all that is required.

### DIFF
--- a/sigv4/sigv4.go
+++ b/sigv4/sigv4.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/textproto"
+	"path"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
Can we have a test?

sigv4/sigv4.go
```diff
@@ -115,6 +115,9 @@ func (rt *sigV4RoundTripper) RoundTrip(req *http.Request) (*http.Response, error
 	}()
 	req.Body = ioutil.NopCloser(seeker)
 
+	// Escape double slashes if present
```
roidelapluie, line 118:
```suggestion
	// Escape URL like documented in <xxx>.
```

importhuman:
Did you mean a test like this?

It's failing currently, because it's not being escaped. `req.URL.EscapedPath()` uses the `RawPath` of the URL, which is coming out to be `""`, in which case it doesn't escape. Any suggestions?

SuperQ (approved):
LGTM

importhuman:
Were the tests not run in CI, because I expected the test I added to fail :thinking:
Don't think this is ready to merge.

SuperQ:
Yes, we run a `make test` on this package in CI.

SuperQ:
Looks like the `make test` incorrectly skips `go test`. 

Fixed in https://github.com/prometheus/common/pull/365.

kakkoyun (approved):

importhuman:
Ah, it's inside circleci, didn't realize that.

Screenshot from 2022-03-28 17-24-51
This is what I'm concerned about. Why is it passing on the CI?

SuperQ:
Ok, a rebase here will make the tests run correctly.

roidelapluie:
It turns out that the underlying API is already calling EscapePath. The fix for the user's issue is to use path.Clean. I have updated the pull request to reflect this and added a test.

WDYT @importhuman ?

importhuman:
LGTM @roidelapluie

roidelapluie:
Thanks!!!

PR #371, opened 2022-04-13
go.mod: Exclude prometheus/common v0.32.1

mrueg:
This is mainly a cosmetical change, see also https://github.com/prometheus/client_golang/issues/1012
As suggested in https://github.com/prometheus/client_golang/issues/1012#issuecomment-1090482644

I'll update this one once there's a new release of client_golang.


roidelapluie:
This seems to be pure technical debt put on maintainers...

SuperQ:
Yes, I don't think this is something we want to do. It's just going to cause a lot of toil.

PR #373, opened 2022-04-15
TLS config: Enable selection of min and max TLS version

roidelapluie:
go1.18 changes the default minimum TLS version to 1.2.

In order not to break our users, let's make the default minimum version
configurable, with conservative defaults.

The allowed values (TLS10, ..) come from the exporter-toolkit:
https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

TLSVersion is exported so the exporter toolkit can reuse them later.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -36,23 +36,61 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// DefaultHTTPClientConfig is the default HTTP client configuration.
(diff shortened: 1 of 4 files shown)


roidelapluie:
Hold on. We might want the default to be tls1.2 and change it in Prometheus itself.

roidelapluie:
Closed for #375

PR #375, opened 2022-04-18
TLS config: Enable selection of min TLS version

roidelapluie:
go1.18 changes the default minimum TLS version to 1.2.

Let's make the default minimum version configurable, while following go
default.

The allowed values (TLS10, ..) come from the exporter-toolkit:
https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

TLSVersion is exported so the exporter toolkit can reuse them later.


config/http_config.go
```diff
@@ -36,25 +36,58 @@ import (
...
+type TLSVersion uint16
+
+var TLSVersions = map[string]TLSVersion{
+	"TLS13": (TLSVersion)(tls.VersionTLS13),
+	"TLS12": (TLSVersion)(tls.VersionTLS12),
+	"TLS11": (TLSVersion)(tls.VersionTLS11),
+	"TLS10": (TLSVersion)(tls.VersionTLS10),
+}
+
+func (tv *TLSVersion) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
Nexucis, line 69: other configurations/ struct in this file are supporting json. I guess you should also override the method `UnmarshalJSON` and `MarshalJSON`

config/http_config.go
```diff
@@ -714,6 +750,8 @@ type TLSConfig struct {
 	ServerName string `yaml:"server_name,omitempty" json:"server_name,omitempty"`
 	// Disable target certificate validation.
 	InsecureSkipVerify bool `yaml:"insecure_skip_verify" json:"insecure_skip_verify"`
+	// Minimum TLS version.
+	MinVersion TLSVersion `yaml:"min_version,omitempty"`
```
Nexucis, line 754: it is missing here the json tag

```suggestion
MinVersion TLSVersion `yaml:"min_version,omitempty" json:"min_version,omitempty"`
```

Nexucis: if `MinVersion` is not mandatory, don't you want to provide a default value ?

You could do it by overriding the Unmarshal's method.

Nexucis (commented):
Overall it looks ok. I put a couple of comments / notes.

config/http_config.go
roidelapluie: No, I want to use go default. I think we should let them define the default while providing or users a way to divert if really needed.

config/http_config.go
roidelapluie: :(

config/http_config.go
Nexucis: mmm ok. And I guess in Prometheus you will define that the default / Minimum version is the previous one used, right ?

roidelapluie: Nope, the plan is to mark it as a CHANGE in the changelog.

I think out users will benefit from defaults that go runtime provides. We should probably not make the call.

Possibly it can break some users, but they have the option to opt-out this change by explicitly setting tls1.0 as supported version.

config/http_config.go
```diff
@@ -36,25 +36,84 @@ import (
...
+func (tv *TLSVersion) MarshalJSON() ([]byte, error) {
+	if tv != nil || *tv == 0 {
+		return []byte("null"), nil
+	}
+	for s, v := range TLSVersions {
+		if *tv == v {
+			return []byte(s), nil
+		}
+	}
+	return nil, fmt.Errorf("unknown TLS version: %d", tv)
```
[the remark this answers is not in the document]
roidelapluie: aligned

config/http_config.go
Nexucis: I can agree on this assumption and I agree it's better to use the default tls version supported by Go.
I'm just wondering how much it will impact the users.
Let's see if someone answers to your email.

PR #380, opened 2022-05-17
Add ability to sign requests for all AWS services

transacid:
This add the ability to utilize sigv4 signing for all AWS services not
just "aps". When the newly introduced property "service" is not set in
config it will default to "aps".

I found it hard to think of a way to test this. When you have ideas, please let me know.

### DIFF
--- a/sigv4/sigv4.go
+++ b/sigv4/sigv4.go
@@ -37,9 +37,10 @@ var sigv4HeaderDenylist = []string{
 }
 
 type sigV4RoundTripper struct {
-	region string
-	next   http.RoundTripper
-	pool   sync.Pool
+	region  string
+	next    http.RoundTripper
+	pool    sync.Pool
+	service string
 
 	signer *signer.Signer
 }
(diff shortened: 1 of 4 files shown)


roidelapluie:
Is it relevant for Prometheus?

Le mar. 17 mai 2022, 13:37, Boris Petersen ***@***.***> a
écrit :

transacid:
yes it is. I'd also had to change the documentation.
Some words about the reasoning behind this PR:
We are running our TSDB in an EKS cluster in AWS and would like to use API gateway for authentication. To sign requests for the API gateway the "service" needs to be "exectute-api". However prometheus hard codes it to eb "aps.

transacid:
hey, is there anything I can do? something that is missing?

krmnn:
This patch enables users of your client library who operate Prometheus (or compatible services) *on their own* in AWS and want to protect the exposed endpoints with the help of AWS IAM instead of static http basic_auth credentials. 

We want to use AWS APIGateway for that, so the requests must be signed for the service 'execute-api' to be valid. But also other frontends (e.g. load balancers) are thinkable.

Currently the signing process is tied to only support AWS Managed Service for Prometheus. This PR increases the flexibility of your users.

SuperQ:

PR #382, opened 2022-05-27
Update Go modules

SuperQ:
* Update Go testing to 1.18.
* Update Go minimum version to 1.17.
* Update Go module format for 1.17.
* Bump Go modules.
* Update yaml.v3 use.

Signed-off-by: SuperQ <superq@gmail.com>


.circleci/config.yml
```diff
@@ -13,7 +13,10 @@ jobs:
         type: boolean
         default: true
     docker:
-      - image: circleci/golang:<< parameters.go_version >>
+      - image: cimg/go:<< parameters.go_version >>
+    environment:
+      # Override Go 1.18 security deprecations.
+      GODEBUG: "x509sha1=1,tls10default=1"
```
kakkoyun, line 19:
```
      GODEBUG: "x509sha1=1,tls10default=1"
```
So this does the trick for certificate failures of the tests! Is there documentation somewhere for these parameters?

kakkoyun (approved):
LGTM :100:

kakkoyun:
> I am against the move to yamlv3 because we can't move prometheus at this moment.

It doesn't seem to have any API changes. I'm curious what's the reason behind it? Is there something I can check out?

SuperQ:
The only major change is that "Strict" loading doesn't work as it did before.

.circleci/config.yml
SuperQ: It's in the 1.18 release notes.

roidelapluie:
Thanks, can we update minimum version of client_golang first?

mrueg:
@roidelapluie see: https://github.com/prometheus/client_golang/commit/e38d614cd66e92b71ac44ff012a1b362f1a5680b

SuperQ:
I think we need to have a client_golang release and then I can bump the dependency here.

kakkoyun:
> I think we need to have a client_golang release and then I can bump the dependency here.

@SuperQ client_golang released https://github.com/prometheus/client_golang/releases/tag/v1.13.0

.github/workflows/golangci-lint.yml
```diff
@@ -20,8 +20,11 @@ jobs:
       - name: install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17.x
+          go-version: 1.18.x
```
kakkoyun, line 23: If I'm not mistaken, this file synced from Prometheus's main repo. 

We might need to change it in there, no?

https://github.com/prometheus/prometheus/blob/main/scripts/sync_repo_files.sh

kakkoyun (commented):
Only have a question about the workflow file otherwise LGTM

.github/workflows/golangci-lint.yml
SuperQ: This is actually doing the sync from upstream, since we need some of the other updates in this PR to make it pass CI.

SuperQ: https://github.com/prometheus/common/pull/379 was the last prombot sync, couldn't pass CI.

PR #385, opened 2022-06-17
Add more mimetypes

roidelapluie:

### DIFF
--- a/server/static_file_server.go
+++ b/server/static_file_server.go
@@ -24,6 +24,8 @@ var mimeTypes = map[string]string{
 	".png": "image/png",
 	".jpg": "image/jpeg",
 	".gif": "image/gif",
+	".ico": "image/x-icon",
+	".ttf": "font/ttf",
 }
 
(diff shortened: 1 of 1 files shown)


Nexucis:
to avoid the same situation, can you add more mimes. For example this list should cover all cases we have in React: 

https://github.com/perses/perses/blob/main/internal/api/front/generate.go#L31-L43

Nexucis:
by checking on https://developer.mozilla.org/fr/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types, it looks like my MimeTypes are not accurate for JS and ICO but for the rest they are correct.

If you can update the list, we should be good for a while

roidelapluie:
cjs
css
eot

ico
js
json
less
map
svg
ttf
txt
woff
woff2


is what we have. I have removed html from the list for now.

Nexucis (approved):
lgtm.

Just in case, the react-app has an `index.html`. SO HTML file exist. But I think anyway we are good like that. HTML files seems to be naturally supported somehow lol.

Issue #384, opened 2022-06-15
Feature Request: Ability to set custom User-Agent on oauth2 


cig2312:
Currently the User-Agent on the oauth2 client is `Go-http-client/1.1`. 
This leads to the oauth2 flow being ratelimited on many sites. 

Having the option of setting a custom UA would resolve this issue.

roidelapluie:
Can you please be more specific with the issue you try to solve? I don't think rate limiting on UA is something useful.

PR #386, opened 2022-06-28
Add User-Agent header to oauth2 requests

cig2312:
This PR adds the ability to specify the User-Agent header for oauth2 requests. 

The default User-Agent: `go-http-client/1.1` is ratelimited by many sites which leads to a lot of oauth2 requests failing. 

Fixes: [384](https://github.com/prometheus/common/issues/384)

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -224,6 +224,8 @@ type OAuth2 struct {
 	ProxyURL URL `yaml:"proxy_url,omitempty" json:"proxy_url,omitempty"`
 	// TLSConfig is used to connect to the token URL.
 	TLSConfig TLSConfig `yaml:"tls_config,omitempty"`
+	// UserAgent is used to set a custom User-Agent http header while making the oauth request.
+	UserAgent string `yaml:"user_agent,omitempty" json:"user_agent,omitempty"`
 }
 
 // SetDirectory joins any relative file paths with dir.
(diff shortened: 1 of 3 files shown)


roidelapluie:
Can you please be more specific with the issue you try to solve?

cig2312:
I have updated the PR description
but the problem is that the default User-Agent: `go-http-client/1.1` is ratelimited by many sites which leads to a lot of oauth2 requests failing. 

My usecase: 
Probing some oauth2 authenticated endpoints via the blackbox prober. Because of the default client in the UA my probes are failing.

roidelapluie:
I don't think we should expose this to users. I agree we could pass on the Blackbox Exporter and Prometheus user agents.

Can you move to code to an http option like WithIdleConnTimeout ?

cig2312:
I did think about this but there are a couple of reasons i wanted to expose this to users of this pkg:
- The reason the default http UA is limited is because of the sheer number of requests. While this wouldn't be the case with Blackbox Exporter/Prometheus UA's, i worry it may still trigger limits on sites that have set them very aggressively. 
- Having something custom also lets us group/filter out requests based on UA on our metrics etc. 
-  While i know this is a common library for all of prometheus, the blackbox prober does allow the user the specify the UA. I tried to follow the same convention. 

Maybe we can keep a default UA(prometheus/blackbox) and allow the user the customise if needed?

roidelapluie:
In BBE's case, it would be exposed to the user, but not directly with the http config.

cig2312:
Isn't the `proxy_url` in the oauth2 config also similar? That is exposed to the users with the http config.

roidelapluie:
> Isn't the `proxy_url` in the oauth2 config also similar? That is exposed to the users with the http config.

Yes, but it is valid for users to change it.

User agent is already set in many cases, I think we should just make sure to move it here with an http option so it's also applied to oauth 2.

cig2312:
> Can you move to code to an http option like WithIdleConnTimeout ?

Do you mean adding it as an optional func to the http client? 
I dont think that would work as the client generated by `NewClientFromConfig` isnt reused by the oauth roundtripper.

roidelapluie:
What about https://github.com/prometheus/common/pull/387

cig2312:
LGTM

PR #389, opened 2022-07-08
config: ignore deprecated warning in tests

simonpasquier:
(*CertPool).Subjects is deprecated in Go 1.18 because it may not include
the system certs but it isn't a problem for the tests because we're not
loading the system certs.

Follow-up to #388 which we would need to upgrade to the latest golangci-lint version.


roidelapluie:
Thanks!

PR #390, opened 2022-07-08
OAuth2: Respect disable keepalives option; Implement close idle connections

roidelapluie:


1337andre:
I think it was fixed with 0.22.0. is running for 2 days without any problems

PR #391, opened 2022-07-14
Improve OAuth2 user agent handling

roidelapluie:
Instead of relying on the User Agent in the configuration, take the User
Agent of the original request. This enables compatibility with existing
user agents.


Nexucis (approved):
👍

Issue #393, opened 2022-07-17
Add username_file in prometheus.yaml configuration


lushaorong:
I want to try this work https://github.com/prometheus/prometheus/issues/11007, and it needs to modify config/http_ config，so i open this issues

roidelapluie:
I don't think we should do this. Prometheus only reads secrets from files. Username is not considered a secret.

I am also concerned by the inability to read username and password files atomically.

PR #396, opened 2022-07-30
Add username_file in BasicAuth

lushaorong:


lushaorong:
Fixes #393 .

Changes proposed in this pull request:
- Add username_file in BasicAuth

roidelapluie:
I don't think we should do this. Prometheus only reads secrets from files. Username is not considered a secret.

I am also concerned by the inability to read username and password files atomically.

lushaorong:
Maybe you are right. I will close

PR #403, opened 2022-08-30
Add platform info to labels

lucacome:
Adds the information about the platform to the list of labels

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -31,6 +31,8 @@ var (
 	BuildUser string
 	BuildDate string
 	GoVersion = runtime.Version()
+	GoOS      = runtime.GOOS
+	GoArch    = runtime.GOARCH
 )
 
 // NewCollector returns a collector that exports metrics about current version
@@ -41,7 +43,7 @@ func NewCollector(program string) prometheus.Collector {
 			Namespace: program,
 			Name:      "build_info",
 			Help: fmt.Sprintf(
-				"A metric with a constant '1' value labeled by version, revision, branch, and goversion from which %s was built.",
+				"A metric with a constant '1' value labeled by version, revision, branch, goversion from which %s was built, and the goos and goarch for the build.",
 				program,
 			),
 			ConstLabels: prometheus.Labels{
@@ -49,6 +51,8 @@ func NewCollector(program string) prometheus.Collector {
 				"revision":  getRevision(),
 				"branch":    Branch,
 				"goversion": GoVersion,
+				"goos":      GoOS,
+				"goarch":    GoArch,
 			},
 		},
 		func() float64 { return 1 },
@@ -74,7 +78,7 @@ func Print(program string) string {
 		"buildUser": BuildUser,
 		"buildDate": BuildDate,
 		"goVersion": GoVersion,
-		"platform":  runtime.GOOS + "/" + runtime.GOARCH,
+		"platform":  GoOS + "/" + GoArch,
 	}
 	t := template.Must(template.New("version").Parse(versionInfoTmpl))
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
I have ping other people to get an opinion on this.

version/info.go
```diff
@@ -41,14 +42,15 @@ func NewCollector(program string) prometheus.Collector {
 			Namespace: program,
 			Name:      "build_info",
 			Help: fmt.Sprintf(
-				"A metric with a constant '1' value labeled by version, revision, branch, and goversion from which %s was built.",
+				"A metric with a constant '1' value labeled by version, revision, branch, goversion and platform from which %s was built.",
```
juliusv, line 45: The "from which %s was built" part is now not fully correct anymore, since the platform it was built on is not always the platform it was built for (and is running on). How about keeping the original string but then ending it as "[...] goversion from which %s was built, and the target platform for the build."

version/info.go
```diff
@@ -31,6 +31,7 @@ var (
 	BuildUser string
 	BuildDate string
 	GoVersion = runtime.Version()
+	Platform  = runtime.GOOS + "/" + runtime.GOARCH
```
juliusv, line 34: Two labels might be even better? After all, if you just want to select by Linux or architecture, that'll make things even easier then.

SuperQ: +1, let's split this into separate fields.

lucacome: what would the fields be? `OS` and `Architecture`? Is it okay to leave it as one field in line 79 for the version information?

beorn7:
Looks good to me.

The tests seem to fail because of a problem upstream in prometheus/common. Maybe a rebase will help?

version/info.go
```diff
@@ -41,14 +43,16 @@ func NewCollector(program string) prometheus.Collector {
...
-				"A metric with a constant '1' value labeled by version, revision, branch, and goversion from which %s was built.",
+				"A metric with a constant '1' value labeled by version, revision, branch, goversion from which %s was built, and the goos and goarch for the build.",
 				program,
 			),
 			ConstLabels: prometheus.Labels{
 				"version":   Version,
 				"revision":  Revision,
 				"branch":    Branch,
 				"goversion": GoVersion,
+				"os":        GoOS,
```
juliusv, line 54: @roidelapluie Since this is going to affect a lot of components, just double-checking: are you good with just `os` and `arch` here vs. `goos` / `goarch`? Maybe we should use the latter in case the system is an `amd64` one but the binary is compiled for `386`, so it could be confusing to people if they think this label designates the arch of the underlying system?

lucacome: I think I'm a little bit confused then and why I wasn't sure about the help message 😅 
The binary can be compiled for a different arch? I assumed the arch would match the one of the system it's running on.

juliusv: So you can totally run a Go binary compiled for the `386` architecture on the (backwards-compatible but newer) `amd64` architecture. You can try it by doing:

```bash
GOARCH=386 go build main.go
./main.go
```

...on an `amd64` system. It's not a common thing at all, but in that case, just `arch` might be misleading.

roidelapluie:
Note that prometheus is already displaying this:

```
prometheus, version 2.37.4 (branch: release-2.37, revision: a9a24ba58c901a0972e4c38d5b0d2502951de16d)
  build user:       roidelapluie@nixos
  build date:       20221206-08:16:58
  go version:       go1.19
  platform:         linux/amd64
```

version/info.go
lucacome: updated it

juliusv:
Thanks! :+1:

PR #406, opened 2022-10-14
Fix bug for TLSVersion.MarshalYAML

StevenYCChou:
Signed-off-by: Yen-Cheng Chou <3055688+StevenYCChou@users.noreply.github.com>

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -106,7 +106,7 @@ func (tv *TLSVersion) UnmarshalJSON(data []byte) error {
 
 // MarshalJSON implements the json.Marshaler interface for TLSVersion.
 func (tv *TLSVersion) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


StevenYCChou:
@roidelapluie PTAL.

simonpasquier (commented):
While there's a bug, I'm not sure that the fix is correct. I believe that `MarshalJSON()` needs to be changed to a value receiver function to work properly.
I've experiment a bit with the Go playground:
https://go.dev/play/p/P9nFVvIu-Mb

roidelapluie:
> While there's a bug, I'm not sure that the fix is correct. I believe that `MarshalJSON()` needs to be changed to a value receiver function to work properly. I've experiment a bit with the Go playground: https://go.dev/play/p/P9nFVvIu-Mb

I agree

Issue #402, opened 2022-08-23
Support ProxyConnectHeader


mem:
The roundtripper created using NewRoundTripperFromConfig supports proxies, but it doesn't support proxy-specific headers using `ProxyConnectHeader`. This is simply an option that could be passed to the configuration and added to the transport that is created. Going thru the Go code in the standard library, it _looks like_ this only has effect if the connection method is `CONNECT`.

PR #409, opened 2022-11-11
Add support for proxy connect headers

mem:
Some proxy configurations require additional headers to be able to use them (e.g. authorization token specific to the proxy).

Fixes: #402

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"encoding/json"
+	"net/http"
 	"path/filepath"
 )
 
@@ -46,6 +47,29 @@ func (s Secret) MarshalJSON() ([]byte, error) {
 		return json.Marshal("")
 	}
 	return json.Marshal(secretToken)
+}
+
+type Header map[string][]Secret
+
+func (h *Header) HTTPHeader() http.Header {
+	if h == nil || *h == nil {
+		return nil
+	}
+
+	header := make(http.Header)
+
+	for name, values := range *h {
+		var s []string
+		if values != nil {
+			s = make([]string, 0, len(values))
+			for _, value := range values {
+				s = append(s, string(value))
+			}
+		}
+		header[name] = s
+	}
+
+	return header
 }
 
 // DirectorySetter is a config type that contains file paths that may
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -14,8 +14,13 @@
 package config
 
 import (
+	"bytes"
 	"encoding/json"
+	"net/http"
+	"reflect"
 	"testing"
+
+	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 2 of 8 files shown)


mem:
Ok, this is weird. The files use ioutil, but they don't import it. I don't see the change that introduced the usage introducing the import, so it's not like the import was accidentally removed. I don't understand how this even compiles...

roidelapluie:
This is explicitly for secrets, so it would be a security breach to allow this to be marshalled back. We have to use our `secrets`, which I think would be better to duplicate the auth parameters as we have them now, with credentials and basic_auth.

config/config.go
```diff
@@ -48,6 +49,29 @@ func (s Secret) MarshalJSON() ([]byte, error) {
 	return json.Marshal(secretToken)
 }
 
+type Header map[string][]Secret
```
mem, line 52: this is added mostly to be able to test the conversion to http.Header and the marshaling / unmarshaling.

mem:
@roidelapluie I would appreciate it if you can take another look. Thanks!

config/config.go
```diff
@@ -48,6 +49,29 @@ func (s Secret) MarshalJSON() ([]byte, error) {
...
+
+func (h Header) HttpHeader() http.Header {
```
roidelapluie, line 54:
```suggestion
func (h Header) HTTPHeader() http.Header {
```

config/config_test.go
```diff
@@ -51,3 +56,110 @@ func TestJSONMarshalSecret(t *testing.T) {
 		})
 	}
 }
+
+func TestHeaderHttpHeader(t *testing.T) {
```
roidelapluie, line 60:
```suggestion
func TestHeaderHTTPHeader(t *testing.T) {
```

roidelapluie (commented):
We also support JSON, json tests would be welcome

roidelapluie (approved):

roidelapluie:
Thanks!

config/config.go
```diff
@@ -48,6 +49,29 @@ func (s Secret) MarshalJSON() ([]byte, error) {
...
+
+	for name, values := range *h {
+		var s []string
+		if values != nil {
+			s = make([]string, 0, len(values))
+			for _, value := range values {
+				s = append(s, string(value))
+			}
+		}
+		header[name] = s
```
dswarbrick, line 69: This code has a slightly odd smell to me. Why not simply use `header.Add(name, string(value))` (https://pkg.go.dev/net/http#Header.Add)?

The test cases would need to be adjusted, since the bogus items like nil header names or empty value lists won't be added, as they would be illegal headers - and as such I don't see the point of including them as test cases anyway. Header.Add() will also canonicalize the header names.

mem: oh, yeah, I completely missed that. I'll follow up with another PR.

PR #412, opened 2022-11-14
Version Bump     gopkg.in/yaml.v2 v2.4.0 -->     gopkg.in/yaml.v3 v3.0.1

Lepidopteron:
Version Bump of gopgk.in/yaml.v2  v3 is being maintained and has been released 4 years ago.


mem:
While this is a desired change (I have a bunch of code that uses v3, but pulls v2 into the binary because of this dependency) we need to make sure existing configurations are loaded correctly. Durations changed so that in v3 they are strings. I'm not sure how that impacts / relates to this change.

tobiasjaster:
Are there any updates on this?

PR #415, opened 2022-11-26
feat: make LoadHTTPConfigFile set directory and move from tests file

FUSAKLA:
Move away the functions from the test files since they are not only test helper.
For motivation, see https://github.com/prometheus/prometheus/pull/11487

Also add setting directory `cfg.SetDirectory(filepath.Dir(filepath.Dir(filename)))` as used in Alertmanager see 
https://github.com/prometheus/alertmanager/blob/1a9f55b939ab81b86428c013ae294ad80779c9b6/cli/config/http_config.go#L36

Prometheus does not use the `LoadHTTPConfigFile` function AFAIK, so the change in setting the directory won't break anything, hopefully. 

Other possibility is to avoid setting the directory and fix this on the Alertmanager side outside the common library. 
That might be safer (function is exported, so it could be breaking change for someone, possibly?) 

@roidelapluie PTAL

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thanks!

FUSAKLA:
Thanks, @roidelapluie, for merging! 

Should I wait with using of the functions in Prometheus and Alertmanager for new release/tag of this repo? 
Not sure what the release lifecycle is.

roidelapluie:
I will try to tag asap, I am reviewing a few things

Le jeu. 8 déc. 2022, 13:38, Martin Chodur ***@***.***> a
écrit :

PR #417, opened 2022-12-08
Add histogram vectors to model

zenador:
Addresses https://github.com/prometheus/client_golang/issues/1179

Tested the JSON decoding successfully by scraping a simple server running native histograms with Prometheus, then using the HTTP API in `client_golang` (which does not need to be updated, only this library needs updating) to query it.


beorn7:
Cool. I hope I can help with a review soon…

model/value.go
```diff
@@ -221,80 +187,74 @@ func (s Samples) Equal(o Samples) bool {
 
 // SampleStream is a stream of Values belonging to an attached COWMetric.
 type SampleStream struct {
-	Metric Metric       `json:"metric"`
-	Values []SamplePair `json:"values"`
+	Metric     Metric                `json:"metric"`
+	Values     []SamplePair          `json:"values"`
+	Histograms []SampleHistogramPair `json:"histograms"`
+	Type       SampleType
```
beorn7, line 193: I fairly certain that we have to support SampleStreams of mixed type. (The Prometheus query API can definitely return a series that has both.) I think you can just get rid of the `Type` field. The usual case of a pure histogram SampleStream or a pure float SampleStream would then simply be represented by the corresponding other slice being empty. A mixed SampleStream would have both slices populated. (And that matches the semantics of the JSON response in the query API, IIRC.)

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (v FloatString) Equal(o FloatString) bool {
+	return v == o
+}
+
+type HistogramBucket struct {
+	Boundaries int
+	Lower      FloatString
+	Upper      FloatString
+	Count      IntString
```
beorn7, line 78: The count can be a float, actually. Histograms can be so-called "float histograms". (Result of the `rate` function, but it could even happen in instrumentation, as K8s has proven.)

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+
+package model
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+)
+
+type IntString uint64
```
beorn7, line 22: I don't think we need that. All the quoted numbers are actually floats, even if they mostly look like integers.

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	for i, bucket := range s {
+		if !bucket.Equal(o[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+type SampleHistogram struct {
+	Count   IntString        `json:"count"`
```
beorn7, line 133: This can be a float, see above.

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	Sum     FloatString      `json:"sum"`
+	Buckets HistogramBuckets `json:"buckets"`
+}
+
+func (s SampleHistogram) String() string {
+	return fmt.Sprintf("Count: %d, Sum: %f, Buckets: [redacted]", s.Count, s.Sum)
```
beorn7, line 139: Do you plan to work on the "redacted" part? In the prometheus/prometheus repo, there is already code for string representation.

model/value.go
```diff
@@ -38,82 +30,20 @@ var (
...
 // Sample is a sample pair associated with a metric.
 type Sample struct {
-	Metric    Metric      `json:"metric"`
-	Value     SampleValue `json:"value"`
-	Timestamp Time        `json:"timestamp"`
+	Metric    Metric          `json:"metric"`
+	Value     SampleValue     `json:"value"`
+	Timestamp Time            `json:"timestamp"`
+	Histogram SampleHistogram `json:"histogram"`
+	Type      SampleType
```
beorn7, line 46: Maybe you can get away without this `Type` field? `Histogram` could be a `*SampleHistogram`, and then the contract could be that a `Sample` with `Histogram == nil` is a float sample, and one with `Histogram != nil` is a histogram sample.

model/value_histogram.go
zenador: Noted, will change all `IntString`s to `FloatString`s, and the corresponding ints in equivalent Mimir structs to floats. The counts will still always be positive right? Or perhaps possibly negative if there's a rate? I guess for simplicity we'll just use float histograms for everything even though they are rarer, since we can cast from int to float but not reverse?

model/value_histogram.go
zenador: Sure, I was just leaving it to later since this can probably be changed without affecting any other code. Thanks for the tip, will look for it!

model/value.go
zenador: Oh I didn't know that, I've only observed JSON responses with either values or histograms, not both. In what cases would it have both? How can we instrument a server that sends mixed type metrics, or write a query that returns both? It'll help to know what these cases mean so that we can handle the corresponding logic properly in Mimir (we've been assuming only one type per time series / sample stream so far, and that if we somehow receive both, it's a mistake and we'll ignore the floats). Would the mixed type mean that the promql.Points have both V and H defined?

model/value.go
zenador: Okay, so for this one, we don't allow both a float and a histogram in a single Sample (unlike SampleStream)?

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+		return err
+	}
+	if g, e := len(tmp), wantLen; g != e {
+		return fmt.Errorf("wrong number of fields: %d != %d", g, e)
+	}
+	return nil
+}
+
+func (s *HistogramBucket) Equal(o *HistogramBucket) bool {
+	return s == o || (s.Boundaries == o.Boundaries && s.Lower.Equal(o.Lower) && s.Upper.Equal(o.Upper) && s.Count.Equal(o.Count))
```
codesome, line 114: The `.Equal` method is just doing a `==` check. Do we need that method?

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+func (s *SampleHistogram) Equal(o *SampleHistogram) bool {
+	return s == o || (s.Count.Equal(o.Count) && s.Sum.Equal(o.Sum) && s.Buckets.Equal(o.Buckets))
+}
+
+type SampleHistogramPair struct {
+	Timestamp Time
+	Histogram SampleHistogram
+}
+
+func (s SampleHistogramPair) MarshalJSON() ([]byte, error) {
```
codesome, line 151: Same comment about making this marshaling more efficient iff required.

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	return []byte(fmt.Sprintf("[%s,%s]", t, v)), nil
+}
+
+func (s *SampleHistogramPair) UnmarshalJSON(buf []byte) error {
+	tmp := []interface{}{&s.Timestamp, &s.Histogram}
+	wantLen := len(tmp)
+	if err := json.Unmarshal(buf, &tmp); err != nil {
+		return err
+	}
+	if g, e := len(tmp), wantLen; g != e {
```
codesome, line 169: Same comment as above about the slice shrinking and variable naming

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return nil
+}
+
+func (s SampleHistogramPair) String() string {
+	return fmt.Sprintf("%s @[%s]", s.Histogram, s.Timestamp)
+}
+
+func (s *SampleHistogramPair) Equal(o *SampleHistogramPair) bool {
+	return s == o || (s.Histogram.Equal(&o.Histogram) && s.Timestamp.Equal(o.Timestamp))
```
codesome, line 180: Same comment as above about `.Equal` method

codesome (commented):
Looking good overall. I just have a lot of nits and agree with Beorn's comment about not requiring the `Type` field.

PS: this review does not cover the unit tests.

model/value.go
codesome: > Would the mixed type mean that the promql.Points have both V and H defined?

No. With mixed type we mean that in a stream of samples a sample can be float, and the next sample can be a histogram for example. It is possible when someone is migrating from old histogram (with float series) to new histograms, or conflicting names for instrumentation (a user errors).

Example of what we mean, for the above struct (basically both samples can co-exist in a stream).
```
Metric: {foo="bar"}
Values: [ (100, 1.0), (200, 2.2), (300, 3.3) ]
Histograms: [ (150, h1), (250, h2) ]
```

model/value.go
codesome: > Okay, so for this one, we don't allow both a float and a histogram in a single Sample (unlike SampleStream)?

Yes. A single sample is always either float or histogram, not both. Since there are multiple samples in a stream, we can find a mix of types, but a single sample within that always has a single type.

model/value.go
beorn7: Yeah, it's definitely expected to be rare, but it is possible, and thus we have to handle it.

model/value_histogram.go
beorn7: I think the `Count` will always be positive (or zero, or NaN). But I don't think this gives us anything. Since we treat it as a float, we would be able to handle negative values, should they show up for some weird reason. (Maybe we allow a delta histogram of two gauge histograms one day, and then the values could be negative…)

And yes, as discussed elsewhere, PromQL is only acting on floats, so it is always acting on float histograms, so all the output from the HTTP query API is always float histograms, even if the float numbers in them all happen to be integral.

model/value.go
```diff
@@ -129,53 +59,89 @@ func (s *Sample) Equal(o *Sample) bool {
...
 	}
 
 	if err := json.Unmarshal(b, &v); err != nil {
 		return err
 	}
 
 	s.Metric = v.Metric
-	s.Timestamp = v.Value.Timestamp
-	s.Value = v.Value.Value
+	if v.Histogram.Timestamp != 0 {
```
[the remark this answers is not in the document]
zenador: Fixed this, but did something very hacky with a local struct called `sampleHistogramPairPtr` because I don't know how else to fix the unmarshalling with a pointer.

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	return []byte(fmt.Sprintf("[%s,%s,%s,%s]", b, l, u, c)), nil
+}
+
+func (s *HistogramBucket) UnmarshalJSON(buf []byte) error {
+	tmp := []interface{}{&s.Boundaries, &s.Lower, &s.Upper, &s.Count}
+	wantLen := len(tmp)
+	if err := json.Unmarshal(buf, &tmp); err != nil {
+		return err
+	}
+	if g, e := len(tmp), wantLen; g != e {
```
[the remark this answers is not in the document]
zenador: Fixed the second line, but I don't understand what your first question means.

model/value_histogram.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type HistogramBucket struct {
+	Boundaries int
+	Lower      FloatString
+	Upper      FloatString
+	Count      IntString
+}
+
+func (s HistogramBucket) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
zenador: Sorry, I don't know enough about JSON marshalling to understand how to do something like this :(

beorn7:
Quick note: Bumping the minimum required Go version is a bit of an issue as prometheus/client_golang has to be conservative in this regard (and it uses prometheus/common). Currently, client_golang requires go1.17. Generally, they try to support the last three Go minor releases at the very least.

In short, I think, it would be better at this time to avoid generics and only require go1.17.

model/value.go
```diff
@@ -152,30 +95,56 @@ func (s Sample) MarshalJSON() ([]byte, error) {
...
+}
+
 // UnmarshalJSON implements json.Unmarshaler.
 func (s *Sample) UnmarshalJSON(b []byte) error {
 	v := struct {
-		Metric Metric     `json:"metric"`
-		Value  SamplePair `json:"value"`
+		Metric    Metric                 `json:"metric"`
+		Value     SamplePair             `json:"value"`
+		Histogram sampleHistogramPairPtr `json:"histogram"`
```
codesome, line 123: We could probably do this as well and check for `v.Histogram != nil`. But if this gives problems, the current implementation looks fine as well.
```suggestion
		Histogram *SampleHistogramPair `json:"histogram"`
```

model/value.go
```diff
@@ -221,80 +190,76 @@ func (s Samples) Equal(o Samples) bool {
...
 }
 
 func (ss SampleStream) String() string {
-	vals := make([]string, len(ss.Values))
+	valuesLength := len(ss.Values)
+	vals := make([]string, valuesLength+len(ss.Histograms))
 	for i, v := range ss.Values {
 		vals[i] = v.String()
 	}
+	for i, v := range ss.Histograms {
```
codesome, line 204: I think we should have the samples in the order of timestamps. But grouping with type might be more helpful as well. @beorn7 do you have any opinion on this?

model/value.go
```diff
@@ -221,80 +190,76 @@ func (s Samples) Equal(o Samples) bool {
...
+	}
 	return fmt.Sprintf("%s =>\n%s", ss.Metric, strings.Join(vals, "\n"))
 }
 
-// Value is a generic interface for values resulting from a query evaluation.
-type Value interface {
-	Type() ValueType
-	String() string
+func (ss SampleStream) MarshalJSON() ([]byte, error) {
+	if len(ss.Histograms) > 0 && len(ss.Values) > 0 {
```
codesome, line 211: If we do not care about showing empty arrays in json, like `histograms: []`, we can skip these conditions and have a simple marshal (the first condition here). WDYT @beorn7, does it matter?

model/value.go
```diff
@@ -414,3 +379,8 @@ func (mat Matrix) String() string {
 
 	return strings.Join(strs, "\n")
 }
+
+type GenericSamplePair interface {
```
codesome, line 383: Is this used anywhere?

model/value_float.go
```diff
@@ -0,0 +1,105 @@
...
+	// non-existing sample pair. It is a SamplePair with timestamp Earliest and
+	// value 0.0. Note that the natural zero value of SamplePair has a timestamp
+	// of 0, which is possible to appear in a real SamplePair and thus not
+	// suitable to signal a non-existing SamplePair.
+	ZeroSamplePair = SamplePair{Timestamp: Earliest}
+)
+
+// A SampleValue is a representation of a value for a given sample at a given
+// time.
+type SampleValue float64
```
codesome, line 34: We could possibly do this and remove duplication methods. (and get the FloatString implementation to this file)

```suggestion
type SampleValue FloatString
```

model/value_histogram.go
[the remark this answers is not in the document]
codesome: No problem, we can start with this for now and do any optimizations later if deemed required.

model/value_histogram.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+	if gotLen := len(tmp); gotLen != wantLen {
+		return fmt.Errorf("wrong number of fields: %d != %d", gotLen, wantLen)
+	}
+	return nil
+}
+
+func (s *HistogramBucket) Equal(o *HistogramBucket) bool {
+	return s == o || (s.Boundaries == o.Boundaries && s.Lower == o.Lower && s.Upper == o.Upper && s.Count == o.Count)
```
codesome, line 85: How does `s == o` help here? Only when both variables are the same with same pointer? If yes, then we can remove it. I doubt if same variables will be checked with each other.

model/value_histogram.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (b HistogramBucket) String() string {
+	var sb strings.Builder
+	lowerInclusive := b.Boundaries == 1 || b.Boundaries == 3
+	upperInclusive := b.Boundaries == 0 || b.Boundaries == 3
```
codesome, line 91: Are you sure you got the conditions right? I would have expected to see `2` somewhere and `3` not repeated.

model/value_histogram.go
```diff
@@ -0,0 +1,175 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+type SampleHistogram struct {
+	Count   FloatString      `json:"count"`
+	Sum     FloatString      `json:"sum"`
+	Buckets HistogramBuckets `json:"buckets"`
```
codesome, line 125: Question to @beorn7: I guess the zero bucket goes inside this as well?

model/value_histogram.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s SampleHistogram) String() string {
+	return fmt.Sprintf("Count: %f, Sum: %f, Buckets: %v", s.Count, s.Sum, s.Buckets)
+}
+
+func (s *SampleHistogram) Equal(o *SampleHistogram) bool {
+	return s == o || (s.Count == o.Count && s.Sum == o.Sum && s.Buckets.Equal(o.Buckets))
```
codesome, line 133: Same comment about requiring `s == o`

model/value_histogram.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return nil
+}
+
+func (s SampleHistogramPair) String() string {
+	return fmt.Sprintf("%s @[%s]", s.Histogram, s.Timestamp)
+}
+
+func (s *SampleHistogramPair) Equal(o *SampleHistogramPair) bool {
+	return s == o || (s.Histogram.Equal(&o.Histogram) && s.Timestamp.Equal(o.Timestamp))
```
codesome, line 170: Same `s == o` comment.

model/value_histogram_test.go
```diff
@@ -0,0 +1,253 @@
...
+package model
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+)
+
+func genSampleHistogram() SampleHistogram {
+	return SampleHistogram{
```
codesome, line 23: It would be good to have >1 bucket when testing.

model/value_histogram_test.go
```diff
@@ -0,0 +1,253 @@
...
+			value: Vector{&Sample{
+				Metric: Metric{
+					MetricNameLabel: "test_metric",
+				},
+				Histogram: genSampleHistogramPtr(),
+				Timestamp: 1234567,
+			}},
+		},
+		{
+			plain: `[{"metric":{"__name__":"test_metric"},"histogram":[1234.567,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}]},{"metric":{"foo":"bar"},"histogram":[1.234,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}]}]`,
```
codesome, line 139: Can we make this string multiline here for readability?

model/value_histogram_test.go
```diff
@@ -0,0 +1,253 @@
...
+	input := []struct {
+		plain string
+		value Matrix
+	}{
+		{
+			plain: `[]`,
+			value: Matrix{},
+		},
+		{
+			plain: `[{"metric":{"__name__":"test_metric"},"histograms":[[1234.567,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}],[12345.678,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}]]},{"metric":{"foo":"bar"},"histograms":[[2234.567,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}],[22345.678,{"count":"1","sum":"4500","buckets":[[0,"4466.7196729968955","4870.992343051145","1"]]}]]}]`,
```
codesome, line 194: Can we make this string multiline here for readability?

codesome (commented):
Nice work! I just have a bunch more nits with some questions for Björn.

model/value.go
beorn7: Yes, I think (and hope) that will work.

beorn7: Or maybe we can just call the current `sampleHistogramPairPtr` `SampleHistogramPair` and get rid of the current `SampleHistogramPair`? Does something not work just because `SampleHistogramPair` will now contain a pointer to a histogram rather than the histogram itself?

model/value.go
beorn7: The HTTP query API is first grouping by type, and then only within each type, it returns samples in the order of timestamps, https://prometheus.io/docs/prometheus/latest/querying/api/#range-vectors

model/value.go
beorn7: I would like to not show empty `histograms: []` arrays because that's how the HTTP API is behaving.

model/value_histogram.go
beorn7: From https://prometheus.io/docs/prometheus/latest/querying/api/#native-histograms :

* 0: “open left” (left boundary is exclusive, right boundary in inclusive)
* 1: “open right” (left boundary is inclusive, right boundary in exclusive)
* 2: “open both” (both boundaries are exclusive)
* 3: “closed both” (both boundaries are inclusive)

Thus, I believe the code is correct as is.

model/value_histogram.go
beorn7: Yes.

model/value_histogram.go
codesome: I should have checked first :) thanks

model/value.go
codesome: Then this comment can be ignored.

model/value_histogram.go
zenador: I don't mind removing all these, but I was just following the original code for consistency, e.g. `SamplePair` was already checking `s == o`. Should they all be removed, including the checks in float sample structs?

model/value_float.go
zenador: I tried to do this, but it didn't remove the need for duplicate methods as we still needed to implement the methods for SampleValue directly. I think this is because for the new histogram structs, we use FloatString as the type directly, but for the old float struct we alias it to SampleValue so it requires its own set of methods. If we wanted to remove the redundancy, we should remove the SampleValue alias and make the Sample use FloatString directly, but this seems like a very big change to existing code, and others importing this library probably already use SampleValue, including Mimir. Alternatively, we could remove FloatString and use SampleValue in its place, but then the naming would be confusing.

model/value.go
zenador: I tried the first suggestion in a reverted commit but it results in a panic when reading JSONs of float samples, but not sure if I was doing it correctly. As for the second suggestion, I think that would work, but not sure if the change to a pointer everywhere (with added complexity) is worth it just for this use case, since besides unmarshalling Samples, we can assume that everytime we use SampleHistogramPair, it should contain an actual histogram and don't need to check for nil pointers. Perhaps there's a better way to write this unmarshalling method for Sample without this `sampleHistogramPairPtr`, just that my knowledge of unmarshalling is currently too basic to figure out how.

model/value.go
zenador: We use it in Mimir to reduce code duplication, since many functions to process SamplePair and SampleHistogramPair rely only on the timestamp and can be combined into one generic function instead of having 2 separate functions with essentially the same code. Unfortunately Golang doesn't allow us to define such generics for imported types, so we can't do this in Mimir. I understand the need to support 1.17 here for now though, so I've reverted it. I was thinking since the next release is likely to be in the next few months and then (maybe) we can drop 1.17, we could keep the generics and the version bump in a separate branch which we could use in Mimir, and merge only after it's safe to drop 1.17. Even after reverting though, the CI test is failing for only 1.17, any idea why?

model/value_histogram.go
codesome: Let's keep it as it is then, ignore my similar comments

model/value.go
codesome: Test failure seems to be flaky. I re-ran it and now it works.

model/value.go
codesome: If it is getting too complicated, we can keep it as it is

codesome (approved):
LGTM. I will let @beorn7 take a final look.

model/value.go
```diff
@@ -38,82 +30,12 @@ var (
...
-// Timestamps. The semantics of Value equality is defined by SampleValue.Equal.
-func (s *SamplePair) Equal(o *SamplePair) bool {
-	return s == o || (s.Value.Equal(o.Value) && s.Timestamp.Equal(o.Timestamp))
-}
-
-func (s SamplePair) String() string {
-	return fmt.Sprintf("%s @[%s]", s.Value, s.Timestamp)
-}
-
 // Sample is a sample pair associated with a metric.
```
beorn7, line 33: Maybe add to the doc comment that either the `Histogram` field or the `Value` field is used. `Histogram == nil` implies that the `Value` field is used.

model/value.go
beorn7: OK, let's keep it as is, but add a TODO to find out if there is a way to get rid of `SampleHistogramPair` (and rename `sampleHistogramPairPtr` to `SampleHistogramPair`).

beorn7 (approved):
Great. Thank you. Just two commenting nits.

beorn7:
I assume if @roidelapluie had issues with this, he would have chimed in by now.

So, merging now… Thank you very much, @zenador .

zenador:
Thank you very much for the reviews, @beorn7 and @codesome !

PR #426, opened 2022-12-17
Clean up config.Header to http.Header conversion

mem:
It was pointed out by @dswarbrick that the current implementation is unnecessary as http.Header has an Add method. Header.Add canonicalizes header names, so the test needs to be adjusted, too.


roidelapluie:
I don't think we have to bother with empty headers. Users should pass an empty string if they want it.

Issue #317, opened 2021-07-15
Consider moving to google.golang.org/protobuf/proto


mem:
github.com/golang/protobuf/proto has been "deprecated" in favor of google.golang.org/protobuf/proto. For this codebase in particular this involves changing calls to `proto.MarshalTextString` to `prototext.Marshal`.

This is related to #281.

conways-glider:
Looks like this dependency now has a security issue: https://nvd.nist.gov/vuln/detail/CVE-2021-3121

beorn7:
My assumption so far is that the move would mean a breaking change for the users of prometheus/client_golang (which exposes a few protobuf specific parts in its interface). If we can pull it off without a breaking change, great. But if it does mean a breaking change, it has to be navigated carefully and in coordination with @kakkoyun and @bwplotka .

bwplotka:
Ack, I think we have to do it at some point. Added ticket on our side: https://github.com/prometheus/client_golang/issues/1175

zhsj:
Continue my comments https://github.com/prometheus/common/pull/427#issuecomment-1362478446 here, since I find it's not easy to get rid of `proto.MarshalTextString`.

proto.MarshalTextString results a string with `<,>` as delim, but `prototext.Marshal` uses `{,}` as delim. The upstream doc says the text format is not stable. Should we rely on this?

Besides, there's no decoder for `FmtProtoText`, is there any user for this format?

beorn7:
I don't think anyone depends on the exact format of `FmtProtoText` (at least I don't think anyone should depend on it).

SuperQ:
I fixed the issue of `proto.MarshalTextString()` by updating prometheus/client_model to `google.golang.org/protobuf/proto` first. 

This is now fixed with #479

PR #427, opened 2022-12-22
Replace deprecated github.com/golang/protobuf package

zhsj:
This replaces usage of proto.{Float64,Int32,Int64,String,Uint32,Uint64}, which doesn't break the interface.

And replaces ptypes.TimestampProto by google.golang.org/protobuf/types/known/timestamppb. This works since github.com/golang/protobuf has been bumped to v1.5.2, which makes ptypes.TimestampProto an alias to timestamppb.Timestamp. This is only used test (openmetrics_create_test.go), which won't break interfaces.

Updates: #317

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/expfmt/decode_test.go
+++ b/expfmt/decode_test.go
@@ -22,7 +22,8 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
+	"google.golang.org/protobuf/proto"
+
 	dto "github.com/prometheus/client_model/go"
 
 	"github.com/prometheus/common/model"
@@ -100,8 +101,7 @@ mf2 4
 }
 
 func TestProtoDecoder(t *testing.T) {
-
-	var testTime = model.Now()
+	testTime := model.Now()
 
 	scenarios := []struct {
 		in       string
@@ -369,7 +369,7 @@ func TestProtoDecoder(t *testing.T) {
 }
 
 func testDiscriminatorHTTPHeader(t testing.TB) {
-	var scenarios = []struct {
+	scenarios := []struct {
 		input  map[string]string
 		output Format
 	}{
@@ -435,7 +435,7 @@ func TestExtractSamples(t *testing.T) {
 			Help: proto.String("Help for foo."),
 			Type: dto.MetricType_COUNTER.Enum(),
 			Metric: []*dto.Metric{
-				&dto.Metric{
+				{
 					Counter: &dto.Counter{
 						Value: proto.Float64(4711),
 					},
@@ -447,7 +447,7 @@ func TestExtractSamples(t *testing.T) {
 			Help: proto.String("Help for bar."),
 			Type: dto.MetricType_GAUGE.Enum(),
 			Metric: []*dto.Metric{
-				&dto.Metric{
+				{
 					Gauge: &dto.Gauge{
 						Value: proto.Float64(3.14),
 					},
@@ -459,7 +459,7 @@ func TestExtractSamples(t *testing.T) {
 			Help: proto.String("Help for bad."),
 			Type: dto.MetricType(42).Enum(),
 			Metric: []*dto.Metric{
-				&dto.Metric{
+				{
 					Gauge: &dto.Gauge{
 						Value: proto.Float64(2.7),
 					},
--- a/expfmt/encode_test.go
+++ b/expfmt/encode_test.go
@@ -18,7 +18,8 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
+	"google.golang.org/protobuf/proto"
+
 	dto "github.com/prometheus/client_model/go"
 )
 
--- a/expfmt/openmetrics_create_test.go
+++ b/expfmt/openmetrics_create_test.go
@@ -20,20 +20,16 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/protobuf/proto"  //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
-	"github.com/golang/protobuf/ptypes" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	dto "github.com/prometheus/client_model/go"
 )
 
 func TestCreateOpenMetrics(t *testing.T) {
-	openMetricsTimestamp, err := ptypes.TimestampProto(time.Unix(12345, 600000000)) //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
+	openMetricsTimestamp := timestamppb.New(time.Unix(12345, 600000000))
 
-	if err != nil {
-		t.Error(err)
-	}
-
-	var scenarios = []struct {
+	scenarios := []struct {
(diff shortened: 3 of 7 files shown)


zhsj:
Some format changes in the PR is caused by gopls format, I can leave them untouched if you don't like that.

And this PR doesn't replace `proto.MarshalTextString`. After https://github.com/golang/protobuf/issues/1442 is fixed, we can only use google.golang.org/protobuf and still consume v1 message from client_model package.

zhsj:
For `proto.MarshalTextString`, I think we can do like this

```
diff --git a/expfmt/encode.go b/expfmt/encode.go
index 64dc0eb..73e84ee 100644
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -18,9 +18,10 @@ import (
        "io"
        "net/http"
 
-       "github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
        "github.com/matttproud/golang_protobuf_extensions/pbutil"
        "github.com/prometheus/common/internal/bitbucket.org/ww/goautoneg"
+       "google.golang.org/protobuf/encoding/prototext"
+       "google.golang.org/protobuf/runtime/protoimpl"
 
 )
@@ -133,7 +134,11 @@ func NewEncoder(w io.Writer, format Format) Encoder {
        case FmtProtoText:
                return encoderCloser{
-                               _, err := fmt.Fprintln(w, proto.MarshalTextString(v))
+                               m, err := prototext.Marshal(protoimpl.X.ProtoMessageV2Of(v))
+                               if err != nil {
+                                       return err
+                               }
+                               _, err = w.Write(m)
                                return err
                        },
                        close: func() error { return nil },

```

This uses `google.golang.org/protobuf/runtime/protoimpl`. Regarding https://github.com/golang/protobuf/issues/1442, the grpc people feel it's a temporary solution to use that package. But well I think it's already in the public exported place..

roidelapluie:
CircleCI Pipeline — Could not find a usable config.yml, you may have revoked the CircleCI OAuth app.

zhsj:
> CircleCI Pipeline — Could not find a usable config.yml, you may have revoked the CircleCI OAuth app.

It's now fixed. I logged in CircleCI again. Probably I've revoked it before.

SuperQ:
This needs a rebase to pick up go.mod conflicts.

zhsj:
rebased.

SuperQ:
Sorry, needs a rebase again. :joy:

expfmt/decode_test.go
```diff
@@ -22,10 +22,10 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
-	dto "github.com/prometheus/client_model/go"
-
 	"github.com/prometheus/common/model"
+
+	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
```
beorn7, line 28: Nit: In the Prometheus code base, we tend to keep imports from other repos than the current one in its own block _above_ the imports from the same repo (as it was before). And we usually also have the named imports in a different block (which wasn't adhered to before, but you could fix it now that we are on it).

expfmt/openmetrics_create_test.go
```diff
@@ -20,20 +20,15 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/protobuf/proto"  //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
-	"github.com/golang/protobuf/ptypes" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
-
 	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
```
beorn7, line 25: Import nit as before: Have the named imports in their own block after the unnamed imports (as it was before).

expfmt/encode_test.go
```diff
@@ -18,8 +18,8 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
 	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
```
beorn7, line 22: You could fix the block structure here (it wasn't quite right before, see other nitpicking comments).

expfmt/text_parse.go
```diff
@@ -22,10 +22,10 @@ import (
 	"strconv"
 	"strings"
 
-	dto "github.com/prometheus/client_model/go"
-
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
 	"github.com/prometheus/common/model"
+
+	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
```
beorn7, line 28: Same nitpicking as before. This has to be

```
"google.golang.org/protobuf/proto"

```

And yes, it was wrong in a different way before...

expfmt/text_parse_test.go
```diff
@@ -19,12 +19,12 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/golang/protobuf/proto" //nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
 	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
```
beorn7, line 23: Import nitpick as above.

beorn7 (commented):
(Individual comments are just nitpicking.)

If I remember correctly, the problem is that consumers of this package have to change there import paths in the same way at the same time. Even though the interface doesn't change, a different import path still means we have different types. This could even lead to really tricky situations if someone has multiple dependencies, and some are still using the old version of prometheus/common while others use the new. That's usually caught by changing the import path with new major versions. But prometheus/common is still 0.x. However, prometheus/client_golang is already at 1.x, so there is an expectation to not do breaking changes.

I think the call if we can make this change or not should be made by @kakkoyun and @bwplotka . Given how frequently prometheus/client_golang is used, that's not an easy decision.

beorn7:
Hmmm, thinking more about it, the only usage of the proto package outside of tests is in text_parse.go line 288, and this particular function just returns a pointer to a (standard) `string`. Which means that the change in this PR should actually be transparent, and importers of this package should be able to continue using the proto package of their choice. (I was already wondering why this even works without changing prometheus/client_model…)

So I would cautiously say now that this PR is OK (modulo import nitpicking ;). Still, would be good to have @kakkoyun / @bwplotka confirm.

zhsj:
> If I remember correctly, the problem is that consumers of this package have to change their import paths in the same way at the same time. Even though the interface doesn't change, a different import path still means we have different types.

The break already happened when bumping google/protobuf from 1.3 to 1.4, which forces using google.golang.org/protobuf.

As you can see the go.mod changes, it only moves google.golang.org/protobuf from indirect to direct.

expfmt/decode_test.go
zhsj: format is fixed.

zhsj:
> The break already happened when bumping google/protobuf from 1.3 to 1.4, which forces using google.golang.org/protobuf.

The above is not related to this repo. But for other repos that have pb.go files.

This repo as @beorn7 said, "returns a pointer to a (standard) string".

And the string format should be its interface. That's why `proto.MarshalTextString` is still using the old golang/protobuf package, see https://github.com/prometheus/common/issues/317#issuecomment-1362608722.

SuperQ:
Fixed in #479

PR #429, opened 2022-12-22
Fix marshalling of TLSVersion

roidelapluie:

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -80,12 +80,9 @@ func (tv *TLSVersion) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return fmt.Errorf("unknown TLS version: %s", s)
 }
 
-func (tv *TLSVersion) MarshalYAML() (interface{}, error) {
-	if tv == nil || *tv == 0 {
-		return []byte("null"), nil
-	}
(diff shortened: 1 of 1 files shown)


roidelapluie:
This change updates the MarshalYAML and MarshalJSON functions for the TLSVersion type to remove the pointer notation *tv. This means that the functions will now take a value of type TLSVersion instead of a pointer to a TLSVersion as an argument.

The reason for this change is because the functions previously did not implement the json.Marshaler or yaml.Marshaler interfaces, which expect a value to be passed as the argument rather than a pointer. By removing the pointer notation, the functions should now work correctly when passed a value of type TLSVersion.

It's worth noting that the UnmarshalYAML and UnmarshalJSON functions for TLSVersion are not changed, and continue to accept a pointer as the argument. This is because these functions implement the yaml.Unmarshaler and json.Unmarshaler interfaces, respectively, which expect a pointer to be passed as the argument.

*Note: generated with the help of chatgpt*

LeviHarrison:
Looks good, but do you mind adding a test to be safe?

LeviHarrison:
> Note: generated with the help of chatgpt

🤣

roidelapluie:
I will add tests later.

PR #433, opened 2023-01-10
Add dependabot

lucacome:
Adds dependabot to automate GitHub workflows and go dependencies updates.


lucacome:
@roidelapluie I can also help merge these dependencies PRs once this is merged 🙂

bboreham:

lucacome:
@bboreham I'm not sure what you mean. This library uses dependencies that need to be kept updated to get bug fixes and more importantly CVEs fixes.

bboreham:
I mean this has no practical benefit. 
Only the dependencies built into _programs_ (executables) matter. Changing this library does not affect any programs, unless they separately decide to update. So let’s let them do that and avoid the clutter here.

SuperQ:
@bboreham In the Go ecosystem, it does matter, as downstream Go mod takes module versions included here as part of the update tree.

We already have dependabot on prometheus/client_golang.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
```
SuperQ, line 6: While I like the idea of doing github-actions updates, I think our plan is to mostly control this from `prometheus/prometheus` and use our repo sync updates to manage things.

```suggestion
```

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: daily
```
SuperQ, line 10: We only do updates monthly to avoid too much PR noise.

```suggestion
      interval: monthly
```

SuperQ (changes requested):

mmorel-35:
Don’t forget sigv4 submodule

bboreham:
Downstream only takes an update if they specifically decide to update this library. It’s either a no-op or an accident, and I don’t want to promote either. 

I’m fine with removing dependabot from all Prometheus libraries.

SuperQ:
Except that's not how Go mod works in practice for indirect dependencies. It's not a noop or an accident, it's a core of how Go modules work. It doesn't matter if it's a library or an end-user binary repo. We need to maintain the versions we depend upon here. Automating it is necessary for maintainer sanity.

bboreham:
You are suggesting dependabot doesn’t work for indirect dependencies?

SuperQ:
Nope, dependabot will not update indirect unless it's for a security vulnerability.

SuperQ (approved):

kakkoyun (approved):

PR #436, opened 2023-01-20
ci(deps): setup dependabot 

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


bboreham:

lucacome:
This doesn't close #433 but some parts of this are a duplicate of it

mmorel-35:
Hi @lucacome ,
It's now aligned 🙂!

SuperQ:
Closing in favor of #433

PR #439, opened 2023-01-26
Make it more efficient and compatible to use SampleHistogram

krajorama:
Store SampleHistogram in SampleHistogramPair as pointer.
More compatible with what protobuf generates.
More efficient when taking the histogram from the pair.
Simpler test code.

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>

### DIFF
--- a/model/value_histogram.go
+++ b/model/value_histogram.go
@@ -135,6 +135,7 @@ func (s *SampleHistogram) Equal(o *SampleHistogram) bool {
 
 type SampleHistogramPair struct {
 	Timestamp Time
+	// Histogram should never bill nil, it's only stored as pointer for efficiency
 	Histogram *SampleHistogram
 }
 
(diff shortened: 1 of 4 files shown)


model/value_histogram.go
```diff
@@ -135,7 +135,7 @@ func (s *SampleHistogram) Equal(o *SampleHistogram) bool {
 
 type SampleHistogramPair struct {
 	Timestamp Time
-	Histogram SampleHistogram
+	Histogram *SampleHistogram
```
codesome, line 138: Worth commenting that it should never be `nil` and it is as a pointer for efficiency.

krajorama: should MarshalJSON check that it's nil and error out if it is?

krajorama: I've ended up convincing myself to error on `nil` histogram, to avoid ever generating invalid JSON from this

PR #443, opened 2023-02-09
expfmt: only call TextParser.TextToMetricFamilies once from textDecoder

### LINKED ISSUE #442: expfmt: TextDecoder infinite bufio Read recursion
## Problem

If the `io.Reader` passed to `expfmt.NewDecoder` with the  `expfmt.FmtText` format is a `*bufio.Reader`, then `Decoder.Decode` will infinitely recurse in the `bufio.(*Reader).Read` function.

## Explanation

[expfmt.(*textDecoder).Decode](https://github.com/prometheus/common/blob/v0.39.0/expfmt/decode.go#L123-L144) calls [expfmt.(*TextParser).TextToMetricFamilies](https://github.com/prometheus/common/blob/v0.39.0/expfmt/text_parse.go#L100-L119) passing its same `io.Reader` each time, which is then passed to [expfmt.(*TextParser).reset](https://github.com/prometheus/common/blob/v0.39.0/expfmt/text_parse.go#L121-L138), which contains the following snippet:

```go
	if p.buf == nil {
		p.buf = bufio.NewReader(in)
	} else {
		p.buf.Reset(in)
	}
```

The problem is that `bufio.NewReader` will return the given `io.Reader` without wrapping it, if it happens to already be a `*bufio.Reader` of sufficient size.  All good on the first call from `Decode`, but the second call ends up setting the `*bufio.Reader` as its own underlying `io.Reader` and we get an infinite recursion.

A reasonable person _might_ argue that this is a bug in `bufio`, but we can still fix it in `expfmt`.

## Repro

```go
package main

import (
	"bufio"
	"fmt"
	"io"
	"os"
	"strings"

	"github.com/prometheus/common/expfmt"
)

const example = `
# TYPE foo gauge
foo 0
`

func main() {
	r := bufio.NewReader(strings.NewReader(example))
	dec := expfmt.NewDecoder(r, expfmt.FmtText)
	for {
		var mf dto.MetricFamily
		if err := dec.Decode(&mf); err != nil {
			if err == io.EOF {
				return
			}
			fmt.Fprintln(os.Stderr, err)
			os.Exit(1)
		}
	}
}
```

abursavich:
Fixes #442

It turns out that TextParser.TextToMetricFamilies reads all metrics the first time and only returns partial results if there's an error. The bug was triggered by calling it twice. We only need to call it once.

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -115,32 +115,28 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 // textDecoder implements the Decoder interface for the text protocol.
 type textDecoder struct {
 	r    io.Reader
-	p    TextParser
-	fams []*dto.MetricFamily
+	fams map[string]*dto.MetricFamily
+	err  error
 }
 
 // Decode implements the Decoder interface.
 func (d *textDecoder) Decode(v *dto.MetricFamily) error {
-	// TODO(fabxc): Wrap this as a line reader to make streaming safer.
-	if len(d.fams) == 0 {
-		// No cached metric families, read everything and parse metrics.
-		fams, err := d.p.TextToMetricFamilies(d.r)
-		if err != nil {
-			return err
-		}
-		if len(fams) == 0 {
-			return io.EOF
-		}
-		d.fams = make([]*dto.MetricFamily, 0, len(fams))
-		for _, f := range fams {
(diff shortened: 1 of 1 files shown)


beorn7 (approved):
Thank you. I think this should work.

@roidelapluie I lease final approval/merge to you.

roidelapluie (approved):

PR #444, opened 2023-02-09
expfmt: only ignore io.EOF errors in TextParse.startOfLine

abursavich:
We shouldn't ignore all errors here, only `io.EOF`. If an unexpected connection break just happens to align with the start of a new line, that error shouldn't be suppressed.


beorn7 (changes requested):
Thank you very much, makes total sense. But could you change the code comment to reflect your reasoning?

abursavich:
Updated comment. Please take another look.

beorn7 (approved):
Thanks again.

Over to @roidelapluie for final approval and merging.

PR #446, opened 2023-02-18
Enable dependabot

SuperQ:
Add dependabot for monthly updates.


roidelapluie:
all broken?

mmorel-35:
Duplicates #436

bboreham:

lucacome:
well this one and #436 are actually a duplicate of #433

SuperQ:
Closing in favor of #433

PR #447, opened 2023-02-20
Update Go versions

SuperQ:
* Add Go 1.20.
* Set minimum Go version to 1.18.
* Bump golang.org/x/net.


roidelapluie:
client_golang depends on ; have they / will they also bump the go version requirement?

SuperQ:
Added client_golang maintainers to this.

beorn7:
IIRC we are usually trying to support the last three Go minor releases. From that perspective, Go1.19 would be fine once Go1.21 is out. But I might misremember, and I'm pretty sure it was always a best-effort thing. (Although it's usually easy with build tags etc. to maintain backwards compatibility even if using more recent Go features.)

SuperQ:
@beorn7 We used to do that, but recently we have been only trying to support only what Go officially supports.

beorn7:
I guess @bwplotka and @kakkoyun should confirm or deny…

SuperQ:
Ok, I added Go 1.18 back in.

I still want to get sign-off from both client_golang and prometheus maintainers.

kakkoyun:
I won't have access to a computer until Sunday. If it's urgent, leaving the decision to @bwplotka

bwplotka:
LGTM, thanks for keeping 1.18 in, not big harm here to keep it IMO.

bwplotka (approved):

roidelapluie:
Thanks!

kakkoyun (commented):
LGTM

PR #448, opened 2023-02-20
expfmt: add test case for issue with *bufio.Reader and FmtText format

roidelapluie:
The test case checks that a *bufio.Reader can be passed to expfmt.NewDecoder with the FmtText format without causing an infinite recursion. This is a regression test for the issue reported in https://github.com/prometheus/common/issues/442.

### DIFF
--- a/expfmt/decode_test.go
+++ b/expfmt/decode_test.go
@@ -14,6 +14,7 @@
 package expfmt
 
 import (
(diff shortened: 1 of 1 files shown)


beorn7 (approved):

abursavich (commented):
LGTM

PR #449, opened 2023-02-20
expfmt: add test cases for TextParser startOfLine error handling

roidelapluie:
The test cases cover the various scenarios that can occur when the TextParser `startOfLine` method encounters an error. The first subtest ensures that the method handles the `io.EOF` error correctly when the input `io.Reader` is empty.

The second subtest uses a custom implementation of `io.Reader` to test that `startOfLine` handles errors other than `io.EOF` correctly.

See #444

### DIFF
--- a/expfmt/text_parse_test.go
+++ b/expfmt/text_parse_test.go
@@ -14,6 +14,7 @@
 package expfmt
 
 import (
+	"errors"
 	"math"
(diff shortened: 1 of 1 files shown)


beorn7 (approved):

abursavich (commented):
LGTM

PR #455, opened 2023-02-24
version: add Go build tags to output

bboreham:
Build tags allow different implementations of code sections to be included in the binary, so can be important to understand the exact behaviour expected at runtime.

This change includes build tags at the end of the `BuildContext` string, and as an extra label on the info metric.

Example output when linked into Prometheus:
```
ts=2023-02-24T11:33:59.355Z caller=main.go:561 level=info build_context="(go=go1.19.3, platform=linux/amd64, user=vagrant@vagrant, date=20230224-11:12:08, tags=netgo,builtinassets,stringlabels)"
```

### DIFF
--- a/version/info_go118.go
+++ b/version/info_go118.go
@@ -19,6 +19,7 @@ package version
 import "runtime/debug"
 
 var computedRevision string
+var computedTags string
 
 func getRevision() string {
 	if Revision != "" {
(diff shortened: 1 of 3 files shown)


version/info_go118.go
```diff
@@ -27,19 +28,24 @@ func getRevision() string {
...
 func init() {
-	computedRevision = computeRevision()
+	computedRevision, computedTags = computeRevision()
 }
 
-func computeRevision() string {
+func computeRevision() (string, string) {
 	var (
 		rev      = "unknown"
+		tags     = "unknown"
```
SuperQ, line 42: Wouldn't this be `none` if we're building with no tags?

roidelapluie: I am fine with unknown, maybe this can be removed by the compiler? let's proceed as is.

PR #458, opened 2023-02-25
Update kingpin module source

hairyhenderson:
The `gopkg.in/alecthomas/kingpin.v2` repo is now located at `github.com/alecthomas/kingpin/v2` as of v2.3.1. I'm bumping this so that updates can continue to be detected.


SuperQ (approved):
Oh, great, thanks! We'll need to update this in a bunch of places.

PR #461, opened 2023-03-06
Parse Durations much faster

bboreham:
Without using a regexp.

This shows up in profiles of Prometheus service discovery, since `ParseDuration` is called twice per discovered target.

```
name             old time/op    new time/op    delta
ParseDuration-4    2.54µs ±28%    0.04µs ± 3%   -98.30%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
ParseDuration-4      480B ± 0%        0B       -100.00%  (p=0.079 n=4+5)

name             old allocs/op  new allocs/op  delta
ParseDuration-4      2.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
```

Code inspired by Go standard library `time.ParseDuration`, but simplified and supporting different units.

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thanks!

bboreham:
Now I realise this makes one of the Prometheus tests fail due to changed error message:

```
--- FAIL: TestPopulateLabels (0.00s)
    manager_test.go:436: 
                Error Trace:    /home/vagrant/src/github.com/prometheus/prometheus/scrape/manager_test.go:436
                Error:          Error message not equal:
                                expected: "error parsing scrape interval: not a valid duration string: \"2notseconds\""
                                actual  : "error parsing scrape interval: unknown unit \"notseconds\" in duration \"2notseconds\""
                Test:           TestPopulateLabels
```

Do you think it's important to retain consistency of error messages?  I.e. shall I fix it here or in the test?

soundvibe:
This also fails to parse `"0"` durations, breaking a lot of code. Would be great if such improvements would be compatible with previous versions.

PR #462, opened 2023-03-06
Add support to use Proxy From Environment

roidelapluie:
This commit adds a new field `ProxyFromEnvironment` to the `HTTPClientConfig` and `OAuth2` structs to enable Prometheus to use the default proxy from the environment variable (as specified in net/http ProxyFromEnvironment.) This is opt-out by default, so does not change anything for existing users.

This change is necessary because Prometheus previously had insufficient behavior to avoid proxying EC2 metadata requests when `proxy_url` is set in `ec2_sd_configs`.

See https://github.com/prometheus/prometheus/issues/11991


config/http_config.go
```diff
@@ -1072,3 +1084,75 @@ func (c HTTPClientConfig) String() string {
...
+	// value type instead of string.
+	ProxyConnectHeader Header `yaml:"proxy_connect_header,omitempty" json:"proxy_connect_header,omitempty"`
+
+	proxyFunc func(*http.Request) (*url.URL, error)
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (c *ProxyConfig) Validate() error {
+	if len(c.ProxyConnectHeader) > 0 && (!c.ProxyFromEnvironment && (c.ProxyURL.URL == nil || c.ProxyURL.String() == "")) {
+		return fmt.Errorf("if proxy_connect_header is configured proxy_url or proxy_from_environment must also be configured")
```
LeviHarrison, line 1108: small readability nit (applies to other messages too)

```suggestion
		return fmt.Errorf("if proxy_connect_header is configured, proxy_url or proxy_from_environment must also be configured")
```

config/http_config.go
```diff
@@ -1072,3 +1084,75 @@ func (c HTTPClientConfig) String() string {
...
+	}
+	if c.ProxyFromEnvironment && c.ProxyURL.URL != nil && c.ProxyURL.String() != "" {
+		return fmt.Errorf("if proxy_from_environment is configured proxy_url must not be configured")
+	}
+	if c.ProxyFromEnvironment && c.NoProxy != "" {
+		return fmt.Errorf("if proxy_from_environment is configured no_proxy must not be configured")
+	}
```
LeviHarrison, line 1115: Following the logic here, might it also make sense to have a check for if `no_proxy` is configured but not `proxy_url`? Although it technically wouldn't be an issue, it's like having `proxy_connect_header` but no any proxy url's, which is checked for.

config/http_config.go
```diff
@@ -724,7 +735,8 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		tlsTransport := func(tlsConfig *tls.Config) (http.RoundTripper, error) {
 			return &http.Transport{
 				TLSClientConfig:       tlsConfig,
-				Proxy:                 http.ProxyURL(rt.config.ProxyURL.URL),
+				Proxy:                 rt.config.ProxyConfig.Proxy(),
+				ProxyConnectHeader:    rt.config.ProxyConfig.GetProxyConnectHeader(),
```
LeviHarrison, line 739: So was `ProxyConnectHeader` supposed to be there before but was left out?

config/http_config.go
```diff
@@ -1072,3 +1084,75 @@ func (c HTTPClientConfig) String() string {
...
+		proxy := &httpproxy.Config{
+			HTTPProxy:  c.ProxyURL.String(),
+			HTTPSProxy: c.ProxyURL.String(),
+			NoProxy:    c.NoProxy,
+		}
+		proxyFn := proxy.ProxyFunc()
+		c.proxyFunc = func(req *http.Request) (*url.URL, error) {
+			return proxyFn(req.URL)
+		}
+	}
```
LeviHarrison, line 1151: Is there a reason to do this here and not put it in a function called from `NewRoundTripperFromConfig()` or something like that, because it might be simpler.

roidelapluie: Yes Proxy is a function and need to be set to the httpTransport directly

config/http_config.go
roidelapluie: Yes

SuperQ (approved):
LGTM

PR #475, opened 2023-04-27
Negotiate OM v1.0.0

### LINKED ISSUE #456: Content negotiation is broken for expfmt and for promhttp
According to [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18913#issuecomment-1444579266) on OTel's Prometheus exporter `expfmt` and `promhttp` are returning the wrong format if `openmetrics-text 1.0.0` is requested. It seems `openmetrics-text 1.0.0` is not supported only `openmetrics-text 0.0.1` is.

If a user requests `openmetrics-text 1.0.0`:
```
Accept: application/openmetrics-text; version=1.0.0; charset=utf-8
```
The expected result should be in the OpenMetrics format, also the content-type should be:
```
Content-Type: application/openmetrics-text; version=1.0.0; charset=utf-8
```
But the result is in Prometheus plaintext format, the content-type says the same:
```
Content-Type: text/plain; version=0.0.4; charset=utf-8
```
If a user requests `openmetrics-text 0.0.1`:
```
Accept: application/openmetrics-text; version=0.0.1; charset=utf-8
```
The result is in the expected OpenMetrics format:
```
Content-Type: application/openmetrics-text; version=0.0.1; charset=utf-8
```

As far as I know, `0.0.1` should not be used, only `1.0.0`.
According to the [OpenMetrics specs](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure):
>The content type MUST be: application/openmetrics-text; version=1.0.0; charset=utf-8

Also the Prometheus server uses 1.0.0 for scraping: https://github.com/prometheus/prometheus/issues/9430

gouthamve:
Also return OM v1.0.0, and not v0.0.1

Fixes #456

### DIFF
--- a/expfmt/encode_test.go
+++ b/expfmt/encode_test.go
@@ -63,6 +63,45 @@ func TestNegotiate(t *testing.T) {
 	}
 }
 
+func TestNegotiateOpenMetrics(t *testing.T) {
+	tests := []struct {
+		name              string
+		acceptHeaderValue string
+		expectedFmt       string
+	}{
+		{
+			name:              "OM format, no version",
+			acceptHeaderValue: "application/openmetrics-text",
+			expectedFmt:       string(FmtOpenMetrics_1_0_0),
+		},
+		{
+			name:              "OM format, 0.0.1 version",
+			acceptHeaderValue: "application/openmetrics-text;version=0.0.1",
+			expectedFmt:       string(FmtOpenMetrics_0_0_1),
+		},
+		{
+			name:              "OM format, 1.0.0 version",
+			acceptHeaderValue: "application/openmetrics-text;version=1.0.0",
+			expectedFmt:       string(FmtOpenMetrics_1_0_0),
+		},
+		{
+			name:              "OM format, invalid version",
+			acceptHeaderValue: "application/openmetrics-text;version=0.0.4",
+			expectedFmt:       string(FmtText),
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			h := http.Header{}
+			h.Add(hdrAccept, test.acceptHeaderValue)
+			actualFmt := string(NegotiateIncludingOpenMetrics(h))
+			if actualFmt != test.expectedFmt {
+				t.Errorf("expected Negotiate to return format %s, but got %s instead", test.expectedFmt, actualFmt)
+			}
+		})
+	}
+}
 func TestEncode(t *testing.T) {
 	var buff bytes.Buffer
 	delimEncoder := NewEncoder(&buff, FmtProtoDelim)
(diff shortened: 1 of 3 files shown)


jesusvazquez (approved):
Overall LGTM. Lets merge to solve the fix and see if this can get in the next release.

roidelapluie:
It feels like we should return 0.0.1 if no version is specified to be backwards compatible

gouthamve:
It kinda goes against the spec as mentioned here: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure

expfmt/encode_test.go
```diff
@@ -63,6 +63,45 @@ func TestNegotiate(t *testing.T) {
...
+		},
+		{
+			name:              "OM format, 1.0.0 version",
+			acceptHeaderValue: "application/openmetrics-text;version=1.0.0",
+			expectedFmt:       string(FmtOpenMetrics_1_0_0),
+		},
+		{
+			name:              "OM format, invalid version",
+			acceptHeaderValue: "application/openmetrics-text;version=0.0.4",
+			expectedFmt:       string(FmtText),
```
bboreham, line 90: Why?  Shouldn't it be an error?

beorn7: Historically, the fall-back was always the (classic) text format.

beorn7:
If the content type really MUST be `application/openmetrics-text; version=1.0.0; charset=utf-8`, what does that mean for the content negotiation? Wouldn't it already be against the spec if the `Accept` header contains any content type different from the above? And if so, then we surely can respond to a non-spec-conforming `Accept` header with non-spec-conforming content type header, can't we?

Or less formally argued: If I announce that I accept OM of _any_ version, then I cannot expect to get 1.0.0 back. If I want 1.0.0, then I have to put it into the accept header in exactly that way.

expfmt/encode_test.go
jonatan-ivanov: Also, this is the current behavior if you request a version of OM that is not supported (e.g.: 1.0.0).

jonatan-ivanov (approved):

gouthamve:
@roidelapluie @beorn7 Now returning 0.0.1 with no version for backwards compat.

beorn7 (approved):

beorn7:
@jonatan-ivanov I actually tried to argue that returning v0.0.1 when no version is requested is OK. :)

Note that whatever content-type we return, the actual content is the same. Plus, the Go implementation of OpenMetrics isn't even complete yet. And second-order plus, there is a discussion if we should go for v1.1 or even v2.0 instead of trying to fully implement v1.0. So the discussion is a bit academic at this point.

I think with this PR, we should fix the current behavior of not even returning v1.0.0 even when requested, but we should try to change as little as possible to minimize breakages at this point.

@jonatan-ivanov and @bboreham I see your points, and this should be discussed. But by not changing the behavior now, we still keep the door open for changes in the future. While changing more than needed right now might create breakage noise that we might not need to create, e.g. if "the real thing" will be OM v1.1 or v2.0 anyway.

PR #487, opened 2023-05-15
config: allow exposing real secret value through marshal

GiedriusS:
There are external projects out there that marshal/unmarshal the Prometheus config that has secrets. To get the real value currently such horrible workarounds are in place:
https://github.com/tkestack/kvass/blob/master/pkg/sidecar/injector.go#L175-L214

Let's add a way to get the original values through the `Secret` type to avoid such things.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -28,9 +28,11 @@ func TestJSONMarshalSecret(t *testing.T) {
 		S Secret
 	}
 	for _, tc := range []struct {
-		desc     string
-		data     tmp
-		expected string
+		desc      string
+		data      tmp
+		expected  string
+		trueValue bool
+		testYAML  bool
 	}{
 		{
 			desc: "inhabited",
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -28,9 +28,11 @@ func TestJSONMarshalSecret(t *testing.T) {
 		S Secret
 	}
 	for _, tc := range []struct {
-		desc     string
-		data     tmp
-		expected string
+		desc      string
+		data      tmp
+		expected  string
+		trueValue bool
```
SuperQ, line 34: "trueValue" isn't a very descriptive name. How about:

```suggestion
		marshalSecret bool
```

SuperQ:
Good idea, minor nit about the test setup.

alita1991:
Is there any plan to merge the code in the next period of time?

roidelapluie:
thanks, I think we can do the test improvement later on.

PR #490, opened 2023-05-22
Add read-only token permissions

pnacht:
Fixes https://github.com/prometheus/prometheus/issues/12379.

As mentioned there, this PR ensures the golangci-lint workflow always runs with read-only permissions, protecting the projects that use it from supply-chain attacks.

This PR was originally submitted as https://github.com/prometheus/procfs/pull/525, but following @discordianfish's suggestion there, I'm re-submitting it here.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -10,6 +10,9 @@ on:
       - ".golangci.yml"
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   golangci:
     name: lint


discordianfish (approved):
LGTM, thanks!

pnacht:
Hey, are there any changes you'd like to see on this policy?

discordianfish:
Technically @roidelapluie needs to approve this first

roidelapluie:
LGTM

pnacht:
Thanks for merging this! However, I just noticed that `golangci-lint.yml` exists both here in prometheus/common and in [prometheus/prometheus/scripts](https://github.com/prometheus/prometheus/tree/main/scripts).

Also, the prometheus/prometheus version seems to be more up-to-date than this one (it runs Go 1.54.2, while here it's on 1.51.2, for example).

I also noticed that prometheus/procfs just received an update to its `golangci-lint.yml` taken from prometheus/prometheus. Should I repeat this PR over there?

discordianfish:
Uhmm.. good question, I assumed this is the source of truth but might be out of the loop a bit - @roidelapluie?

PR #494, opened 2023-06-16
Add Sigv4 Endpoint config parameter

kanwren:
Exposes an `endpoint` field in the sigv4 config, so that the call to STS when assuming a role with a web identity can use a custom STS endpoint, e.g. a VPC endpoint.

No clients other than the STS one are made from the session config here, so setting `Endpoint` should be safe, but it can also be a resolver if this is a concern.

Addresses https://github.com/prometheus/prometheus/issues/11710.

### DIFF
--- a/sigv4/sigv4.go
+++ b/sigv4/sigv4.go
@@ -62,6 +62,7 @@ func NewSigV4RoundTripper(cfg *SigV4Config, next http.RoundTripper) (http.RoundT
 	sess, err := session.NewSessionWithOptions(session.Options{
 		Config: aws.Config{
 			Region:      aws.String(cfg.Region),
+			Endpoint:    aws.String(cfg.Endpoint),
 			Credentials: creds,
 		},
 		Profile: cfg.Profile,
--- a/sigv4/sigv4_config.go
+++ b/sigv4/sigv4_config.go
@@ -28,6 +28,7 @@ type SigV4Config struct {
 	SecretKey config.Secret `yaml:"secret_key,omitempty"`
 	Profile   string        `yaml:"profile,omitempty"`
 	RoleARN   string        `yaml:"role_arn,omitempty"`
+	Endpoint  string        `yaml:"endpoint,omitempty"`
 }
 
 func (c *SigV4Config) Validate() error {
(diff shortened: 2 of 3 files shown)


qswinson:
Any chance on getting this merged? It would be very helpful for AWS VPC deployments that need to keep all network traffic within AWS.

rapphil:
Thanks for creating this PR.

sigv4/sigv4.go
```diff
@@ -62,6 +62,7 @@ func NewSigV4RoundTripper(cfg *SigV4Config, next http.RoundTripper) (http.RoundT
 	sess, err := session.NewSessionWithOptions(session.Options{
 		Config: aws.Config{
 			Region:      aws.String(cfg.Region),
+			Endpoint:    aws.String(cfg.Endpoint),
```
rapphil, line 65: I think what you want to change is the property `STSRegionalEndpoint` and set it to `RegionalSTSEndpoint`

References:
https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config
https://github.com/aws/aws-sdk-go/blob/main/aws/endpoints/endpoints.go#L186C2-L186C21

Then we don't need to add a configuration because it is fine to always use the regional STS endpoint.

rapphil: On the other hand, I think it is safer to add a configuration parameter whose default value is the current behaviour (global endpoint - value `LegacySTSEndpoint`). In fact you the zero value will be the legacy endpoint so it is fine to not set if the configuration is not setting.

 https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#id_credentials_region-endpoints

rapphil: still we have the benefit of not needing to provide the sts endpoint explicitly and we leave the SDK to do that for us.

alexgreenbank:
Hello from the bug scrub.

There is some feedback on this PR that needs attention. Is there any further work required?

SuperQ:

sigv4/sigv4.go
kanwren: (Responding even though this PR is closed and will be moved)

Note that `STSRegionalEndpoint` is unrelated to the problem at hand here, which is the lack of ability to use a VPC endpoint for STS when one needs to avoid public-facing endpoints altogether.

PR #504, opened 2023-07-17
Write created lines when negotiating OpenMetrics

ArthurSens:
Following up on the [Design doc](https://docs.google.com/document/d/1kakDVn8aP1JerimLeazuMfy2jaB14W2kZDHMRokPcv4/edit), this PR aims to implement writing created timestamps when using OpenMetrics.

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -139,7 +139,13 @@ func NegotiateIncludingOpenMetrics(h http.Header) Format {
 // interface is kept for backwards compatibility.
 // In cases where the Format does not allow for UTF-8 names, the global
 // NameEscapingScheme will be applied.
-func NewEncoder(w io.Writer, format Format) Encoder {
+//
+// NewEncoder can be called with additional options to customize the OpenMetrics text output.
+// For example:
+// NewEncoder(w, FmtOpenMetrics_1_0_0, WithCreatedLines())
+//
+// Extra options are ignored for all other formats.
+func NewEncoder(w io.Writer, format Format, options ...EncoderOption) Encoder {
 	escapingScheme := format.ToEscapingScheme()
 
 	switch format.FormatType() {
@@ -178,7 +184,7 @@ func NewEncoder(w io.Writer, format Format) Encoder {
 	case TypeOpenMetrics:
 		return encoderCloser{
 			encode: func(v *dto.MetricFamily) error {
-				_, err := MetricFamilyToOpenMetrics(w, model.EscapeMetricFamily(v, escapingScheme))
+				_, err := MetricFamilyToOpenMetrics(w, model.EscapeMetricFamily(v, escapingScheme), options...)
 				return err
 			},
 			close: func() error {
--- a/expfmt/openmetrics_create.go
+++ b/expfmt/openmetrics_create.go
@@ -28,14 +28,24 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
-type toOpenMetrics struct {
+type encoderOption struct {
 	withCreatedLines bool
 }
 
-type ToOpenMetricsOption func(*toOpenMetrics)
+type EncoderOption func(*encoderOption)
 
-func WithCreatedLines() ToOpenMetricsOption {
-	return func(t *toOpenMetrics) {
+// WithCreatedLines is an EncoderOption that configures the OpenMetrics encoder
+// to include _created lines (See
+// https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter-1).
+// Created timestamps can improve the accuracy of series reset detection, but
+// come with a bandwidth cost.
+//
+// At the time of writing, created timestamp ingestion is still experimental in
+// Prometheus and need to be enabled with the feature-flag
+// `--feature-flag=created-timestamp-zero-ingestion`, and breaking changes are
+// still possible. Therefore, it is recommended to use this feature with caution.
+func WithCreatedLines() EncoderOption {
+	return func(t *encoderOption) {
 		t.withCreatedLines = true
 	}
 }
@@ -85,8 +95,8 @@ func WithCreatedLines() ToOpenMetricsOption {
 //
 //   - The value of Counters is not checked. (OpenMetrics doesn't allow counters
 //     with a `NaN` value.)
-func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...ToOpenMetricsOption) (written int, err error) {
-	toOM := toOpenMetrics{}
+func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...EncoderOption) (written int, err error) {
+	toOM := encoderOption{}
 	for _, option := range options {
 		option(&toOM)
 	}
--- a/expfmt/openmetrics_create.go
+++ b/expfmt/openmetrics_create.go
@@ -23,9 +23,22 @@ import (
 	"strings"
 
 	"github.com/prometheus/common/model"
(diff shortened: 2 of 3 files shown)


ArthurSens:
cc @bwplotka @macxamin @TheSpiritXIII

expfmt/openmetrics_create_test.go
```diff
@@ -397,14 +403,16 @@ request_duration_microseconds_count 2693
 				Metric: []*dto.Metric{
 					&dto.Metric{
 						Counter: &dto.Counter{
-							Value: proto.Float64(42),
+							Value:            proto.Float64(42),
+							CreatedTimestamp: openMetricsTimestamp,
```
TheSpiritXIII, line 407: Looks like you added created timestamps to all counters and summary tests. Since created timestamp is optional, should we maybe keep a variation with created and not created?

TheSpiritXIII (commented):
Nice work! Just need that client_model changed upstream really :)

expfmt/openmetrics_create.go
```diff
@@ -169,6 +171,10 @@ func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily) (written int
 				metric.Counter.GetValue(), 0, false,
 				metric.Counter.Exemplar,
 			)
+			if metric.Counter.CreatedTimestamp != nil {
```
bwplotka, line 174: Are we sure this will not add OpenMetrics confusing feature to all callers/importers of this functionality? It is used by client_golang which might decide to remove CT from dto, but it feels safer and generally easier to use if this function would accept another argument (or we create another function e.g. `MetricFamilyToOpenMetricsWithCTSeries`) to accomplish this in opt-in fashion?

ArthurSens:
@TheSpiritXIII @bwplotka I opened this PR before I understood what needs to be done to have protobuf working in client_golang. Looking at this now, it seems unnecessary since it is focused on OM's text.

Do we want to proceed anyway? (happy to address the comments if this is still useful)

bwplotka:
I think this might be useful, especially once we implement _Created TS handling for OM in Prometheus (and to test that). But it's low prio, so feel free to close for now OR finish it by adding second function / option for opt-in created timestamp series (:

expfmt/openmetrics_create_test.go
```diff
@@ -397,14 +403,16 @@ request_duration_microseconds_count 2693
...
 						},
 					},
 				},
 			},
 			out: `# HELP foos Number of foos.
 # TYPE foos counter
 foos_total 42.0
+foos_created 12345.6
```
roidelapluie, line 415: Are we sure about the order?

SuperQ: Yes, this is the correct order.

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily-1

ArthurSens:
Once https://github.com/prometheus/prometheus/pull/12733 gets merged I'll get back to this and start working towards ingesting _created from OM :)

ArthurSens:
@bwplotka @roidelapluie 

I'm finally continuing this PR! Rebased and made printing of created lines optional as requested

ArthurSens:
I'm now looking at how client_golang can use this, maybe this PR will need to change strategies 🤔 

Client golang calls [NegotiateWithOpenMetrics](https://github.com/prometheus/common/blob/658f673b7580e72a7d0c0fac26a67858b782fbd5/expfmt/encode.go#L86) to decide which content type to use, and then a [new encoder](https://github.com/prometheus/common/blob/658f673b7580e72a7d0c0fac26a67858b782fbd5/expfmt/encode.go#L118) is created according to the chosen format.

https://github.com/prometheus/client_golang/blob/e96fb182c22dee42f79827f9b78b4e283a5f8589/prometheus/promhttp/http.go#L162-L183

If we add another option to NewEncoder, telling that there is yet another OM option we will need another option for content-type negotiation as well 🤔 For example:

```go
	case FmtOpenMetrics_0_0_1, FmtOpenMetrics_1_0_0:
		return encoderCloser{
				_, err := MetricFamilyToOpenMetrics(w, v)
				return err
			},
			close: func() error {
				_, err := FinalizeOpenMetrics(w)
				return err
			},
		}
	case FmtOpenMetrics_with_CT:
				_, err := MetricFamilyToOpenMetrics(w, v, WithCreatedLines())
				return err
			},
			close: func() error {
				_, err := FinalizeOpenMetrics(w)
				return err
			},
		}
```

And that doesn't sound like a good approach.

ArthurSens:
Added an extra commit that adds the same `ToOpenMetricsOption` argument to `NewEncoder()`, so downstream projects can safely opt-in for the extra created lines

ArthurSens:
@bwplotka, @TheSpiritXIII @SuperQ 

It's been a while, but I'm finally continuing this PR and will work on supporting _created lines in OM Text in client_golang and Prometheus afterward.

WIP PRs:
* https://github.com/prometheus/prometheus/pull/13506
* https://github.com/prometheus/client_golang/pull/1408


Do you mind giving reviewing this one again?

expfmt/encode.go
```diff
@@ -116,7 +116,13 @@ func NegotiateIncludingOpenMetrics(h http.Header) Format {
...
 // to the Encoder interface directly. The current version of the Encoder
 // interface is kept for backwards compatibility.
-func NewEncoder(w io.Writer, format Format) Encoder {
+//
+// NewEncoder can be called with additional options to customize the OpenMetrics text output.
+// For example:
+// NewEncoder(w, FmtOpenMetrics_1_0_0, WithCreatedLines())
+//
+// Extra options are ignored for all other formats.
+func NewEncoder(w io.Writer, format Format, options ...ToOpenMetricsOption) Encoder {
```
bwplotka, line 125: Hm, I think given we can't add additional variadic argument it has be more open, no?

```suggestion
func NewEncoder(w io.Writer, format Format, options ...EncoderOption) Encoder {
```

expfmt/openmetrics_create.go
```diff
@@ -22,11 +22,25 @@ import (
...
 	dto "github.com/prometheus/client_model/go"
 )
 
+type toOpenMetrics struct {
+	withCreatedLines bool
+}
+
+type ToOpenMetricsOption func(*toOpenMetrics)
+
+func WithCreatedLines() ToOpenMetricsOption {
```
bwplotka, line 38: Maybe add comment why this can be dangerous to use and e.g. we consider changing this? https://github.com/prometheus/prometheus/issues/13537

bwplotka (commented):
Nice, well-done, only a small thing, looking good  👍🏽

expfmt/openmetrics_create.go
ArthurSens: Good point!

ArthurSens:
Conflicts resolved again 😅

expfmt/openmetrics_create.go
```diff
@@ -473,6 +515,50 @@ func writeOpenMetricsNameAndLabelPairs(
...
+	written++
+	if err != nil {
+		return written, err
+	}
+
+	ts := createdTimestamp.AsTime()
+	// TODO(beorn7): Format this directly from components of ts to
+	// avoid overflow/underflow and precision issues of the float
+	// conversion.
+	n, err = writeOpenMetricsFloat(w, float64(ts.UnixNano())/1e9)
```
SuperQ, line 548:
```suggestion
	n, err = writeOpenMetricsFloat(w, float64(createdTimestamp.AsTime().UnixNano())/1e9)
```

SuperQ (approved):
Minor nit about some memory allocation, otherwise LGTM.

PR #505, opened 2023-07-31
Add CRL support to Prometheus

RoryKirsi:
This commit adds support for certificate revocation status by Certificate Revocation List (CRL) in Prometheus.

New fields 'CRL' and 'CRLFile' have been added to the Prometheus config ('tls_config' section). These enable certificate revocation validation with the provided CRL.

The Prometheus loads the Certificate Revocation List (CRL) to validate the revocation status of the peer's certificate chain by invoking the 'verifyPeerCertificate' (https://pkg.go.dev/crypto/tls) function during a TLS handshake.


RoryKirsi:
@roidelapluie Hi, Julien.
In case you missed my pull request, I tagged you here to help my commit's review. Thank you!

roidelapluie:
I think that CRL is questionable: the fact we have to implement it ourselve is prone to add security bugs, additionally, there are concerns about CRL in general by the crypto/tls author itsels: https://www.imperialviolet.org/2014/04/19/revchecking.html

What about we implement this when the go team decides to implement it in crypto/tls ?

PR #512, opened 2023-09-18
allow configuration of PingTimeout/ReadIdleTimeout

dongjiang1989:
Allow configuration of PingTimeout/ReadIdleTimeout  at the HTTP2.

`PingTimeout` defaults to `15s`, which is relaxed enough in most network conditions.

`ReadIdleTimeout` defaults to `1m`,  which I think is not going to cause performance issues.

Ref. https://github.com/kubernetes/kubernetes/pull/95981

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -52,6 +52,18 @@ var (
 		// 5 minutes is typically above the maximum sane scrape interval. So we can
 		// use keepalive for all configurations.
 		idleConnTimeout: 5 * time.Minute,
+		// readIdleTimeout is the timeout after which a health check using ping
+		// frame will be carried out if no frame is received on the connection.
+		// Note that a ping response will is considered a received frame, so if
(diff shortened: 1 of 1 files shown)


dongjiang1989:
> What is your usecase? I feel like PingTimeout might be tricky, how well is that supported by http2 vendors?

When I use common to establish an http2 connection to nginx, the connection is found to be use of closed network connection.

Ref. https://github.com/kubernetes/kubernetes/pull/95981

Using PingTimeout, the client will find that the connection is complete and respond, and actively close the connection.

roidelapluie:
Where do you plan to configure this? Are you using common in your own software or is it for prometheus?

dongjiang1989:
> Where do you plan to configure this? Are you using common in your own software or is it for prometheus?

It is used for prometheus and prometheus-adapter frist.

PR #523, opened 2023-10-09
Support environment variables for LOG_LEVEL and LOG_FORMAT

OskarsPakers:
Added Envars for log level and log format to support reading from environment variables.
Useful when running exporter within docker container such as Oracle exporter.
Related issue: https://github.com/iamseth/oracledb_exporter/issues/342


SuperQ (commented):
Sorry, we don't generally use env vars for configuration flags. Docker containers can support arguments, so there is no support issue here.

PR #533, opened 2023-11-22
model: add metric type values

bboreham:
Taken from https://github.com/prometheus/prometheus/blob/a2d5c022985013221a0c7700e92690afe211c521/model/textparse/interface.go#L115-L127

Also found at https://github.com/prometheus/client_golang/blob/297fea317b926e4bb8d3c72022e172def2f68ed2/api/prometheus/v1/api.go#L433-L441


SuperQ (approved):
LGTM

Issue #527, opened 2023-10-25
Implement changes needed to support UTF-8 metric and label names in Prometheus


ywwg:
ref: https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md

Some changes are needed in this library to support the utf8 proposal:

- [x] openmetrics_create needs to understand and construct new syntax (POC exists) https://github.com/prometheus/common/pull/537
- [x] library needs to support either legacy mode or utf8 mode, with a way to check for either as well. https://github.com/prometheus/common/pull/537
- [x] #554
- [x] https://github.com/ywwg/common/pull/2

More tasks will be added as code comes together.

ywwg:
this is done!

PR #537, opened 2023-11-29
UTF-8 support in validation, and some parsers and formatters

ywwg:
Adds UTF8 metric and label name validation mode and adds support for parsing and formatting the UTF8 text format. 

The expfmt package will use the new quoting syntax iff the metric name does not conform to the legacy name format.  Old metrics will still be presented as `foo{}` whereas utf8 metrics will be presented as `{"foo"}`. (Formatters already assume that names are already validated and will not throw an error if names are invalid.)

The model package has a new library flag, `NameValidationScheme`, which determines how validation will be done on all calls. If a caller sometimes needs to validate utf8 but sometimes also needs to validate against the legacy scheme, they should use IsValidLegacyMetricName where needed. Changing NameValidationScheme after startup, especially if there are multiple goroutines active, is not recommended and can result in undefined behavior.

NameValidationScheme is Legacy by default, so clients of this library should see no changes in behavior unless they change that value.

There is some minor style cleanup of redundant type names in tests.

Works towards https://github.com/prometheus/common/issues/527

Future work will be needed on the content negotiation code.


expfmt/text_create.go
```diff
@@ -330,32 +323,67 @@ func writeSample(
...
+		// If the name does not pass the legacy validity check, we must put the
+		// metric name inside the braces. Note, it will already have been
+		if !model.IsValidLegacyMetricName(model.LabelValue(name)) {
+			metricInsideBraces = true
+			err := w.WriteByte(separator)
+			written++
+			if err != nil {
+				return written, err
+			}
+			name = fmt.Sprintf(`"%s"`, name)
```
ywwg, line 355: same as before (and elsewhere)

beorn7:
I'll try to get to this ASAP.

@ywwg I assume this is ready for review and will change the title accordingly.

@roidelapluie I plan to do a code level review, but eventually, we'll need your approval, too.

model/labels.go
```diff
@@ -97,17 +97,25 @@ var LabelNameRE = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 // therewith.
 type LabelName string
 
-// IsValid is true iff the label name matches the pattern of LabelNameRE. This
-// method, however, does not use LabelNameRE for the check but a much faster
-// hardcoded implementation.
+// IsValid returns true iff name matches the pattern of LabelNameRE
+// for legacy names, and iff it's valid UTF-8 if isUtf8 is true.
```
beorn7, line 101: Outdate doc comment? (This should refer to the global `NameValidationScheme` rather than `isUtf8`.)

model/metric.go
```diff
@@ -18,13 +18,25 @@ import (
...
 	"sort"
 	"strings"
+	"unicode/utf8"
+)
+
+type validationSchemeId int
+
+const (
+	LegacyValidation = validationSchemeId(0)
+	UTF8Validation   = validationSchemeId(1)
```
beorn7, line 28: Does this work?

```suggestion
	LegacyValidation validationSchemeId = iota
	UTF8Validation
```

Maybe make `validationSchemeId` and call it `ValidationScheme` ? That would be closer to how I understand Go (pseudo) enums.

beorn7 (changes requested):
Thanks for doing this.

As discussed, the package-level "config variable" might be the least bad way of "flipping the switch". However, I'm not sure if just one switch is enough for our use case. I see three modes:
1. Use this library in exactly the old way, i.e. this PR shouldn't change the behavior at all.
2. Use this library in the new way, i.e. all UTF-8 characters are allowed in names, and new exposition formats are used accordingly.
3. Allow the new way but still accommodate legacy clients, i.e. special UTF-8 characters need to be escaped with the schema from the design doc.

It seems this PR fulfills (1) and (2), by setting `NameValidationScheme` accordingly. However, the `expfmt` package also implements content negotiation, and this has to be modified to allow mode (3).

In different news, `text_parse.go` eventually needs an update to parse the new classic text format. In Prometheus, we use a different parser implementation, so it's not in the critical path to make Prometheus work, but there should be at least a TODO somewhere.

beorn7:
About the `strconv.Quote` thing: text_create.go contains `writeEscapedString`, which should do the trick.

expfmt/openmetrics_create_test.go
```diff
@@ -82,20 +82,92 @@ name{labelname="val1",basename="basevalue"} 42.0
 name{labelname="val2",basename="basevalue"} 0.23 1.23456789e+06
 `,
 		},
-		// 1: Gauge, some escaping required, +Inf as value, multi-byte characters in label values.
+		// 0.5: Dots in name
```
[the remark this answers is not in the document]
ywwg: oh I thought I changed that! sorry

model/metric.go
ywwg: nice -- I couldn't figure out the syntax

ywwg:
> However, the `expfmt` package also implements content negotiation, and this has to be modified to allow mode (3).

I believe this is covered with the existence of the IsValidLegacyMetricName function -- the library can be enabled for mode 2 in general, and then specific calls can apply that function when needed.

Ah, I missed text_parse.go! I will put that on the list

beorn7:
> > However, the `expfmt` package also implements content negotiation, and this has to be modified to allow mode (3).
>
> I believe this is covered with the existence of the IsValidLegacyMetricName function -- the library can be enabled for mode 2 in general, and then specific calls can apply that function when needed.
>
> Ah, I missed text_parse.go! I will put that on the list

I was more thinking about the code in expfmt/encode.go and expfmt/expfmt.go.

We need new Content-Type strings for "format XYZ but with arbitrary UTF-8 characters allowed in names" (see design doc).

And then the `Negotiate` and `NegotiateIncludingOpenMetrics` functions need to return them appropriately, and then `NewEncoder` has to return a correspondingly configured `Encoder`.

ywwg:
ah ok. I'll take a look

ywwg:
OK so the question is how to select the escaping mode for old recipients, and that mode might be different for different situations. I think we'll use a similar pattern to the parsing situation: a library-wide setting for default escaping, and then a new function where users can specify the scaping.

So the current MetricFamilyToOpenMetrics and MetricFamilyToText functions will use the library-default escaping, and then I'll make a couple new functions that take an extra parameter to specify an escaping on a call-by-call basis.

ywwg:
It might make more sense to shrink this PR down so it just handles the parsing, and then do another one to handing the exposition format creation, what do you think? Otherwise it might start to get large

ywwg:
open question: who is in charge of which escaping mechanism gets used? should that be part of the content negotiation, like the scraper says "I wish to have things escaped this way" or is that a setting on the metrics producer?

ywwg:
converting to draft while I do some work

beorn7:
> open question: who is in charge of which escaping mechanism gets used? should that be part of the content negotiation, like the scraper says "I wish to have things escaped this way" or is that a setting on the metrics producer?

The MVP here is that any UTF-8-enabled binary instrumented with prometheus/client_golang should still be able to be scraped by a non-UTF-8-enabled scraper. I would assume that means to use the `U__` escaping schema described in the design doc. This should be done via content negotiation, as also described in the design doc.

Allowing other escaping schemas would IMHO be "nice to have" (so that users can pick them if they need them for compatibility with legacy systems). Those other escaping schemas already exist, but they have no support in client_golang. Therefore, I see no pressing needs to add support for them now. (One might even argue that we actually want to discourage the other schemas ASAP, and providing UTF-8 support plus a sophisticated escaping schema for old backends has been invented precisely to get rid of those other, problematic escaping schemas.) But if we want to support them, I guess we should use the same knob to select them, i.e. content negotiation. It ultimately depends on the use case, but I'm not sure yet what the use case for those other escaping schemas will be.

ywwg:
After doing a bunch of work, indeed I think it's best to put content negotiation in a separate PR.  I believe this PR still does what it says and should not negatively affect people using this library (unless they turn on the UTF8 validation mode, which is not really supported anywhere else and will result in errors)

ywwg:
(I will squash this and sign the commit when review is done to solve the DCO failure)

beorn7:
> After doing a bunch of work, indeed I think it's best to put content negotiation in a separate PR. I believe this PR still does what it says and should not negatively affect people using this library (unless they turn on the UTF8 validation mode, which is not really supported anywhere else and will result in errors)

Fair enough. But I think we need to document a bit better what the precise behavior after this PR is and what still has to be done. I'm mostly worried that it could now happen very easily that somebody uses the code to produce metrics rendered in the new UTF-8 capable style after negotiating the existing old formats. (Once content negotiation is fully implemented, the encoders would use the underscore encoding described in the design doc in that case.)

I guess the doc comment for `func MetricFamilyToText` and `func MetricFamilyToOpenMetrics` should include how the output changes if metric names or label names are using characters excluded in the legacy format, and that it is the callers responsibility to not include names with those characters if valid legacy format is desired.

beorn7:
To clarify: Despite the title, this PR doesn't really change the parsing side, does it? I mean, the configurable validation implicitly adds support to the protobuf decoder. But the other parsers don't change. Maybe that could be added as a TODO.

Also, a TODO in the content negotiation part might help to avoid confusion.

beorn7 (changes requested):
Codewise LGTM. Just need the hints in doc comments as discussed in the comments.

ywwg:
> I guess the doc comment for `func MetricFamilyToText` and `func MetricFamilyToOpenMetrics` should include how the output changes if metric names or label names are using characters excluded in the legacy format

the comment already says  "this function assumes the input is already sanitized and does not perform any
// sanity checks. If the input contains duplicate metrics or invalid metric or
// label names, the conversion will result in invalid text format output."  

Do you want more than that?

ywwg:
oh I get it now, adding...

expfmt/openmetrics_create.go
```diff
@@ -35,6 +35,12 @@ import (
 // sanity checks. If the input contains duplicate metrics or invalid metric or
 // label names, the conversion will result in invalid text format output.
 //
+// If metric names conform to the legacy validation pattern, they will be placed
+// outside the brackets in the traditional way, like `foo{}`. If the metric name
+// fails the legacy validation check, it will be placed quoted inside the
+// brackets: `{"foo"}`. As stated above, the input is assumed to be santized and
+// no error will be thrown in this case.
```
beorn7, line 42: Very good. But we also need a similar sentence about _label_ names (which are quoted if they contain formerly invalid characters).

expfmt/text_create.go
```diff
@@ -62,6 +62,12 @@ var (
 // contains duplicate metrics or invalid metric or label names, the conversion
 // will result in invalid text format output.
 //
+// If metric names conform to the legacy validation pattern, they will be placed
+// outside the brackets in the traditional way, like `foo{}`. If the metric name
+// fails the legacy validation check, it will be placed quoted inside the
+// brackets: `{"foo"}`. As stated above, the input is assumed to be santized and
+// no error will be thrown in this case.
```
beorn7, line 69: Same comment as above.

expfmt/text_create.go
```diff
@@ -330,32 +324,62 @@ func writeSample(
...
 }
 
-// writeLabelPairs converts a slice of LabelPair proto messages plus the
+// writeNameAndLabelPairs converts a slice of LabelPair proto messages plus the
 // explicitly given additional label pair into text formatted as required by the
 // text format and writes it to 'w'. An empty slice in combination with an empty
 // string 'additionalLabelName' results in nothing being written. Otherwise, the
 // label pairs are written, escaped as required by the text format, and enclosed
 // in '{...}'. The function returns the number of bytes written and any error
 // encountered.
```
beorn7, line 333: The doc comment here would benefit from an update for the new features (additionally passed `name`, special syntax/quoting for `name` and label names containing formerly invalid characters).

beorn7 (changes requested):
Thank you very much. And sorry for requesting a few more doc comment tweaks.

About the TODO I would like to see in the content negotiation part: In https://github.com/prometheus/common/blob/main/expfmt/expfmt.go , there should be a TODO that reads something like this: "The content-type strings here are all for the legacy exposition formats, where valid characters for metric names and label names are limited. The support for arbitrary UTF-8 characters in those names is already partially implemented in this module (see model.ValidationScheme), but to actually use it on the wire, new content-type strings have to be agreed upon and added here."

expfmt/openmetrics_create.go
ywwg: done (pending push)

expfmt/text_create.go
ywwg: done (pending push)

expfmt/text_create.go
ywwg: done

beorn7 (changes requested):
Thank you very much.

But what do you think about my earlier comment? This one:

> About the TODO I would like to see in the content negotiation part: In https://github.com/prometheus/common/blob/main/expfmt/expfmt.go , there should be a TODO that reads something like this: "The content-type strings here are all for the legacy exposition formats, where valid characters for metric names and label names are limited. The support for arbitrary UTF-8 characters in those names is already partially implemented in this module (see model.ValidationScheme), but to actually use it on the wire, new content-type strings have to be agreed upon and added here."

Maybe you just missed it?

ywwg:
ah, sorry about that! done now

beorn7 (approved):

beorn7:
Thank you very much.

PR #538, opened 2023-11-30
Use common interface to fetch secrets in HTTP client config

TheSpiritXIII:
Part 1 for secret provider support within `common`. See: https://groups.google.com/g/prometheus-developers/c/WKOej_pnhXg

This change consolidates fetching secrets across all authorization types. Authorization types which previously had different structs per secret type now have a single type.

The unification helps with adding secret manager support which part 2 will include. There are two ways we could do this.
1. With a secret "reference" (or "handle"), e.g. a name to a secret which is configured somewhere else. This PR implements this approach: https://github.com/prometheus/common/pull/572
2. Inline. This would not be backwards-compatible. The only way I could think of serialization to work with the type defined outside of this package (secret provider configs would be part of the main prometheus repository since they're plugins), is to use a generic secret config type.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unsafe"
 
 	conntrack "github.com/mwitkow/go-conntrack"
 	"golang.org/x/net/http/httpproxy"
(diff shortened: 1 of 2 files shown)


config/http_config.go
```diff
@@ -589,59 +586,93 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
-		req.Header.Set("Authorization", fmt.Sprintf("%s %s", rt.authType, string(rt.authCredentials)))
+func (s *inlineSecret) fetch() (string, error) {
+	return s.text, nil
+}
+
+func (s *inlineSecret) description() string {
+	return "inline"
+}
+
+func (s *inlineSecret) immutable() bool {
```
bwplotka, line 615: nice, this will optimize things 👍🏽

bwplotka (approved):
LGTM, thanks!

Exciting for https://github.com/prometheus/prometheus/pull/13955#issuecomment-2090614673 which this PR enables! ❤️ 

*Disclaimer: @TheSpiritXIII is in my Google team. I tried to get more hands on this review for weeks, but no response. The secret provider feature was generally accepted [1](https://github.com/prometheus/alertmanager/issues/3108), [2](https://github.com/prometheus/prometheus/pull/13955#issuecomment-2090614673), so allowing myself to move forward after the careful review*

alanprot:
Should we export the secret types implementations or the `secretFrom` methods? 

Downstream projects like thanos/cortex uses the methods that now receives the `secret` interface but both interface and the implementations methods are not being exported:

Ex: 

https://github.com/thanos-io/thanos/blob/58447e1a6e354470f746cfd5d37dc080acb569a3/pkg/clientconfig/http.go#L164-L175

Ex of how we changed on this PR: 

https://github.com/prometheus/common/pull/538/files#diff-82fa1ce44db0403c7b0bac6e1fa6030746f8ef9e3958ad4f74637dd621d3cda3L724

config/http_config.go
```diff
@@ -589,59 +586,93 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
 
-// NewAuthorizationCredentialsFileRoundTripper adds the authorization
-// credentials read from the provided file to a request unless the authorization
-// header has already been set. This file is read for every request.
-func NewAuthorizationCredentialsFileRoundTripper(authType, authCredentialsFile string, rt http.RoundTripper) http.RoundTripper {
-	return &authorizationCredentialsFileRoundTripper{authType, authCredentialsFile, rt}
+// NewAuthorizationCredentialsRoundTripper adds the authorization credentials
+// read from the provided secret to a request unless the authorization header
+// has already been set.
+func NewAuthorizationCredentialsRoundTripper(authType string, authCredentials secret, rt http.RoundTripper) http.RoundTripper {
```
nicolasparada, line 662: Breaking change. Also `secret` is not exported.

nicolasparada: Ah, fixed already https://github.com/prometheus/common/pull/650

bwplotka: yup, plus the lib is 0.x.z so we can break it. We try to do it rarely and only if there is a good reason - here it was useful to make it more consistent if we can, sorry for pain though.

PR #544, opened 2023-12-02
Add unit

vesari:
Not ready for review, still a WIP. Dependency of PR https://github.com/prometheus/client_golang/pull/1392


vesari:
@ArthurSens this is now ready for review, thanks.

ArthurSens:
Implementation looks good to me! 

On the other hand, we had a [brief discussion on slack](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1701897052520909) and we suspect that enforcing units to be present in the MetricFamily name is a breaking change that we should be careful about.

We know that client_golang uses prometheus/common but maybe there are others? We could make this method opt-in with something like this

```go
type ToOpenMetrics struct {
  withUnit bool
  // withCreated bool can be added in the future
}

type ToOpenMetricsOption = func(*ToOpenMetrics)

func ToOpenMetricsWithUnit() ToOpenMetricsOption {
  return func(om *ToOpenMetrics) {
    om.withUnit = true
  }
}

func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...ToOpenMetricsOption) (written int, err error) {
  toOM = ToOpenMetrics{}
  for _, option := range options {
    option(&toOM)
  }
.
.
.
  if toOM.withUnit && in.Unit != nil {
  // print Unit as usual
  }
.
.
.
}
```

Downstream projects could safely bump promethes/common version, while having the option to adopt units when appropriate:

```go
// Will keep the same behavior as today, even if unit is set
common.MetricFamilyToOpenMetrics(out, in)

// Will add unit to the metricFamily name as suffix
common.MetricFamilyToOpenMetrics(out, in, common.ToOpenMetricsWithUnit())
```

vesari:
@ArthurSens this is now ready for review. I hope I have interpreted correctly both your suggestions and the gist of your Slack conversation.

ArthurSens (commented):
Great improvement! Have you tried using this version of prometheus/common in your PR in client_golang? How did the changes here impacted over there?

I suspect we also need some changes to [expfmt.NewEnconder](https://github.com/prometheus/common/blob/7e4424239c53cb194ccf81ed41cf67a414663d3d/expfmt/encode.go#L119) as well.

ArthurSens:
pinging @beorn7 and @bwplotka, maybe you also have some input here

ArthurSens (commented):
Could we also have some tests for the case where the unit is set, but the option withUnit is not?

beorn7:
Looks good in general. I don't have the time for a detailed review, but I trust you to do that alright on your own.

I should also add that I don't feel qualified to make any calls how to handle OpenMetrics and its quirks appropriately. (Not that I wouldn't have context. It's more that I have too much context, and that clouds my judgement.)

vesari:
> Great improvement! Have you tried using this version of prometheus/common in your PR in client_golang? How did the changes here impacted over there?
> 
> I suspect we also need some changes to [expfmt.NewEnconder](https://github.com/prometheus/common/blob/7e4424239c53cb194ccf81ed41cf67a414663d3d/expfmt/encode.go#L119) as well.

Thank you very much :) I have tried it on my branch (https://github.com/prometheus/client_golang/pull/1392) and nothing seems to break or fail; however, I'll have to check again once I've modified `expfmt.NewEncoder` (thanks a lot for pointing that out!). I'll now convert this PR to draft for the time being, while I carry out the new changes and figure out their impact and ramifications in `client_golang`.

vesari:
@ArthurSens thank you very much for your guidance! I've addressed your change requests. This is now ready for review. About the `client_model` version upgrade, I’m not sure how to proceed, as I would think only maintainers are allowed to change it? Thanks for letting me know :)

vesari:
@SuperQ done. Thank you!

model/labelset.go
```diff
@@ -131,27 +129,14 @@ func (l LabelSet) Merge(other LabelSet) LabelSet {
...
-		}
-		b.WriteString(string(name))
-		b.WriteByte('=')
-		b.Write(strconv.AppendQuote(b.AvailableBuffer(), string(l[name])))
-	}
-	b.WriteByte('}')
-	return b.String()
+
+	sort.Strings(lstrs)
+	return fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))
```
ArthurSens, line 139: Accidental change with the rebase?

model/labelset_test.go
```diff
@@ -120,23 +117,3 @@ func TestLabelSetMerge(t *testing.T) {
...
-	ls := make(LabelSet)
-	ls["monitor"] = "codelab"
-	ls["foo2"] = "bar"
-	ls["foo"] = "bar"
-	ls["abc"] = "prometheus"
-	ls["foo11"] = "bar11"
-	for i := 0; i < b.N; i++ {
-		_ = ls.String()
-	}
-}
```
ArthurSens, line 142: same thing here, probably by accident?

ArthurSens (approved):
@vesari LGTM! thanks for your patience through all those months :)

I've noticed that a few unrelated changes were made, probably by mistake? Could you take a look?

vesari:
@ArthurSens thanks a lot for pointing that out, I'll have a look at it tomorrow morning then get back to you immediately after, thanks again

vesari:
@ArthurSens , I have now amended my previous commit and rebased again. Thank you very much for catching those 2 files that I had indeed not modified actively; like you, I can just assume something happened during the previous rebase. I hope now it's ready for merging :)

ArthurSens:
Awesome, could someone with merge rights also give a review?

Maybe @bwplotka, @SuperQ or @beorn7 ?

SuperQ (approved):
Thanks!

PR #546, opened 2023-12-05
Allow using empty Authorization credentials

### LINKED ISSUE #545: Authorization credentials should be optional
As part of [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-introduction), authorization is defined in ABNF as:

```
Authorization = auth-scheme [ 1*SP ( token68 / [ auth-param *( OWS "," OWS auth-param ) ] ) ]
```

This means that the scheme MUST be provided but credentials are optional.

TheSpiritXIII:
Resolves #545. Looks like we didn't have any validation here. That may make this a breaking change if users rely on this (incorrect) behavior.

cc @roidelapluie who seems to have written most of the original code. :)

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -546,10 +546,10 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
 
 		// If a authorization_credentials is provided, create a round tripper that will set the
 		// Authorization header correctly on each request.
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
LGTM! 

Waiting for @roidelapluie for the final approve.

roidelapluie:
Thanks!

PR #547, opened 2023-12-06
Support empty OAuth2 inline secrets

### LINKED ISSUE #528: OAuth 2 client_secret should be optional
After #294, client_secret was made to be required.

According to the OAuth 2.0 spec (https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), it is optional:
> The client MAY omit the parameter if the client secret is an empty string.

You can see this with OAuth 2.0 providers points out it's optional, e.g. [Microsoft](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-access-token-with-a-client_secret).

We even show it as optional in the [configuration documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#oauth2) but this is not what our behavior _actually_ is.

While it is best practice to include it in production environments, we should let users decide. Alternatively, let's be consistent and clear in our documentation if we diverge from expectations.

TheSpiritXIII:
Resolves #528.

`client_secret_file` also supports using an empty file for no client secret. The same is not true for `client_secret`. This change makes it so that if neither are set or if `client_secret` is an empty string, that will be used for the secret.

This also adds a subtest on `client_secret_file` for testing with an empty file.

cc @roidelapluie who seems to have written most of the original code. :)

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -378,9 +378,6 @@ func (c *HTTPClientConfig) Validate() error {
 		if len(c.OAuth2.ClientID) == 0 {
 			return fmt.Errorf("oauth2 client_id must be configured")
 		}
-		if len(c.OAuth2.ClientSecret) == 0 && len(c.OAuth2.ClientSecretFile) == 0 {
-			return fmt.Errorf("either oauth2 client_secret or client_secret_file must be configured")
-		}
 		if len(c.OAuth2.TokenURL) == 0 {
 			return fmt.Errorf("oauth2 token_url must be configured")
 		}
@@ -729,13 +726,12 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		rt.mtx.RLock()
 		changed = secret != rt.secret
 		rt.mtx.RUnlock()
+	} else {
+		// Either an inline secret or nothing (use an empty string) was provided.
+		secret = string(rt.config.ClientSecret)
 	}
 
 	if changed || rt.rt == nil {
-		if rt.config.ClientSecret != "" {
-			secret = string(rt.config.ClientSecret)
-		}
-
 		config := &clientcredentials.Config{
 			ClientID:       rt.config.ClientID,
 			ClientSecret:   secret,
--- a/config/generate.go
+++ b/config/generate.go
@@ -71,7 +71,6 @@ func SerialNumber() *big.Int {
 	serialNumber.Add(&serial, big.NewInt(1))
 
 	return &serial
-
 }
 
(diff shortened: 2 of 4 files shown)


config/http_config.go
```diff
@@ -729,13 +726,11 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		rt.mtx.RLock()
 		changed = secret != rt.secret
 		rt.mtx.RUnlock()
+	} else {
```
bwplotka, line 729: Perhaps commentary, that this string can be empty and that's ok?

bwplotka (approved):
Epic work, LGTM!

Waiting for @roidelapluie for the final approve.

NOTE @TheSpiritXIII works in my Google Team

TheSpiritXIII:
> What is the use case?

My use-case is this that I am working on the [Google Prometheus Operator for Kubernetes (which we call Managed Prometheus)](https://github.com/GoogleCloudPlatform/prometheus-engine).

We want to provide OAuth2 support and there's essentially two flows: User provides the client secret in a Kubernetes secret or they don't. If they do not specify a Kubernetes secret, the only way to get an empty client secret is to create an empty file and use that, and that involves a lot of setup: create a new empty directory, then use bash script to create an empty file, then mount it to the Prometheus instances and finally update the Prometheus configuration. If the inline client secrets was consistent and allowed empty client secrets like the file option, that would simplify my use-case dramatically. It would essentially be 1 step which is to just not provide the field in the Prometheus configuration.

Why allow users to provide an empty client secret? Because the spec allows for it and some users might want it. For example, they may want to create a simple OAuth2 server which skips secret validation. If you Google-search for "oauth2 empty client string" you will see a plethora of users asking for the feature in various frameworks. It seems it's a common issue that applications do not implement the OAuth2 specification fully.

TheSpiritXIII:
Hi @roidelapluie friendly ping :) do you have any thoughts on the motivating statements?

As an additional motivating factor, BasicAuth `password` and Authorization Header `credentials` both accept empty strings. As mentioned in my other posts, a workaround is to create a `client_secret_file` with empty contents but it involves extraneous setup in the Kubernetes world.

bwplotka (approved):
Given [the consensus on the last DevSummit to support secret providers](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.s5i4bzsff689) let's merging this so the work on providers is unblocked 🤗 

If no objection, will merge later today.

PR #549, opened 2023-12-08
http_config: Add host

jkroepke:
Follow-up: https://github.com/prometheus/prometheus/pull/13157

Host allows to override the hostname from URL. This can be useful if DNS is not stable but the endpoint required an valid host.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -311,7 +311,7 @@ type HTTPClientConfig struct {
 	EnableHTTP2 bool `yaml:"enable_http2" json:"enable_http2"`
 	// Host optionally overrides the Host header to send.
 	// If empty, the host from the URL will be used.
-	Host string `yaml:"host" json:"host"`
+	Host string `yaml:"host,omitempty" json:"host,omitempty"`
 	// Proxy configuration.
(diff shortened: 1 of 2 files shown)


config/http_config.go
```diff
@@ -309,6 +309,9 @@ type HTTPClientConfig struct {
 	// The omitempty flag is not set, because it would be hidden from the
 	// marshalled configuration when set to false.
 	EnableHTTP2 bool `yaml:"enable_http2" json:"enable_http2"`
+	// Host optionally overrides the Host header to send.
+	// If empty, the host from the URL will be used.
+	Host string `yaml:"host" json:"host"`
```
SuperQ, line 314: If this is optional, it should be `omitempty` I think.

```suggestion
	Host string `yaml:"host,omitempty" json:"host,omitempty"`
```

SuperQ (changes requested):

SuperQ (approved):

jkroepke:
Can I help here?

bboreham:
Unfortunately this clashes with the field of the same Yaml name in Prometheus Docker SD.

PR #556, opened 2023-12-19
Add hints for promlog

lucacome:
Adds hints for completion when using promlog.

### DIFF
--- a/promlog/log.go
+++ b/promlog/log.go
@@ -34,6 +34,9 @@ var (
 		func() time.Time { return time.Now().UTC() },
 		"2006-01-02T15:04:05.000Z07:00",
 	)
+
+	LevelFlagOptions  = []string{"debug", "info", "warn", "error"}
+	FormatFlagOptions = []string{"logfmt, json"}
 )
 
 // AllowedLevel is a settable identifier for the minimum level a log entry
--- a/promlog/flag/flag.go
+++ b/promlog/flag/flag.go
@@ -14,6 +14,8 @@
 package flag
 
 import (
+	"strings"
+
 	kingpin "github.com/alecthomas/kingpin/v2"
 	"github.com/prometheus/common/promlog"
 )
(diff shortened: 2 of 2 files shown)


SuperQ (approved):
This is probably fine, but it would be nice if we could have the hint lists defined in the promlog package.

lucacome:
@SuperQ I pushed a new commit, let me know if that was what you were thinking

promlog/log.go
```diff
@@ -26,14 +26,17 @@ import (
...
+// This timestamp format differs from RFC3339Nano by using .000 instead
+// of .999999999 which changes the timestamp from 9 variable to 3 fixed
+// decimals (.130 instead of .130987456).
+var timestampFormat = log.TimestampFormat(
+	func() time.Time { return time.Now().UTC() },
+	"2006-01-02T15:04:05.000Z07:00",
+)
+
+const (
+	LevelFlagOptions  = "debug, info, warn, error"
```
SuperQ, line 38: I would go the other way, make this a slice literal and `strings.Join()` in the help function.

```suggestion
var LevelFlagOptions  = []string{"debug", "info", "warn", "error"}
```

lucacome: I thought about doing that but it seemed like a waste because all we need is a string...

SuperQ: We have to do one string join/split no matter what we do. It makes more sense to join than to split.

promlog/log.go
```diff
@@ -26,14 +26,17 @@ import (
 	"github.com/go-kit/log/level"
 )
 
+// This timestamp format differs from RFC3339Nano by using .000 instead
+// of .999999999 which changes the timestamp from 9 variable to 3 fixed
+// decimals (.130 instead of .130987456).
```
SuperQ, line 31: I think this comment needs to go back where it was.

lucacome: Fixed! I need to get more sleep 😃

SuperQ (approved):
Nice work.

PR #563, opened 2024-01-11
Make version getRevision public

SuperQ:
Make version package `getRevision()` a public function to allow implementing `NewCollector()` outside of the version package easier.


ArthurSens (approved):
For context @bwplotka @kakkoyun, I was talking with Ben and we thought we could move the NewVersionCollector() to client golang

That would be enough to remove the dependency here, but all the other stuff from the version package would not go with client_golang.

bwplotka:
Sounds perfect to me! Thanks 💪🏽

bwplotka (approved):

PR #565, opened 2024-01-11
version: make GetTegs() public

ArthurSens:
So it can be used by external collectors

I think it was forgotten in #563 @SuperQ @bwplotka @kakkoyun


SuperQ (approved):
Thanks!

PR #569, opened 2024-01-13
Avoid off-spec openmetrics exposition when exemplars have empty labels

orls:
Will fix https://github.com/prometheus/client_golang/issues/1333

`client_golang` currently allows creating exemplars with an empty label set, and they get rendered like:

```
...
foos_total 42.0 #  1.0 12345.6
...
```

(Note the double space after the exemplar's opening `#`)

Based on the [current text of the specification](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format), the openmetrics text format effectively requires one label – so it seems best to just ignore the exemplars.

### DIFF
--- a/expfmt/openmetrics_create.go
+++ b/expfmt/openmetrics_create.go
@@ -350,7 +350,7 @@ func writeOpenMetricsSample(
 			return written, err
 		}
 	}
-	if exemplar != nil {
+	if exemplar != nil && len(exemplar.Label) > 0 {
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):
LGTM, thanks for the fix :)

cc @bwplotka or @kakkoyun, who have write access in this repo

SuperQ (approved):

PR #570, opened 2024-01-18
UTF8: Content negotiation (encoding side)

ywwg:
This PR builds on https://github.com/prometheus/common/pull/537 by adding content negotation during scraping.

Scrapers can announce that they support utf-8 names by adding "validchars=utf8" in the Content-Type header. In cases where utf8 is not available, metric providers can be configured to escape names in a few different ways: values (U__ utf value escaping for perfect round-tripping), underscores (all invalid chars become _), dots (dots become _dot_, _ becomes __, all other values become __). Escaping can either be a global default (model.NameEscapingScheme) or can also be specific in Content-Type with the "escaping=" term, which can be "none" (for UTF8-compatible), "underscores", "dots", or "values".

This is still a noop for existing configurations because scrapers will not be passing the validchars or escaping key in the Content-Type header.  Existing functionality is maintained.

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/protobuf/encoding/prototext"
 
 	"github.com/prometheus/common/internal/bitbucket.org/ww/goautoneg"
+	"github.com/prometheus/common/model"
 
 	dto "github.com/prometheus/client_model/go"
 )
@@ -61,23 +62,32 @@ func (ec encoderCloser) Close() error {
 // as the support is still experimental. To include the option to negotiate
 // FmtOpenMetrics, use NegotiateOpenMetrics.
 func Negotiate(h http.Header) Format {
+	escapingScheme := Format(fmt.Sprintf("; escaping=%s", Format(model.NameEscapingScheme.String())))
 	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
+		if escapeParam := ac.Params[model.EscapingKey]; escapeParam != "" {
+			switch Format(escapeParam) {
+			case model.AllowUTF8, model.EscapeUnderscores, model.EscapeDots, model.EscapeValues:
+				escapingScheme = Format(fmt.Sprintf("; escaping=%s", escapeParam))
+			default:
+				// If the escaping parameter is unknown, ignore it.
+			}
+		}
 		ver := ac.Params["version"]
 		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
 			switch ac.Params["encoding"] {
 			case "delimited":
-				return FmtProtoDelim
+				return FmtProtoDelim + escapingScheme
 			case "text":
-				return FmtProtoText
+				return FmtProtoText + escapingScheme
 			case "compact-text":
-				return FmtProtoCompact
+				return FmtProtoCompact + escapingScheme
 			}
 		}
 		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
-			return FmtText
+			return FmtText + escapingScheme
 		}
 	}
-	return FmtText
+	return FmtText + escapingScheme
 }
 
 // NegotiateIncludingOpenMetrics works like Negotiate but includes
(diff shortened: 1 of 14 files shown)


ywwg:
This will be squashed :)

expfmt/encode.go
```diff
@@ -61,53 +63,113 @@ func (ec encoderCloser) Close() error {
...
+				fmt.Println("NEGOTIATE result open", FmtOpenMetrics_1_0_0+escapingScheme)
+				return FmtOpenMetrics_1_0_0 + escapingScheme
+			default:
+				return FmtOpenMetrics_0_0_1 + escapingScheme
 			}
-			return FmtOpenMetrics_0_0_1
 		}
 	}
-	return FmtText
+	return FmtText_0_0_4 + Format(fmt.Sprintf("; escaping=%s", model.EscapeValues))
```
ywwg, line 172: this specifies a default escaping scheme when none other is specified.  not sure if this is the right thing to do

beorn7 (commented):
General direction looks OK. I will only manage to do a code level review next week.

Some broad thoughts:

First the versioning. See the individual comment below. It feels weird to pretend to support a version that doesn't exist yet. (But as said, I'm not an expert and I don't know how this is usually done.)

Another gut feeling: `validchars` and `escaping` never show up together. Maybe they can be merged into one option? `escaping=none` could take the role of `validchars=utf8`, and `escaping=values` is the default. (Maybe that default should be part of the (upcoming) standard and hardwired in the code. It feels weird that the default is set via a package variable and that would change the behavior for the same accept header.)

beorn7:
Another thought: Maybe we don't need to change version numbers at all right now. Any escaping that is not `none` will produce valid classic text format or OMv1. So we could just document the `escaping=none` (currently `validchars=utf8`, see comment above) as an experimental feature that creates a slightly different format, that will be part of a future standard. We can later decide if we want to keep supporting the "illegal" combination of OMv1 and `escaping=none`, but that's still better than asking clients to negotiate OMv2 although OMv2 will certainly look much differently once it really exists.

ywwg:
I like the idea of using escaping to also select for utf8 mode. maybe instead of "none" we call it `escaping=utf8-ok` or something like that, to make it more explicit?

ywwg:
can't remember if dashes are allowed in content-type parameters though

beorn7:
Yes, let's do that. I.e. only one parameter `escaping` that defines the behavior (missing = old behavior, any escaping schema = do the escaping, `allow-utf-8` or something = use the new format).

I would also say, for now, that we don't change the version numbers but "tolerate" that the client can ask for `allow-utf-8` and we pragmatically create an exposition format that is technically invalid. Just add a comment in the code that says that this is just for experimentation and that there will be new versions of the expositions formats eventually that properly describe the new syntax and should be selected whenever UTF-8 names are requested.

And yes, dashes are allowed in the content type, we use it for `charset=utf-8` and `encoding=compact-text`.

The former reminds me that we should try to be consistent and spell UTF-8 as UTF-8 where possible.

beorn7:
The Go test errors seem to have to do with Go modules. Probably need to run `go mod tidy` or something.

Could you fix that and work in the ideas above? I think the code otherwise is technically fine. I'll do a final review once that's done.

ywwg:
oh except for adding comments

expfmt/encode.go
```diff
@@ -21,6 +21,8 @@ import (
 	"google.golang.org/protobuf/encoding/protodelim"
 	"google.golang.org/protobuf/encoding/prototext"
 
+	"github.com/prometheus/common/model"
+
```
beorn7, line 25: I would not have a blank line here.

beorn7 (approved):
Looks good. I assume you will squash the commits yourself and then take care of DCO signing.

model/metric.go
```diff
@@ -31,27 +50,50 @@ const (
...
-	// in isolation from other components that don't support UTF8 may result in
-	// bugs or other undefined behavior. This value is intended to be set by
-	// UTF8-aware binaries as part of their startup. To avoid need for locking,
-	// this value should be set once, ideally in an init(), before multiple
-	// goroutines are started.
-	NameValidationScheme = LegacyValidation
+type EscapingScheme int
+
+const (
+	// NoEscaping indicates that a name will not be escaped.
```
beorn7, line 61: Maybe add here that this implies a change in the text based exposition formats if a metric or label name doesn't pass the legacy validation, and that this change is still experimental and still needs to be standardized in future versions of the exposition formats.

ywwg:
squashed

beorn7 (approved):
Thank  you very much.

PR #572, opened 2024-02-02
Add support for secret refs via a secret manager

TheSpiritXIII:
Part 2 for secret provider support within common. See: https://groups.google.com/g/prometheus-developers/c/WKOej_pnhXg

Part 1 is here: https://github.com/prometheus/common/pull/538

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -29,7 +30,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"unsafe"
 
 	conntrack "github.com/mwitkow/go-conntrack"
 	"golang.org/x/net/http/httpproxy"
@@ -131,8 +131,12 @@ func (tv *TLSVersion) String() string {
 type BasicAuth struct {
 	Username     string `yaml:"username" json:"username"`
 	UsernameFile string `yaml:"username_file,omitempty" json:"username_file,omitempty"`
+	// UsernameRef is the name of the secret within the secret manager to use as the username.
+	UsernameRef  string `yaml:"username_ref,omitempty" json:"username_ref,omitempty"`
 	Password     Secret `yaml:"password,omitempty" json:"password,omitempty"`
 	PasswordFile string `yaml:"password_file,omitempty" json:"password_file,omitempty"`
+	// PasswordRef is the name of the secret within the secret manager to use as the password.
+	PasswordRef string `yaml:"password_ref,omitempty" json:"password_ref,omitempty"`
 }
 
 // SetDirectory joins any relative file paths with dir.
@@ -149,6 +153,8 @@ type Authorization struct {
 	Type            string `yaml:"type,omitempty" json:"type,omitempty"`
 	Credentials     Secret `yaml:"credentials,omitempty" json:"credentials,omitempty"`
 	CredentialsFile string `yaml:"credentials_file,omitempty" json:"credentials_file,omitempty"`
+	// CredentialsRef is the name of the secret within the secret manager to use as credentials.
+	CredentialsRef string `yaml:"credentials_ref,omitempty" json:"credentials_ref,omitempty"`
 }
 
 // SetDirectory joins any relative file paths with dir.
@@ -225,14 +231,17 @@ func (u URL) MarshalJSON() ([]byte, error) {
 
 // OAuth2 is the oauth2 client configuration.
 type OAuth2 struct {
-	ClientID         string            `yaml:"client_id" json:"client_id"`
-	ClientSecret     Secret            `yaml:"client_secret" json:"client_secret"`
-	ClientSecretFile string            `yaml:"client_secret_file" json:"client_secret_file"`
-	Scopes           []string          `yaml:"scopes,omitempty" json:"scopes,omitempty"`
-	TokenURL         string            `yaml:"token_url" json:"token_url"`
-	EndpointParams   map[string]string `yaml:"endpoint_params,omitempty" json:"endpoint_params,omitempty"`
-	TLSConfig        TLSConfig         `yaml:"tls_config,omitempty"`
-	ProxyConfig      `yaml:",inline"`
+	ClientID         string `yaml:"client_id" json:"client_id"`
+	ClientSecret     Secret `yaml:"client_secret" json:"client_secret"`
+	ClientSecretFile string `yaml:"client_secret_file" json:"client_secret_file"`
+	// ClientSecretRef is the name of the secret within the secret manager to use as the client
+	// secret.
+	ClientSecretRef string            `yaml:"client_secret_ref" json:"client_secret_ref"`
+	Scopes          []string          `yaml:"scopes,omitempty" json:"scopes,omitempty"`
+	TokenURL        string            `yaml:"token_url" json:"token_url"`
+	EndpointParams  map[string]string `yaml:"endpoint_params,omitempty" json:"endpoint_params,omitempty"`
+	TLSConfig       TLSConfig         `yaml:"tls_config,omitempty"`
+	ProxyConfig     `yaml:",inline"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface
@@ -330,6 +339,18 @@ func (c *HTTPClientConfig) SetDirectory(dir string) {
 	c.BearerTokenFile = JoinDir(dir, c.BearerTokenFile)
 }
 
+// nonZeroCount returns the amount of values that are non-zero.
+func nonZeroCount[T comparable](values ...T) int {
+	count := 0
+	var zero T
+	for _, value := range values {
+		if value != zero {
+			count += 1
+		}
+	}
+	return count
+}
+
 // Validate validates the HTTPClientConfig to check only one of BearerToken,
 // BasicAuth and BearerTokenFile is configured. It also validates that ProxyURL
 // is set if ProxyConnectHeader is set.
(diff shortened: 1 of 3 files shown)


config/http_config.go
```diff
@@ -501,6 +527,13 @@ func NewClientFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HTTPClie
 // given config.HTTPClientConfig and config.HTTPClientOption.
 // The name is used as go-conntrack metric label.
 func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HTTPClientOption) (http.RoundTripper, error) {
+	return NewRoundTripperFromConfigWithContext(context.Background(), cfg, name, optFuncs...)
+}
+
+// NewRoundTripperFromConfig returns a new HTTP RoundTripper configured for the
+// given config.HTTPClientConfig and config.HTTPClientOption.
```
bwplotka, line 543: lmao, what happened that we have both *Config and *Option 🙈

config/http_config.go
```diff
@@ -148,6 +152,7 @@ type Authorization struct {
 	Type            string `yaml:"type,omitempty" json:"type,omitempty"`
 	Credentials     Secret `yaml:"credentials,omitempty" json:"credentials,omitempty"`
 	CredentialsFile string `yaml:"credentials_file,omitempty" json:"credentials_file,omitempty"`
+	CredentialsRef  string `yaml:"credentials_ref,omitempty" json:"credentials_ref,omitempty"`
```
bwplotka, line 155: Would it be useful to add quick commentary what the *Ref field allows (e.g generic secret provider)?

config/http_config.go
```diff
@@ -148,6 +152,7 @@ type Authorization struct {
 	Type            string `yaml:"type,omitempty" json:"type,omitempty"`
 	Credentials     Secret `yaml:"credentials,omitempty" json:"credentials,omitempty"`
```
bwplotka, line 153: not(non blocking): Wild idea, but what stops us to use Credentials file as a ref. Or *File? 

I guess potential confusion and lack of ability for users to use inline somewhere, file somewhere and X secret provider in another place on one deployment? 🤔

config/http_config.go
```diff
@@ -325,6 +331,18 @@ func (c *HTTPClientConfig) SetDirectory(dir string) {
 	c.BearerTokenFile = JoinDir(dir, c.BearerTokenFile)
 }
 
+// nonZeroCount returns the amount of values that are non-zero.
+func nonZeroCount[T comparable](values ...T) int {
+	count := 0
+	var zero T
```
bwplotka, line 341: Nice!

config/http_config.go
```diff
@@ -584,64 +629,131 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
-		ci.CloseIdleConnections()
+func (s *fileSecret) description() string {
+	return fmt.Sprintf("file %s", s.file)
+}
+
+func (s *fileSecret) immutable() bool {
+	return false
+}
+
+type refSecret struct {
```
bwplotka, line 707: Can we add some quick commentary - bit more complex than other two

config/http_config.go
```diff
@@ -584,64 +629,131 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
+
+func (s *refSecret) description() string {
+	return fmt.Sprintf("ref %s", s.ref)
+}
+
+func (s *refSecret) immutable() bool {
+	return false
+}
+
+func secretFrom(secretManager SecretManager, text Secret, file, ref string) (secret, error) {
```
bwplotka, line 724: Can we add commentary about some non trivial semantics e.g.

* what if both text and file are set, what's the behaviour? Does this assumes some pre-validation?

bwplotka, line 724:
```suggestion
func toSecret(secretManager SecretManager, text Secret, file, ref string) (secret, error) {
```
Perhaps clearer?

config/http_config.go
```diff
@@ -584,64 +629,131 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
+		}, nil
+	}
+	return nil, nil
 }
 
-type authorizationCredentialsFileRoundTripper struct {
-	authType            string
-	authCredentialsFile string
-	rt                  http.RoundTripper
+type authorizationCredentialsRoundTripper struct {
```
bwplotka, line 747: Love the unification ❤️

config/http_config.go
```diff
@@ -729,19 +835,20 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
 		}
-		secret = strings.TrimSpace(string(data))
-		rt.mtx.RLock()
-		changed = secret != rt.secret
-		rt.mtx.RUnlock()
-	} else {
-		// Either an inline secret or nothing (use an empty string) was provided.
-		secret = string(rt.config.ClientSecret)
 	}
+	rt.mtx.RLock()
```
bwplotka: should we leverage immutable method to avoid any lock, similar to prev part:

image

bwplotka: Plus we can move it to `if clientSecret != nil {` body?

bwplotka (approved):
Amazing! Generally LGTM, but lots of minor suggestions.

*Disclaimer: @TheSpiritXIII is in my Google team. I tried to get more hands on this review for weeks, but no response. The secret provider feature was generally accepted [1](https://github.com/prometheus/alertmanager/issues/3108), [2](https://github.com/prometheus/prometheus/pull/13955#issuecomment-2090614673), so allowing myself to move forward after the careful review*

config/http_config.go
```diff
@@ -769,9 +848,13 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to use client secret: %w", err)
+	}
+	if clientSecret != nil {
 		var err error
-		secret, err = rt.clientSecret.fetch()
+		secret, err = clientSecret.fetch(req.Context())
 		if err != nil {
 			return nil, fmt.Errorf("unable to read oauth2 client secret: %w", err)
 		}
```
bwplotka, line 860: I guess I missed this in prev PR, but this comment is still relevant https://github.com/prometheus/common/pull/572#discussion_r1595453630

config/http_config.go
TheSpiritXIII: The config is the YAML configuration. We must make the fields public for `prometheus/prometheus` to be able to set the values. However, once you do that, `yaml` will start to unmarshal those fields, which we do not want. Hence why there is a separate object for setting additional configuration outside of the user-specified configuration. This is a limitation of the `yaml` library.

config/http_config.go
TheSpiritXIII: Prometheus uses these objects to load the YAML configurations. Prometheus has no way of knowing whether you actually want a file or a ref. If it somehow knew, you could add the files to the secret manager under a special "file secret provider". If we didn't use these objects for unmarshaling, your idea would work.

config/http_config.go
bwplotka: Makes sense

config/http_config.go
```diff
@@ -594,21 +651,28 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
...
 	if tlsSettings.CA == nil || tlsSettings.CA.immutable() {
 		// No need for a RoundTripper that reloads the CA file automatically.
 		return newRT(tlsConfig)
 	}
-	return NewTLSRoundTripper(tlsConfig, tlsSettings, newRT)
+	return NewTLSRoundTripperWithContext(ctx, tlsConfig, tlsSettings, newRT)
+}
+
+type SecretManager interface {
+	Fetch(ctx context.Context, secretRef string) (string, error)
```
bwplotka, line 670: Missing commentary on experted method and interface, would be nice to add it

config/http_config.go
```diff
@@ -878,8 +978,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
 
+type tlsConfigOptions struct {
+	secretManager SecretManager
+}
+
+// TLSConfigOption defines an option that can be applied to the HTTP client.
+type TLSConfigOption func(options *tlsConfigOptions)
+
+// WithSecretManager allows setting the secret manager.
+func withSecretManager(manager SecretManager) TLSConfigOption {
```
bwplotka, line 988: Shouldn't this be:

```suggestion
func WithSecretManagerTLSConfigOption(manager SecretManager) TLSConfigOption {
```
(for dedup and to be exported)?

config/http_config.go
```diff
@@ -769,16 +862,20 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		changed bool
 	)
 
-	if rt.clientSecret != nil {
+	// Fetch the secret if it's our first run or always if the secret can change.
```
bwplotka, line 864: 👍🏽

config/http_config.go
```diff
@@ -769,16 +861,20 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
+			rt.mtx.RLock()
+			changed = secret != rt.lastSecret
+			rt.mtx.RUnlock()
+		}
 	}
-	rt.mtx.RLock()
-	changed = secret != rt.secret
-	rt.mtx.RUnlock()
 
 	if changed || rt.rt == nil {
```
bwplotka, line 879: Would it be harmful if we kill this if (827 line) and do the changed = ... even on rt.rt == nil case? Bit simpler code and then we can remove the `|| rt.rt == nil` part from if 879 (also simpler). WDYT?

bwplotka (approved):
LGTM, just some nits and one potentially blocking item (did you meant to make TLSConfigOption withSecretManager private?)

config/http_config.go
TheSpiritXIII: Is this what you mean?

config/http_config.go
TheSpiritXIII: We can do this with a common object that implements both interfaces. What do you think of the new solution?

bwplotka: Whatever is simpler here

config/http_config.go
```diff
@@ -769,18 +883,28 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		changed bool
 	)
 
-	if rt.clientSecret != nil {
-		var err error
-		secret, err = rt.clientSecret.fetch()
-		if err != nil {
-			return nil, fmt.Errorf("unable to read oauth2 client secret: %w", err)
+	// Fetch the secret if it's our first run or always if the secret can change.
+	if rt.rt == nil || (rt.clientSecret != nil && !rt.clientSecret.immutable()) {
```
bwplotka, line 887: This not ended up better (or worse) after my suggestions -- might be my misunderstanding, so I ended up prototyping how it might look, proposed in https://github.com/TheSpiritXIII/prometheus-common/pull/1

I also fixed a small race with reading `rt.rt` 🤔I think this race was before too.

bwplotka: No regression and I only refactor in my PR, so let's merge this one and review my PR if there is a need (https://github.com/prometheus/common/blob/58f58f894eb3dd8aafd868824747a997c6fb6e83/config/http_config.go#L759)

bwplotka (approved):
Let's merge! thanks!

PR #576, opened 2024-02-14
Unexport Format strings

ywwg:
With the addition of the escaping term, format strings now have many permutations and doing direct string comparisons of formats is not reliable. Instead, users should call FormatType and compare the result against the possible enum values.

I am making this change after breaking a test in prometheus that relied on string matching of formats.

### DIFF
--- a/expfmt/expfmt.go
+++ b/expfmt/expfmt.go
@@ -34,20 +34,21 @@ const (
 	TextVersion              = "0.0.4"
 	ProtoType                = `application/vnd.google.protobuf`
 	ProtoProtocol            = `io.prometheus.client.MetricFamily`
-	ProtoFmt                 = ProtoType + "; proto=" + ProtoProtocol + ";"
+	protoFmt                 = ProtoType + "; proto=" + ProtoProtocol + ";"
 	OpenMetricsType          = `application/openmetrics-text`
 	OpenMetricsVersion_0_0_1 = "0.0.1"
 	OpenMetricsVersion_1_0_0 = "1.0.0"
 
-	// The Content-Type values for the different wire protocols. Do not do direct
-	// comparisons to these constants, instead use the comparison functions.
-	FmtUnknown           Format = `<unknown>`
-	FmtText              Format = `text/plain; version=` + TextVersion + `; charset=utf-8`
-	FmtProtoDelim        Format = ProtoFmt + ` encoding=delimited`
-	FmtProtoText         Format = ProtoFmt + ` encoding=text`
-	FmtProtoCompact      Format = ProtoFmt + ` encoding=compact-text`
-	FmtOpenMetrics_1_0_0 Format = OpenMetricsType + `; version=` + OpenMetricsVersion_1_0_0 + `; charset=utf-8`
-	FmtOpenMetrics_0_0_1 Format = OpenMetricsType + `; version=` + OpenMetricsVersion_0_0_1 + `; charset=utf-8`
+	// The Content-Type values for the different wire protocols. Note that these
+	// values are now unexported. If code was relying on comparisons to these
+	// constants, instead use FormatType().
+	fmtUnknown           Format = `<unknown>`
+	fmtText              Format = `text/plain; version=` + TextVersion + `; charset=utf-8`
+	fmtProtoDelim        Format = protoFmt + ` encoding=delimited`
+	fmtProtoText         Format = protoFmt + ` encoding=text`
+	fmtProtoCompact      Format = protoFmt + ` encoding=compact-text`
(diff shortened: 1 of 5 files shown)


ywwg:
can you do a new library release with this change?

beorn7:
Ack, will do it ASAP.

beorn7:
Done: https://github.com/prometheus/common/releases/tag/v0.48.0

ywwg:
https://github.com/prometheus/client_golang/pull/1447

ArthurSens:
> This breaks a lot of things for us... is there an example porting guide?

We're currently in the process of cutting a new client_golang release. Client_golang 1.19 should work with common 0.48

expfmt/expfmt.go
```diff
@@ -70,6 +71,26 @@ const (
...
+	switch t {
+	case TypeProtoCompact:
+		return fmtProtoCompact
+	case TypeProtoDelim:
+		return fmtProtoDelim
+	case TypeProtoText:
+		return fmtProtoText
+	case TypeTextPlain:
+		return fmtText
+	case TypeOpenMetrics:
```
mrueg, line 87: Is there a way to generate a fmtOpenMetrics_0_0_1?

slonka: yeah, need the same here https://github.com/kumahq/kuma/pull/9462/files#diff-7f7235c97e7b47eb33457b82a7d6a9ae72052d0fe18249e47c58531a3dd7ef64R37

ywwg: yes, the new functions in expfmt.go can create these for you: https://github.com/prometheus/common/blob/main/expfmt/expfmt.go#L77

The story is, because the exposition formats are getting more complex, simple string comparison is no longer safe. There are too many equivalent permutations, and therefore things were moved to functions.  If the available NewFormat function is not sufficient for your needs, please @ me or file an issue and I can fix it up

mrueg: Right now TypeOpenMetrics only generates fmtOpenMetrics_1_0_0

If for some reason you need fmtOpenMetrics_0_0_1 there's no way to generate it

ywwg: I see -- yup I can prep a PR

ywwg: it looks like I should also increase the test coverage -- I'll get to this next week. sorry for the breakage in the meantime!

ywwg: @mrueg please let me know if this approach works for you

https://github.com/prometheus/common/pull/596

kylelemons:
The expfmt.FmtText constant appears to be used [extensively in Kubernetes](https://sourcegraph.com/search?q=context:global+repo:github.com/kubernetes+expfmt.FmtText&patternType=keyword&sm=0).

Now that intermediate libraries are picking this up via client_golang, it is creating headaches for getting dependency trees to converge without replace directives.

Can we get this (at least partially) reverted?

SuperQ:
@kylelemons I think you may be overestimating the use, most of the results in your search are inline `vendor` inclusions.

For example, https://github.com/kubernetes/ingress-gce/pull/2600, seems like they have a very broken depdenabot setup that tries to do everything all at once.

SuperQ:
I can only find [one use](https://sourcegraph.com/search?q=context:global+repo:github.com/kubernetes+expfmt.FmtText+-file:vendor/github.com/prometheus/client_golang+-file:vendor/k8s.io&patternType=keyword&sm=0).

dnephin:
It's used in many places: https://sourcegraph.com/search?q=context:global++expfmt.FmtText+-file:vendor/github.com/prometheus/client_golang+-file:vendor/k8s.io&patternType=keyword&sm=0

The most common one I've seen is

```
/go/pkg/mod/k8s.io/component-base@v0.30.0/metrics/testutil/metrics.go:73:59: undefined: expfmt.FmtText
```

Maybe this is fixed in main so it's not showing up in sourcegraph? There is no non-alpha release with the fix.

ywwg:
@beorn7 it sounds like we could restore expfmt.FmtText as a single const for compatibility, and if people experience issues with escaping / utf-8 they will have to update their code anyway

nyetwurk:
> @beorn7 it sounds like we could restore expfmt.FmtText as a single const for compatibility, and if people experience issues with escaping / utf-8 they will have to update their code anyway

This is what deprecate is for. 
https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go

https://github.com/search?q=repo%3Agolang%2Fgo+Deprecated%3A+language%3AGo&type=code

SuperQ:
Agreed, this should have been marked deprecated first.

PR #591, opened 2024-02-29
Remove deprecated version function

SuperQ:
Remove deprecated version.NewCollector function. This has been moved to `github.com/prometheus/client_golang`. This fixes the circular dependency between the two repos.


ArthurSens (approved):
🥳

SuperQ:
I'm thinking about cutting v0.50.0 with just this change, so it's only a small step for updating client_golang to break the dependency loop.

bboreham (approved):
Can you link in the PR to where the function was added that we should use instead?

PR #597, opened 2024-03-06
Remove unused 'Host' member from HTTPClientConfig

bboreham:
Doesn't seem to be used, and it breaks Prometheus Docker SD which also has a `host` field defined in Yaml.

Added in #549


beorn7:
As you are trying in https://github.com/prometheus/prometheus/pull/13717 ?

bboreham:
Please explain why that is better.

ArthurSens:
Sorry for spreading comments on all communication channels 😅 Posting here what I posted in slack:

If I understood correctly https://github.com/prometheus/common/pull/549 allows users to roundTrip the HTTP request with a new host...? So it is used

The test added to that PR shows how host is supposed to be overriden

SuperQ (approved):

PR #608, opened 2024-03-21
Correct logic in sample naming for counters, add new test

### LINKED ISSUE #605: OpenMetrics: _total suffix repeated multiple times
### Description

Today we faced the problem with counter export in openmetrics format: counter names keep being appended by "_total" suffices, one time for every new labelset.

This is how it looks in grafana metric browser after being collected by prometheus

Screenshot

### Debug

The problem seems to be around the line
https://github.com/vesari/common/blob/6a86956d41af916f0c048f4988d93994bb865b9e/expfmt/openmetrics_create.go#L255
which modifies metric name inside the loop,

and was introduced in [this commit](https://github.com/prometheus/common/pull/544/commits/6a86956d41af916f0c048f4988d93994bb865b9e).

### Affected versions

github.com/prometheus/common v0.50.0

vesari:
This PR solves the bug causing the `_total` suffix to be repeated multiple times in the sample metric name, and adds a new corresponding test case. Fixes #605.

### DIFF
--- a/expfmt/openmetrics_create.go
+++ b/expfmt/openmetrics_create.go
@@ -248,12 +248,12 @@ func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...E
 	var createdTsBytesWritten int
 
 	// Finally the samples, one line for each.
+	if metricType == dto.MetricType_COUNTER && strings.HasSuffix(name, "_total") {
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):
Thanks for solving this so quickly!

vesari:
Thank you @ArthurSens as always

PR #610, opened 2024-03-23
Retract v0.50.0

SuperQ:
Mark the v0.50.0 tag as retracted in the Go modules due to the critical nature of the issue.


ArthurSens (approved):
Thanks!!

PR #612, opened 2024-03-24
Drop support for Go older than 1.18

SuperQ:
We don't support Go older than 1.21 in the Go modules file, so we can drop the special handling for older Go in the version package.


ArthurSens (commented):
Sorry if the comment sounds lazy, I'm still stuck between airports and not very familiar with runtime libraries.

version/info.go
```diff
@@ -74,3 +78,48 @@ func Info() string {
...
+			rev = v.Value
+		}
+		if v.Key == "vcs.modified" {
+			if v.Value == "true" {
+				modified = true
+			}
+		}
+		if v.Key == "-tags" {
+			tags = v.Value
+		}
```
ArthurSens, line 119: I'm reading the BuildSettings docs and I can't find the -tags keys.

Is this something that was removed in more recent versions of go? Or just missing in the docs?

```go
// A BuildSetting is a key-value pair describing one setting that influenced a build.
//
// Defined keys include:
//
//   - -buildmode: the buildmode flag used (typically "exe")
//   - -compiler: the compiler toolchain flag used (typically "gc")
//   - CGO_ENABLED: the effective CGO_ENABLED environment variable
//   - CGO_CFLAGS: the effective CGO_CFLAGS environment variable
//   - CGO_CPPFLAGS: the effective CGO_CPPFLAGS environment variable
//   - CGO_CXXFLAGS:  the effective CGO_CXXFLAGS environment variable
//   - CGO_LDFLAGS: the effective CGO_LDFLAGS environment variable
//   - GOARCH: the architecture target
//   - GOAMD64/GOARM/GO386/etc: the architecture feature level for GOARCH
//   - GOOS: the operating system target
//   - vcs: the version control system for the source tree where the build ran
//   - vcs.revision: the revision identifier for the current commit or checkout
//   - vcs.time: the modification time associated with vcs.revision, in RFC3339 format
//   - vcs.modified: true or false indicating whether the source tree had local modifications
type BuildSetting struct {
	// Key and Value describe the build setting.
	// Key must not contain an equals sign, space, tab, or newline.
	// Value must not contain newlines ('\n').
	Key, Value string
}
```

SuperQ: Hmm, good question. This isn't new code, just moving it from `info_go118.go`. @bboreham added it in https://github.com/prometheus/common/pull/455.

bboreham: The list is not claimed to be exclusive; it says it “includes” those things.

ArthurSens (approved):

PR #567, opened 2024-01-12
switch to protodelim package (which pbutil now calls)

stapelberg:
Go Protobuf v1.30.0 includes the new protodelim package.

github.com/matttproud/golang_protobuf_extensions/v2/pbutil is, at this point, just a wrapper around the protodelim package:

https://github.com/matttproud/golang_protobuf_extensions/commit/7c0096c86977a6345402445d1048134557c72820

https://matttproud.com/blog/posts/retiring-pbutil.html

Therefore, this change calls protodelim directly,
allowing us to get rid of one module dependency :)

cc @matttproud

PR #616, opened 2024-04-03
fix(protobuf): Correctly decode multi-messages streams

srebhan:
When sending multiple metric-family samples in one stream the current implementation only decodes the _first_ of those messages, all other ones are lost. This is the case since #567 which switches to directly using the `protodelim` package. However, in the process or copying the code, the initialization of the `reader` moved to the `Decode()` function and as such, the buffered reader continues _somewhere_ in the actual data-stream (usually at the end) leading to an `EOF` on the next `Decode()` call.

The malfunction is present since release v0.46.0.

This PR moves the initialization out of `Decode()`  into the initialization of the decoder preventing the described issue. Furthermore, we add a unit-test including a multi-message data-sample verifying the issue is fixed.


beorn7 (approved):
Thanks for the catch. I guess this makes sense.

@stapelberg FYI (and just in case you want to raise any concerns about this fix)

stapelberg:
Oh, I see the mistake! Sorry about that, and thanks for sending a fix which even increases test coverage :)

PR #618, opened 2024-04-12
Add StatusAt method for Alert struct

grobinson-grafana:
This commit adds the StatusAt method for the Alert struct. It calls the ResolvedAt method while Status calls the Resolved method.

This method will be used in Alertmanager to fix issue prometheus/alertmanager#3351.


roidelapluie:
Thanks!

PR #619, opened 2024-04-13
LabelSet.String: restore faster sort call

bboreham:
This was changed without explanation at #617; let's put it back.

```
goos: darwin
goarch: arm64
                       │ before.txt  │             after.txt              │
                       │   sec/op    │   sec/op     vs base               │
LabelSetStringMethod-8   500.4n ± 0%   434.4n ± 0%  -13.19% (p=0.002 n=6)

                       │ before.txt  │             after.txt             │
                       │    B/op     │    B/op     vs base               │
LabelSetStringMethod-8   616.00 ± 0%   80.00 ± 0%  -87.01% (p=0.002 n=6)

                       │ before.txt │             after.txt             │
                       │ allocs/op  │ allocs/op   vs base               │
LabelSetStringMethod-8   3.000 ± 0%   1.000 ± 0%  -66.67% (p=0.002 n=6)
```

### DIFF
--- a/model/labelset_string.go
+++ b/model/labelset_string.go
@@ -17,7 +17,7 @@ package model
 
 import (
 	"bytes"
-	"sort"
+	"slices"
 	"strconv"
 )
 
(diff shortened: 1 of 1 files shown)


ArthurSens (approved):
Thanks for also taking care of the fix!

SuperQ (approved):

PR #625, opened 2024-04-19
Move goautoneg to external dependency

mikelolasagasti:
`goautoneg` module was imported to the repo as it was hosted in bitbucket and required Mercurial at the time (2015).

Bitbucket repository is long gone and a GitHub fork is available under `munnerz/goautoneg`. This fork is used by other projects like Kubernetes.


SuperQ (approved):

gotjosh (approved):
LGTM, verified is the same dependency. https://github.com/kubernetes/kubernetes/blob/master/go.mod#L46C1-L47C1

yurishkuro:
goautoneg uses an unusual license by Open Knowledge Foundation Ltd, which is not on the CNCF list https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md. Why is it acceptable as a dependency?

bboreham:
https://github.com/munnerz/goautoneg/blob/master/LICENSE seems to me identical to https://opensource.org/license/BSD-3-Clause, with the addition of the line "All rights reserved."  Is that what you are referring to as unusual ?

yurishkuro:
GitHub would normally recognize BSD3 license. This one is just a bit different, but yeah, it looks BSD3.

machine424:
FYI https://github.com/munnerz/goautoneg/issues/7
(by the way https://github.com/elnormous/contenttype seems to be better tested for me.)

PR #634, opened 2024-05-15
oauth2RoundTripper: Avoid race condition and readability changes.

bwplotka:
After @TheSpiritXIII work we noticed the `rt.rt == nil` was always shared across threads potentially but not locked. This PR fixes that and adjust the flow for readability.

PTAL @TheSpiritXIII 

### Changes

* Avoid race condidtion on rt.rt == nil check
* Trying to improve readability (less ifs)
* Some comment fixes

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -263,12 +263,12 @@ func (o *OAuth2) UnmarshalJSON(data []byte) error {
 }
 
 // SetDirectory joins any relative file paths with dir.
-func (a *OAuth2) SetDirectory(dir string) {
-	if a == nil {
+func (o *OAuth2) SetDirectory(dir string) {
+	if o == nil {
 		return
 	}
-	a.ClientSecretFile = JoinDir(dir, a.ClientSecretFile)
-	a.TLSConfig.SetDirectory(dir)
+	o.ClientSecretFile = JoinDir(dir, o.ClientSecretFile)
+	o.TLSConfig.SetDirectory(dir)
 }
 
 // LoadHTTPConfig parses the YAML input s into a HTTPClientConfig.
@@ -563,7 +563,7 @@ func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, optFuncs ...HT
 	return NewRoundTripperFromConfigWithContext(context.Background(), cfg, name, optFuncs...)
 }
 
-// NewRoundTripperFromConfig returns a new HTTP RoundTripper configured for the
+// NewRoundTripperFromConfigWithContext returns a new HTTP RoundTripper configured for the
 // given config.HTTPClientConfig and config.HTTPClientOption.
 // The name is used as go-conntrack metric label.
 func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientConfig, name string, optFuncs ...HTTPClientOption) (http.RoundTripper, error) {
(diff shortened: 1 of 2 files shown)


config/http_config.go
```diff
@@ -766,9 +765,6 @@ func toSecret(secretManager SecretManager, text Secret, file, ref string) (secre
 		}, nil
 	}
 	if ref != "" {
-		if secretManager == nil {
```
TheSpiritXIII, line 769: The problem with removing this is that you have to always call `WithSecretManager` or else this would crash, so this it not completely backwards compatible.

config/http_config.go
```diff
@@ -858,117 +854,120 @@ func (rt *basicAuthRoundTripper) CloseIdleConnections() {
...
 	clientSecret secret
-	lastSecret   string
-	mtx          sync.RWMutex
 	opts         *httpClientOptions
 	client       *http.Client
 }
 
 func NewOAuth2RoundTripper(clientSecret secret, config *OAuth2, next http.RoundTripper, opts *httpClientOptions) http.RoundTripper {
+	if clientSecret == nil {
+		// TODO(bwplotka): Consider removing support for Oauth2 with just clientID and no secret,
```
TheSpiritXIII, line 870: Note: This is allowed in the OAuth 2 specification although rare. I would remove this comment.

config/http_config_test.go
```diff
@@ -507,46 +507,48 @@ func TestNewClientFromConfig(t *testing.T) {
 	}
 
 	for _, validConfig := range newClientValidConfig {
-		testServer, err := newTestServer(validConfig.handler)
-		if err != nil {
-			t.Fatal(err.Error())
-		}
-		defer testServer.Close()
+		t.Run("", func(t *testing.T) {
```
TheSpiritXIII, line 510: Suggestion: How does this work when multiple tests have the same name? Can you add a description here? The index in the for loop would be good enough here.

config/http_config.go
```diff
@@ -858,117 +854,120 @@ func (rt *basicAuthRoundTripper) CloseIdleConnections() {
 }
 
 type oauth2RoundTripper struct {
+	mtx        sync.RWMutex
+	next       *oauth2.Transport
```
TheSpiritXIII, line 858: Nit (non-blocking): I'm unsure whether I like this name. It's the previous/last transport, not the next one.

TheSpiritXIII (changes requested):
Greatly appreciated! I left a few comments.

config/http_config.go
bwplotka: ayaya, I wanted to remove it but then wanted to get it back, sorry. Ignore this.

bwplotka: Fixed.

config/http_config.go
bwplotka: In one perspective you are right.. it's not "future". But it is a transport to call next after this round trip invocation. We use that e.g. in `NewOAuth2RoundTripper` 🙃 I think I would agree and change to `lastT` or `lastRT`? 🤔

config/http_config.go
bwplotka: Right, I wanted to challenge this, but I can see we do it in other places too, let's kill. I wonder if there is any abstraction which would unify all those roundtrippers to ensure some consistency across them, will check later.

config/http_config_test.go
bwplotka: Empty name generates index (autocount), so I think this should be fine!

bwplotka:
Addressed, thanks! Perhaps there is still room to build some code to unify those rountrippers, maybe put them in separate file etc, but good enough for now 💪🏽

TheSpiritXIII (approved):
Looks great. Thanks!

PR #645, opened 2024-06-03
Expose HTTPHost property in HTTPClientConfig

jkroepke:
While I was integrate that feature in https://github.com/grafana/alloy/pull/698, it figure out that the Host property was removed in https://github.com/prometheus/common/pull/597. Turns out that I forget something.

Since host is problematic with DockerSD, i choice the term `http_host` to follow the pattern from `http_headers`


SuperQ (approved):

config/http_config.go
```diff
@@ -319,6 +319,9 @@ type HTTPClientConfig struct {
 	// The omitempty flag is not set, because it would be hidden from the
 	// marshalled configuration when set to false.
 	EnableHTTP2 bool `yaml:"enable_http2" json:"enable_http2"`
+	// HTTPHost optionally overrides the Host header to send.
+	// If empty, the host from the URL is used.
```
bboreham, line 323: How does this reconcile with the code that says:
```
rt = NewHostRoundTripper(opts.host, rt)
```

bboreham (commented):
> Turns out that I forget something.

Please explain this further.

bboreham:
In #549 you added `Host` to `HTTPClientConfig` and `host` to `httpClientOptions`.

Since only the latter was used, I removed the former.  Now you adding it back, and hopefully using it.
Do we need both?  Why?

jkroepke:
Hey @bboreham,

I only need the property in HTTPClientConfig within the context of Grafana Alloy.

I have to apologize. When I started with Golang a year ago, I implemented the wrong side due to my lack of knowledge and missing tooling. I just replicated the userAgent logic, which misled me.

Looking at it now, I realize that the host property in httpClientOptions is incorrect.

If you agree, I can revert the host in httpClientOptions.

bboreham:
Thanks for the explanation. It's fine, everybody makes mistakes.  Just easier for me to follow if I can see the path.

However we may prefer to deprecate the un-needed one rather than removing it immediately.
That way we can see if anyone else came to depend on it.

config/http_config.go
jkroepke: > If empty, the host from the URL is used.

This is go default behavior. It's similar the the TLSServerName property.

If the property is not set by caller, then the value will be used from URL.

Ref: https://github.com/golang/go/blob/959b3fd4265d7e4efb18af454cd18799ed70b8fe/src/net/http/request.go#L240-L243

bboreham: Would it not be more correct to mention `opts.host` ?

KarstenSiemer:
Hi! Anything missing here?

SuperQ:
Ping @jkroepke @bboreham, any issues to resolve here?

config/http_config.go
jkroepke, line 323:
```suggestion
	// Can be overridden by opts.host.
```

jkroepke:
@SuperQ from my understand, the review asks for a more specific doc note which I applied now.

PR #648, opened 2024-06-06
feat(dependency): add diff package to common

dongjiang1989:
Add diff package to common for client-golang, prometheus and alertmanager.

issue: https://github.com/prometheus/client_golang/issues/1490
pr for client-glang: https://github.com/prometheus/client_golang/pull/1499

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -3,3 +3,8 @@ Copyright 2015 The Prometheus Authors
 
 This product includes software developed at
 SoundCloud Ltd. (http://soundcloud.com/).
+
+diff - a pretty-printed complete of a Go data structure
+https://github.com/kylelemons/godebug
+Copyright 2013 Google Inc.  All rights reserved.
+See source code for license details.
--- a/helpers/diff/diff.go
+++ b/helpers/diff/diff.go
@@ -1,0 +1,193 @@
+// Copyright 2013 Google Inc.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


SuperQ:
Wait, no, we should not be copy-pasting code like this. 

Fixing an EoL package by replacing it with something not EoL is just fine. There's no need to fully remove a dependency.

CC @kakkoyun

kakkoyun:
> Wait, no, we should not be copy-pasting code like this.
> 
> Fixing an EoL package by replacing it with something not EoL is just fine. There's no need to fully remove a dependency.
> 
> CC @kakkoyun

We did this in `client_golang` to reduce the external dependencies. And it is small enough and only used int the tests. I never thought it would be needed elsewhere.

That being said, I'd be more than happy to accept https://github.com/prometheus/client_golang/pull/1539.

SuperQ:
That was incorrectly done in client_golang. The request was to replace the dependency that was abandoned, not reduce it.

kakkoyun:
> That was incorrectly done in client_golang. The request was to replace the dependency that was abandoned, not reduce it.

I'm well aware of the request. We wanted to kill two birds with one stone.

ArthurSens:
> Add diff package to common for client-golang, prometheus and alertmanager.

I'm not sure if I'm missing something, but does prometheus and alertmanager really use this same logic? I'm aware that client_golang uses this for testing, but maybe we could avoid adding it to common if it's only for client_golang.

Now that https://github.com/prometheus/client_golang/pull/1539 was accepted it's even less important to have this here from my point of view 🤔

SuperQ:
IMO, we don't want to get into the habit of in-lining Go dependencies. This sets a hard to deal with policy long-term. What is the limit for an inline? How do we keep inline code up-to-date?

PR #649, opened 2024-06-06
Add SigV4 FIPS STS  endpoint config

rajagopalanand:
This PR is for [#13364](https://github.com/prometheus/prometheus/issues/13364)

### DIFF
--- a/sigv4/sigv4.go
+++ b/sigv4/sigv4.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/endpoints"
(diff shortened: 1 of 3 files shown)


rapphil:
So this module is not using the default credentials chain and instead is relying on configurations? https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html

ArthurSens (approved):
I'm approving since I see approval from AWS folks :)

I've also run the tests with the changes made in `testdata/sigv4_good.yaml` just to make sure the change is backwards compatible, and tests passed

PR #650, opened 2024-06-07
Expose secret as SecretReader and InlineSecret from config package

pracucci:
https://github.com/prometheus/common/pull/572 broke the vendoring in Mimir, where we have an use case calling `NewBasicAuthRoundTripper()`. The problem is that the `secret` interface is private, but it's used in publicly exposed functions like `NewBasicAuthRoundTripper()`.

In this PR I propose to expose `secret` as `SecretReader` (we already have `type Secret string`) and to also expose `InlineSecret` given passing a const secret is a common use case.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -738,7 +738,7 @@ func (s *fileSecret) Immutable() bool {
 	return false
 }
 
-// refSecret fetches a single secret from a SecretReader manager.
+// refSecret fetches a single secret from a SecretManager.
 type refSecret struct {
 	ref     string
 	manager SecretManager // manager is expected to be not nil.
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

I checked and we don't have `SecretReader` nor `InlineSecret` in Prometheus so the names LGTM.

gotjosh:
@SuperQ @ArthurSens any objections?

gotjosh:
I'm going to proceed with the merge as exposing things public is most of the time uncontroversial, but do let me know if there's anything that needs addressing. I've checked Prometheus and the Alertmanger - and these aren't conflicting.

alanprot:
Nice! this was also breaking Thanos: https://github.com/prometheus/common/pull/538#issuecomment-2150596862

alanprot:
Actually thanos also need the FileSecret to be exposed: =/ 

https://github.com/prometheus/common/pull/653

bwplotka:
Thanks for fixing!

PR #416, opened 2022-11-28
Add HTTP headers support to common HTTP client.

roidelapluie:
This is named `http_headers` so it does not clash with blackbox exporter's headers and Prometheus remote client's headers, which are simple maps.

Signed-off-by: Julien Pivotto <roidelapluie@o11y.eu>

PR #651, opened 2024-06-07
Fix HTTPClientConfig JSON marshalling

pracucci:
I think https://github.com/prometheus/common/pull/416 introduced an issue. `json` marshalling doesn't support `inline` and so when `HTTPClientConfig` gets marshalled to JSON it gets marshalled to something like this:

```
{"http_headers":{"Headers":{"Header-Name":{}}}
```

while I would expect something like this:

```
{"http_headers":{"Header-Name":{}}
```

This PR should fix it.

### DIFF
--- a/config/headers.go
+++ b/config/headers.go
@@ -17,6 +17,7 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


roidelapluie (approved):

roidelapluie:
Thanks!

PR #653, opened 2024-06-07
Expose secret as FileSecret from config package

alanprot:
Similar to https://github.com/prometheus/common/pull/650

Thanos also need the FileSecret to be exposed:

See: https://github.com/thanos-io/thanos/blob/c08dc141dd671f9c6ecc6dd85a9f39f8be32419c/pkg/clientconfig/http.go#L167

and https://github.com/prometheus/common/pull/538#issuecomment-2150596862

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -718,11 +718,15 @@ func (s *InlineSecret) Immutable() bool {
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

PR #654, opened 2024-06-10
chore: add HumanizeTimestamp; make ConvertToFloat exportable

freak12techno:
See the discussion here: https://github.com/prometheus/prometheus/pull/14202

Basically copied HumanizeTimestamp from prometheus repo to move it there and added some tests (basically refactored to avoid using templating), the idea is to remove its implementation and the implementation of ConvertToFloat (which are both used in alertmanager and prometheus repos) from both prometheus and alertmanager repos to include the implementation from this repo.

@gotjosh can you check?

### DIFF
--- a/helpers/templates/time.go
+++ b/helpers/templates/time.go
@@ -14,13 +14,18 @@
 package templates
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
 	"time"
+
+	"github.com/prometheus/common/model"
 )
 
-func convertToFloat(i interface{}) (float64, error) {
+var errNaNOrInf = errors.New("value is NaN or Inf")
+
+func ConvertToFloat(i interface{}) (float64, error) {
 	switch v := i.(type) {
 	case float64:
 		return v, nil
@@ -41,8 +46,20 @@ func convertToFloat(i interface{}) (float64, error) {
 	}
 }
 
+func FloatToTime(v float64) (*time.Time, error) {
+	if math.IsNaN(v) || math.IsInf(v, 0) {
+		return nil, errNaNOrInf
+	}
+	timestamp := v * 1e9
+	if timestamp > math.MaxInt64 || timestamp < math.MinInt64 {
+		return nil, fmt.Errorf("%v cannot be represented as a nanoseconds timestamp since it overflows int64", v)
+	}
+	t := model.TimeFromUnixNano(int64(timestamp)).Time().UTC()
+	return &t, nil
+}
+
 func HumanizeDuration(i interface{}) (string, error) {
-	v, err := convertToFloat(i)
+	v, err := ConvertToFloat(i)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


freak12techno:
@machine424 since you reviewed the previous PR on Prometheus repo, maybe you can take a look?

helpers/templates/time_test.go
```diff
@@ -139,3 +140,103 @@ func TestHumanizeDurationSecondsInt(t *testing.T) {
 		})
 	}
 }
+
+func TestHumanizeTimestampInt(t *testing.T) {
+	tc := []struct {
+		name     string
+		input    int
```
ArthurSens, line 147: I've noticed that this part here is equal on all new tests:
```go
			result, err := HumanizeTimestamp(tt.input)
			require.NoError(t, err)
			require.Equal(t, tt.expected, result)
```

If we change the input to `interface`
```suggestion
		input    interface{}
```

Could we merge all those tests into a single one?

helpers/templates/time.go
```diff
@@ -41,8 +46,20 @@ func convertToFloat(i interface{}) (float64, error) {
 	}
 }
 
+func FloatToTime(v float64) (*time.Time, error) {
```
ArthurSens, line 49: In prometheus/prometheus this is unexported. I'm probably lacking context, but any reason to export it here?

ArthurSens (commented):
Sorry for the delayed review, generally it looks good already!

I'm just making a suggestion to reduce a few lines of code in tests

helpers/templates/time.go
freak12techno: My reasoning was the same as ConvertToFloat: we can remove it from prometheus repo and reuse this implementation in both alertmanager and prometheus repos

helpers/templates/time_test.go
freak12techno: Makes sense, will try doing it a bit later

freak12techno: @ArthurSens done

helpers/templates/time_test.go
```diff
@@ -119,23 +73,60 @@ func TestHumanizeDurationSubsecondAndFractionalSecondsString(t *testing.T) {
...
+	result, err := HumanizeTimestamp(1435065584.128)
+	require.NoError(t, err)
+	require.Equal(t, "2015-06-23 13:19:44.128 +0000 UTC", result)
+}
+
+func TestHumanizeTimestampSampleString(t *testing.T) {
+	result, err := HumanizeTimestamp(1435065584.128)
+	require.NoError(t, err)
+	require.Equal(t, "2015-06-23 13:19:44.128 +0000 UTC", result)
+}
```
ArthurSens, line 132: This could also be a single test, but that's just nitpicking :P

ArthurSens (approved):
I'll wait a bit for josh to review because I don't know how Alertmanager plans to use this (again due to my lack of context 😅)

But if it takes too long I'll just merge it, no big deal exporting FloatToTime :)

helpers/templates/time.go
freak12techno: yeah it's also used in `toTime`:
```
			"toTime": func(i interface{}) (*time.Time, error) {
				v, err := convertToFloat(i)
				if err != nil {
					return nil, err
				}

				return floatToTime(v)
			},
```

that way (by exporting it) I can just remove the implementation from the prometheus repo and use the common one

helpers/templates/time_test.go
freak12techno: Actually yeah, can fix right away

helpers/templates/time.go
ArthurSens: perfect

helpers/templates/time_test.go
freak12techno: Also done, should be better now. Thanks for suggesting!
Also found a bug in one of these tests (float64 instead of string), fixed it there as well

gotjosh:
You can merge without me @ArthurSens - if this looks go to you, go right ahead.

freak12techno:
@gotjosh @ArthurSens can either of you bump a version? I'd like to make another PR on Prometheus repo utilizing this and it needs a new version.

Also, appreciate the review.

ArthurSens:
Sure :) 

Let me ping a few people and see if I can proceed with a release!

PR #662, opened 2024-07-14
remove dependency to github.com/prometheus/client_golang

ilius:
Remove dependency to `github.com/prometheus/client_golang`

`client_golang` depends on this module, so it doesn't make sense for this module to depend on `client_golang`.
This can cause `go.sum` file of user repos to be very very large and make it very slow to download deps and build.
Also makes it much harder to fork these two modules.
It's just a design problem.

### DIFF
--- a/sigv4/go.mod
+++ b/sigv4/go.mod
@@ -6,7 +6,6 @@ replace github.com/prometheus/common => ../
 
 require (
 	github.com/aws/aws-sdk-go v1.54.7
-	github.com/prometheus/client_golang v1.19.1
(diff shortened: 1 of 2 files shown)


ilius:
`go mod tidy` adds it back to `go.mod` again.
Perhaps merging this PR allows it to be removed.

Nevertheless this PR helps avoid dependency hell.

SuperQ (approved):
Thanks, yes, we've been trying to break the circular dependency between common and client_golang.

ilius:
I think we need a new tag for commons to be able to remove client_golang.

PR #665, opened 2024-07-15
Revert #576 and add deprecation notice

SuperQ:
Revert #576 that un-exported format strings and replace with a deprecation notice.


kylelemons (approved):

kylelemons:
Confirmed that `go get github.com/prometheus/common@c3e8247156eb95fd21b0d2e5bee633f39ec0fb82` alleviates the k8s dependency issues for my use-case.

SuperQ:
I recommend a full revert so any downstream use can get the deprecation notice.

beorn7 (approved):

beorn7:
If we go down the deprecation path in a 0.x library at all, then let's be as conservative and do it for all the constants.

ArthurSens:
> If we go down the deprecation path in a 0.x library at all, then let's be as conservative and do it for all the constants.

We might want to discuss common 1.0 in a dev summit

ArthurSens (approved):

zarvd:
Hi, do you know when the next release tag with this PR is coming out?

SuperQ:
We should have a new release today or tomorrow. We're finishing up a couple other changes.

PR #666, opened 2024-07-15
Re-export FmtText as a deprecated constant.

ywwg:
Based on the conversation in https://github.com/prometheus/common/pull/576, removing FmtText as an exported constant breaks a lot of existing code.  This change re-exports that value, though marks it as deprecated.


beorn7:
Thanks for doing this. But as discussed over in #665, let's do the full revert.

Issue #554, opened 2023-12-14
Update expfmt/text_parse.go to support the new UTF-8 syntax


ywwg:
text_parse is a go implementation of a parser for the plain prometheus text format

jmichalek132:
I spoke with @ywwg and I would be willing to help with this one over the holidays, could I get this one assigned to me please?

jmichalek132:
Notes from the chat:

the basic changes are:
* current code assumes there must be a metric name before "{". Now, it is ok if a line begins with "{"
* currently, terms inside {} must have an operator, like = or =~. Now, if there is a quoted term without an operator, that is the * metric name. There can only be one term without an operator.
* the left hand side of operators inside {} can be quoted
* for HELP and TYPE and UNIT lines, the metric name may be in quotes.
* best way to attack this is probably creating the test cases and then fixing the code so they all pass

bwplotka:
👍🏽  Thanks!

ywwg:
@jmichalek132 checking in, any progress on this?

jmichalek132:
> @jmichalek132 checking in, any progress on this?

sorry unfortunately not I was sick over the holidays, but I'll take a stab at this over the weekend.

jmichalek132:
So finally looking into this today, first thing I am unclear on based on the proposal is what should be the behaviour with the metric name in the prometheus metadata.
I assume if the metric name is valid it should keep working as is.
If it has utf-8 characters it has to be quoted around in the metadata.
Is this assumption correct @ywwg ?

jmichalek132:
Also when the metric name is surrounded by the double quote I assume we need to validate that if the metric name contains double qoute it's properly escaped right? Otherwise the parsing could break.

ywwg:
In the other parsers, I changed them so they can just read the new format without a flag. It was not practical to create switching logic without duplicating all the parsers. Let's start by not using a flag.

> I assume if the metric name is valid it should keep working as is.
> If it has utf-8 characters it has to be quoted around in the metadata.

Yes that's correct.

The current parser supports escaping already, so it should be fine to use that escaping inside double quotes for the TYPE and HELP lines:

```
# HELP "my.\"noncompliant\".metric" help text
# TYPE "my.\"noncompliant\".metric" counter
```

ywwg:
Checking in on this again, is there a PR attached?

jmichalek132:
[...]

PR #670, opened 2024-07-31
expfmt: Add UTF-8 syntax support in text_parse.go

fedetorres93:
This PR adds support for the new UTF-8 syntax proposed [here](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md). 

Main changes:

- Metric names may include UTF-8 characters and, if so, they will be quoted and inside brackets, without an operator (for example: `{"my.metric", label="value"} 1`)
- Label names may include UTF-8 characters and, if so, they will be quoted (for example: `{"my.metric", "my.label"="value"} 1`)

Addresses https://github.com/prometheus/common/issues/554

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -665,7 +665,6 @@ func (p *TextParser) readTokenAsMetricName() {
 			case 'n':
 				p.currentToken.WriteByte('\n')
 			case '"':
-				p.currentToken.WriteByte('\\')
 				p.currentToken.WriteByte('"')
 			default:
 				p.parseError(fmt.Sprintf("invalid escape sequence '\\%c'", p.currentByte))
@@ -716,7 +715,6 @@ func (p *TextParser) readTokenAsLabelName() {
 			case 'n':
 				p.currentToken.WriteByte('\n')
 			case '"':
-				p.currentToken.WriteByte('\\')
 				p.currentToken.WriteByte('"')
 			default:
 				p.parseError(fmt.Sprintf("invalid escape sequence '\\%c'", p.currentByte))
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -18,16 +18,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/model"
+	"google.golang.org/protobuf/proto"
 	"io"
 	"math"
 	"strconv"
 	"strings"
-
-	dto "github.com/prometheus/client_model/go"
-
-	"google.golang.org/protobuf/proto"
-
-	"github.com/prometheus/common/model"
+	"unicode/utf8"
 )
 
 // A stateFn is a function that represents a state in a state machine. By
@@ -74,6 +72,7 @@ type TextParser struct {
 	// count and sum of that summary/histogram.
 	currentIsSummaryCount, currentIsSummarySum     bool
 	currentIsHistogramCount, currentIsHistogramSum bool
+	currentMetricIsInsideBraces                    bool
 }
 
 // TextToMetricFamilies reads 'in' as the simple and flat text-based exchange
@@ -143,6 +142,7 @@ func (p *TextParser) reset(in io.Reader) {
 // start of a line (or whitespace leading up to it).
 func (p *TextParser) startOfLine() stateFn {
 	p.lineCount++
+	p.currentMetricIsInsideBraces = false
 	if p.skipBlankTab(); p.err != nil {
 		// This is the only place that we expect to see io.EOF,
 		// which is not an error but the signal that we are done.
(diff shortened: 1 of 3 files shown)


expfmt/text_parse.go
```diff
@@ -287,6 +296,35 @@ func (p *TextParser) startLabelName() stateFn {
...
+				p.currentLabel = nil
+				if p.skipBlankTab(); p.err != nil {
+					return nil // Unexpected end of input.
+				}
+				return p.readingValue
+			default:
+				p.parseError(fmt.Sprintf("unexpected end of metric value %q", p.currentByte))
+				return nil
+			}
+		} else {
```
beorn7, line 322: All code paths in the `if` clause above end up returning, so we don't need the `else` here.

expfmt/text_parse.go
```diff
@@ -60,6 +60,7 @@ type TextParser struct {
 	currentMF            *dto.MetricFamily
 	currentMetric        *dto.Metric
 	currentLabelPair     *dto.LabelPair
+	currentLabel         []*dto.LabelPair
```
beorn7, line 63: I know that protobuf uses the singular even for repeated messages. But maybe it would still make more sense and be more readable to use the plural here? `currentLabels` collides with the map of the same name below, but maybe call this `currentLabelPairs`? Also, is there a good comment to add here to explain what this variable is for? (cf. the comment for the `currentLabels` variable below).

expfmt/text_parse_test.go
```diff
@@ -385,6 +385,159 @@ request_duration_microseconds_count 2693
 				},
 			},
 		},
+		// 5: UTF-8 counter
```
beorn7, line 388: Nit: This should end with a period to be consistent with the naming above.

More substantially: Let's be more explicit here about what's significant. It's not special that this is a counter, and "UTF-8" per se doesn't say anything (the old metric names were all valid UTF-8, too). How about "Quoted metric name and quoted label name with dots."?

expfmt/text_parse_test.go
```diff
@@ -385,6 +385,159 @@ request_duration_microseconds_count 2693
...
+							},
+							Counter: &dto.Counter{
+								Value: proto.Float64(1),
+							},
+						},
+					},
+				},
+			},
+		},
+		// 6: Dots in name
```
beorn7, line 416: Suggestion: "Metric family with dots in name."

expfmt/text_parse_test.go
```diff
@@ -385,6 +385,159 @@ request_duration_microseconds_count 2693
...
+							Counter: &dto.Counter{
+								Value: proto.Float64(.23),
+							},
+							TimestampMs: proto.Int64(1234567890),
+						},
+					},
+				},
+			},
+		},
+		// 6: Dots in name, no labels
```
beorn7, line 465: As above, end with period.

expfmt/text_parse_test.go
```diff
@@ -385,6 +385,159 @@ request_duration_microseconds_count 2693
...
+							Counter: &dto.Counter{
+								Value: proto.Float64(.23),
+							},
+							TimestampMs: proto.Int64(1234567890),
+						},
+					},
+				},
+			},
+		},
+		// 7: Gauge, UTF-8, +Inf as value, multi-byte characters in label values.
```
beorn7, line 494: As above, "gauge" is not important. Say something about the special character rather than just stating "UTF-8".

expfmt/text_parse_test.go
```diff
@@ -385,6 +385,159 @@ request_duration_microseconds_count 2693
...
+								},
+							},
+							Gauge: &dto.Gauge{
+								Value: proto.Float64(3.14e42),
+							},
+						},
+					},
+				},
+			},
+		},
```
beorn7, line 540: We need more test cases. Ideas:

- Various escaped special characters in a label name and a metric name (like `\"`, but also including multi-line with `\n`).
- The name is not the first element: `{foo="bar", "late.name"}`.

It would be good to also add a few more cases to `testTextParseError`, e.g. multiple metric names like `{"one.name", "another.name"}`.

beorn7 (changes requested):
Thanks for doing this.

@ywwg mentioning you here in case you haven't seen it. You are probably more qualified to review this than me.

fedetorres93:
@beorn7 Thank you for your review. All feedback has been addressed.

beorn7 (approved):
LGTM. @ywwg do you want to give it a final look?

ywwg (approved):
have you run code coverage on this to make sure the major branches are covered?  Unless something is missing, lgtm

fedetorres93:
> have you run code coverage on this to make sure the major branches are covered? Unless something is missing, lgtm

Most branches were covered, but I added a couple more tests to cover additional error cases.

ywwg (approved):
thank you for the last tests! lgtm

beorn7:
Thank you both. I'll squash this because we have quite a few mini commits in here.

PR #677, opened 2024-08-16
feat: new promslog and promslog/flag packages to wrap log/slog

tjhop:
Prereq for prometheus/prometheus#14355

This adds a new `promslog` package to create an opinionated slog logger
for use within the prometheus ecosystem. By default, logs are written at
`info` level and formatted to add a kv pair for the source caller. If
backwards compatibility for the literal output format of the logging is
required, logs can be styled in a format that is similar to how the
`promlog` package formatted it's go-kit/log output [1].  The `promslog`
package also makes use of an `slog.LevelVar` to dynamically adjust the
level of an existing logger. Similar to the old `promlog` package, we
provide means to log in `logfmt` and `json` output formats.

Creating a logger is similar to the old promlog package -- optionally
populate a `Config` struct, and then call `New()` with the config.

In order to dynamically adjust the logger's level, retain the `Config`
struct as it's own variable to access the `AllowedLevel.Set()` method,
which internally updates the `AllowedLevel`'s slog.LevelVar to the
desired log level. Ex:

```go
config := &promslog.Config{} // Retain as variable if you need to dynamically adjust log level
logger := promslog.New(config)
config.Level.Set("debug")
logger.Debug("your message here", "hello", "world")
```

To use go-kit style log output:
```go
config := &promslog.Config{Style: promslog.GoKitStyle} // Retain as variable if you need to dynamically adjust log level
logger := promslog.New(config)
config.Level.Set("debug")
logger.Debug("your message here", "hello", "world")
```

1. When using the `go-kit` log style, the following changes are made to the
default slog output:
- adjusts slog default timestamp key from `timestamp` -> `ts`
- adjusts the timestamp value to use the same time format string
  ("2006-01-02T15:04:05.000Z07:00")
- adjusts slog default sourcecode key from `source` -> `caller`
- adjusts the formatting of the sourcecode values to trim paths with
  `filepath.Base()`. The formatting of the sourcecode value is similar
to the go-kit/log usage, with the addition of the source function in
parenthesis when debug logging is enabled.

### DIFF
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -1,0 +1,182 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package promslog defines standardised ways to initialize the Go standard
+// library's log/slog logger.
+// It should typically only ever be imported by main packages.
+
+//go:build go1.21
+
+package promslog
+
+import (
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+type LogStyle string
+
+const (
+	SlogStyle  LogStyle = "slog"
+	GoKitStyle LogStyle = "go-kit"
+)
+
+var (
+	LevelFlagOptions  = []string{"debug", "info", "warn", "error"}
+	FormatFlagOptions = []string{"logfmt", "json"}
+
+	callerAddFunc             = false
+	defaultWriter             = os.Stderr
+	goKitStyleReplaceAttrFunc = func(groups []string, a slog.Attr) slog.Attr {
+		key := a.Key
+		switch key {
+		case slog.TimeKey:
+			a.Key = "ts"
+
+			// This timestamp format differs from RFC3339Nano by using .000 instead
+			// of .999999999 which changes the timestamp from 9 variable to 3 fixed
+			// decimals (.130 instead of .130987456).
+			t := a.Value.Time()
+			a.Value = slog.StringValue(t.Format("2006-01-02T15:04:05.000Z07:00"))
+		case slog.SourceKey:
+			a.Key = "caller"
+			src, _ := a.Value.Any().(*slog.Source)
+
+			switch callerAddFunc {
+			case true:
+				a.Value = slog.StringValue(filepath.Base(src.File) + "(" + filepath.Base(src.Function) + "):" + strconv.Itoa(src.Line))
+			default:
+				a.Value = slog.StringValue(filepath.Base(src.File) + ":" + strconv.Itoa(src.Line))
+			}
+		case slog.LevelKey:
+			a.Value = slog.StringValue(strings.ToLower(a.Value.String()))
+		default:
+		}
+
+		return a
+	}
+)
+
+// AllowedLevel is a settable identifier for the minimum level a log entry
+// must be have.
+type AllowedLevel struct {
+	s   string
+	lvl *slog.LevelVar
+}
+
+func (l *AllowedLevel) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var s string
+	type plain string
+	if err := unmarshal((*plain)(&s)); err != nil {
+		return err
+	}
+	if s == "" {
+		return nil
+	}
+	lo := &AllowedLevel{}
+	if err := lo.Set(s); err != nil {
+		return err
+	}
+	*l = *lo
+	return nil
+}
+
+func (l *AllowedLevel) String() string {
+	return l.s
+}
+
+// Set updates the value of the allowed level.
+func (l *AllowedLevel) Set(s string) error {
+	if l.lvl == nil {
+		l.lvl = &slog.LevelVar{}
+	}
+
+	switch s {
+	case "debug":
+		l.lvl.Set(slog.LevelDebug)
+		callerAddFunc = true
+	case "info":
+		l.lvl.Set(slog.LevelInfo)
+		callerAddFunc = false
+	case "warn":
+		l.lvl.Set(slog.LevelWarn)
+		callerAddFunc = false
+	case "error":
+		l.lvl.Set(slog.LevelError)
+		callerAddFunc = false
+	default:
+		return fmt.Errorf("unrecognized log level %s", s)
+	}
+	l.s = s
+	return nil
+}
+
+// AllowedFormat is a settable identifier for the output format that the logger can have.
+type AllowedFormat struct {
+	s string
+}
+
+func (f *AllowedFormat) String() string {
+	return f.s
+}
+
+// Set updates the value of the allowed format.
+func (f *AllowedFormat) Set(s string) error {
+	switch s {
+	case "logfmt", "json":
+		f.s = s
+	default:
+		return fmt.Errorf("unrecognized log format %s", s)
+	}
+	return nil
+}
+
+// Config is a struct containing configurable settings for the logger
+type Config struct {
+	Level    *AllowedLevel
+	Format   *AllowedFormat
+	Style    LogStyle
+	ioWriter io.Writer
+}
+
+// New returns a new slog.Logger. Each logged line will be annotated
+// with a timestamp. The output always goes to stderr.
+func New(config *Config) *slog.Logger {
+	if config.Level == nil {
+		config.Level = &AllowedLevel{}
+		_ = config.Level.Set("info")
+	}
+
+	if config.ioWriter == nil {
+		config.ioWriter = defaultWriter
+	}
+
+	logHandlerOpts := &slog.HandlerOptions{
+		Level:     config.Level.lvl,
+		AddSource: true,
+	}
+
+	if config.Style == GoKitStyle {
+		logHandlerOpts.ReplaceAttr = goKitStyleReplaceAttrFunc
+	}
+
+	if config.Format != nil && config.Format.s == "json" {
+		return slog.New(slog.NewJSONHandler(config.ioWriter, logHandlerOpts))
+	}
+	return slog.New(slog.NewTextHandler(config.ioWriter, logHandlerOpts))
+}
--- a/promslog/slog_test.go
+++ b/promslog/slog_test.go
@@ -1,0 +1,150 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build go1.21
+
+package promslog
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"log/slog"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	slogStyleLogRegexp  = regexp.MustCompile(`(?P<TimeKey>time)=.*level=(?P<LevelValue>WARN|INFO|ERROR|DEBUG).*(?P<SourceKey>source)=.*`)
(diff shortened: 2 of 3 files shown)


tjhop:
@SuperQ @beorn7:

This should be ready for review now. I believe I was able to set things up in a way where flag interactions can remain consistent (by also creating a new promslog/flag package to pair with the logger). Formatting for the literal log lines themselves _should_ be pretty close old formatting of the logger from promlog, with the exception of the value of the `caller` source key -- I didn't look into trying to tap into `runtime.Caller()` myself for slog, as is used behind the scenes when we call `log.Caller()` [here](https://github.com/prometheus/common/blob/2cac84e1490776cacb3f46f14e7adbaf8dcb4596/promlog/log.go#L191). 

Feedback welcome!

SuperQ:
Nice!

My only question so far is how much we care about keeping the same log line formatting as before. Since this is intended for Prometheus 3.0, we can make breaking changes. 

Do we want to deviate from some of the minor slog defaults for formatting? Or use this as an opportunity to reset.

tjhop:
We can certainly use it as an opportunity to reset -- my primary thought was that by maintaining log formatting, we would cause the least downstream disruption for anything that is parsing the resulting log lines. We're ok with making the breaking changes to formatting for prometheus because of the 3.0 release, but any other downstream projects would also inherit those formatting changes. 

I'd like to not be the cause of someone getting paged in the middle of the night because an upgrade to $someService was done and their loki recording/alerting rules broke from unexpected formatting changes, etc :upside_down_face:

SuperQ:
So let's break this down into individual decisions:

1. adjusts slog default timestamp key from timestamp -> ts
2. adjusts the timestamp value to use the same time format string
("2006-01-02T15:04:05.000Z07:00")
3. adjusts slog default sourcecode key from source -> caller
4. adjusts the formatting of the sourcecode values to trim paths with
filepath.Base(). The formatting of the sourcecode value is similar
to the go-kit/log usage, with the addition of the source function in
parenthesis when debug logging is enabled.

My vote:
1. Yes
2. No
3. No
4. Maybe

tjhop:
There's also this (it made it in during a final refactor but I forgot to update commit description/PR summary):

5. The normalization of the `level` flag to use lowercase values, since slog's default is uppercase (`DEBUG` vs `debug`)

That said, I'll cast my vote:
1. I can go either way for timestamp key name
2. No
3. No
4. Yes for filepath trimming, maybe for adding in source func on debug level
5. Yes (purely personal preference)

promlog/log.go
```diff
@@ -14,6 +14,8 @@
 // Package promlog defines standardised ways to initialize Go kit loggers
 // across Prometheus components.
 // It should typically only ever be imported by main packages.
+
+// Deprecated: This package has been replaced with github.com/prometheus/common/promslog.
```
SuperQ, line 18: I'm wondering if we should wait one release of `prometheus/common` before adding the deprecation.

We have a lot of work to do to migrate, and we have linters in place that will complain about deprecated uses.

SuperQ (approved):

promslog/slog.go
```diff
@@ -0,0 +1,197 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+var (
+	LevelFlagOptions  = []string{"debug", "info", "warn", "error"}
+	FormatFlagOptions = []string{"logfmt", "json"}
+	StyleFlagOptions  = []string{"slog", "go-kit"}
```
ArthurSens, line 35: This feels strange 🤔. Why would a user decide on the logging library?

promslog/slog_test.go
```diff
@@ -0,0 +1,152 @@
...
+			levelIndex := re.SubexpIndex("LevelValue")
+
+			counters[strings.ToLower(matches[levelIndex])]++
+		}
+	}
+
+	return counters
+}
+
+func TestDynamigLevels(t *testing.T) {
```
ArthurSens, line 99:
```suggestion
func TestDynamicLevels(t *testing.T) {
```

promslog/slog_test.go
```diff
@@ -0,0 +1,152 @@
...
+	var buf bytes.Buffer
+	wantedLevelCounts := map[string]int{"info": 1, "debug": 1}
+
+	tests := map[string]struct {
+		logStyle         string
+		logStyleRegexp   *regexp.Regexp
+		wantedLevelCount map[string]int
+	}{
+		"empty_log_style":  {logStyle: "", logStyleRegexp: slogStyleLogRegexp, wantedLevelCount: wantedLevelCounts},
+		"bad_log_style":    {logStyle: "foo", logStyleRegexp: slogStyleLogRegexp, wantedLevelCount: wantedLevelCounts},
```
ArthurSens, line 109: Just to clarify, these two test cases will never make it into a real scenario because kingpin flag validation won't allow it, right?

ArthurSens: If that's not the case, could we make it so? It already feels strange that the used library is user decision, but it would feel even weirder if the user makes a typo like `log.style=go-kitt` and ends up with slog without errors.

promslog/slog_test.go
```diff
@@ -0,0 +1,152 @@
...
+			// file.
+			if err := config.Level.Set("info"); err != nil {
+				t.Fatal(err)
+			}
+			logger.Info("info", "hello", "world")
+			logger.Debug("debug", "hello", "world")
+
+			counts = getLogEntryLevelCounts(buf.String(), tc.logStyleRegexp)
+			require.Equal(t, tc.wantedLevelCount["info"], counts["info"], "info log successfully detected")
+			require.NotEqual(t, tc.wantedLevelCount["debug"], counts["debug"], "extra debug log detected")
```
ArthurSens, line 146:
```suggestion
			require.NotEqual(t, 0, counts["debug"], "extra debug log detected")
```

Not sure if I missed something, but shouldn't it be 0 now that debug was disabled?

ArthurSens (commented):
I think I'm a bit late with the discussions, but my main concern here is why would a Prometheus or prometheus exporter user decide the logging library used underneath 😬 

Was there any related discussions that I've missed?

promslog/slog.go
SuperQ: It's not selecting the logging library, but the style of the output. The `go-kit` output still uses slog, but styles slog to look like `go-kit`.

This is meant for providing output compatibility for people who are doing, probably questionable, parsing of the log lines.

SuperQ: We could also call this `promlog` style.

ArthurSens: Aaaah I see, but the idea of only moving to slog in 3.0 wasn't because we're allowed to make breaking changes? I'm fine with key attributes changing here 🤔

ArthurSens: Ah wait, we'll also change things in exporters as well right...

ArthurSens: Would it make sense if the library user (i.e. prometheus and exporter developers) could choose the style, but the tools user (prometheus and exporter users) could not?

I.e. make this an option in the logger constructor, but not a cmdline flag

SuperQ: > make this an option in the logger constructor, but not a cmdline flag

I don't mind either way. The flag idea isn't that important to me.

promslog/slog_test.go
tjhop: The value should indeed be zero -- for sake of reusing the simple predefined map of values, I inverted the logical test for this final case to test `log count != 1`

promslog/slog_test.go
tjhop: Yea, the tests here are using directly setting the log style via the `AllowedStyle`'s string field. That field isn't exported, so it's kinda cheating here to use it for the tests, I suppose. 

With regards to the flag, I'm going to pull it in a commit shortly in favor of a param in the log constructor

tjhop: I'll update the tests to set the style using the `Set()` method that's exposed to better mimic the flow a user of the library will go through when working with it :+1:

promslog/slog_test.go
ArthurSens: Oh, I didn't realize the Equal/NotEqual change 😛. LGTM

promslog/slog.go
tjhop: Currently, the `New()` function only returns an slog logger, no error. If we add a param for `style string`, then we should use the `AllowedStyle.Set()` method to parse/validate and set the provided style. But `AllowedStyle.Set()` can return an error -- how do we best surface that to users when creating a logger? 

Options I can think of:
- update `New()` to also return an error, if needed
- if we get an error, we just ignore and carry on setting to `slog` as default. if we do this, I can't think of a reasonably good way to surface that such a decision has been made to the caller of the function, though :thinking:

I'd vote for option 1 and am going to work towards that, unless there's strong opinions otherwise

tjhop: I'm also not sure it's overly cumbersome to have users set the style via a variable prior to logger creation to keep the current function signature for New(), eg:

```go
// default slog styling
config := &promslog.Config{}
logger := New(config)

// legacy go-kit similar styling
config := &promslog.Config{Style: &AllowedStyle{}}
if err := config.Style.Set("go-kit"); err != nil {
        // handle error
}
logger := New(config)
```

promslog/slog_test.go
tjhop: Tests have been updated to better mock out usage as a consumer of the library

tjhop:
I believe I've addressed all PR feedback, the only remaining item should be determining the approach we want to take on the style initialization, whether we take the constructor approach or advocate consumers use the existing methods when prepping the logger config

ArthurSens:
You could use go's type safety to prevent library users from setting weird log styles:


```go
type LogStyle string

const (
	SlogStyle  LogStyle = "slog"
	GoKitStyle LogStyle = "go-kit"
)

// Config is a struct containing configurable settings for the logger
type Config struct {
	Level    *AllowedLevel
	Format   *AllowedFormat
	Style    LogStyle
	ioWriter io.Writer
}
```

But while quickly trying this out I realized it would require to refactor the whole implementation 😵 . I'm fine with returning error from the constructor

ArthurSens:
Hey, I'm creating a new release to unblock the UTF8 work, but I can create another release tomorrow with this PR as well :)

promslog/slog.go
SuperQ: I'd prefer the `New()` with no `error` option.

I don't actually see any use of having `Config.Style` as a settable. I don't see us needing to change style on-the-fly. In fact, as-is, the style is only adjusted at `New()` time anyway.

We could simplify this a bit.

```go
type Style string

const (
	SlogStyle Style = "slog"
	GoKitStyle Style = "go-kit"
)

...

logger := promslog.New(promslog.Config{Style: promslog.SlogStyle})
```

tjhop: > I'd prefer the New() with no error option.
> I don't actually see any use of having Config.Style as a settable. 

Same, the current design was mostly born forward from aligning it with the existing log level/format configs which are settable. I agree style doesn't need to be a settable -- it's not like we intend to maintain logging styles other than `slog` default and `go-kit`. 

> I don't see us needing to change style on-the-fly. In fact, as-is, the style is only adjusted at New() time anyway.

Correct, and I'm not even sure it's possible without writing a custom slog Handler. The style formatting is done with the `ReplaceAttr()` func in the log handler options, and when the logger is created that handler is stored unexported on the handler and I don't see any stdlib functions that can modify an existing handler

https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/log/slog/logger.go;l=112-114

At the very least, it's possible to create multiple loggers and/or recreate existing loggers, but I think that's further support it doesn't need to be settable.

tjhop: Since we all seem to appreciate the style type approach, I'll get started refactoring :+1:

SuperQ (approved):

ArthurSens (approved):

ArthurSens:
Do you also need a new release ASAP to get things done in Prometheus?

promslog/slog.go
tjhop: log style has been made a custom type, it's no longer settable :+1:

tjhop:
@ArthurSens @SuperQ the last commit adjusts the log style to be a custom type, it's no longer settable :+1:

tjhop:
> Do you also need a new release ASAP to get things done in Prometheus?

A release would be helpful whenever y'all are comfortable with the final change set and have time. For now, I'm using go workspaces to start hacking on the prometheus 3.0 conversion :upside_down_face:

ArthurSens (approved):
Perfect!

SuperQ:
@tjhop Do you want to squash up commits and finalize the commit/PR descriptions?

tjhop:
> @tjhop Do you want to squash up commits and finalize the commit/PR descriptions?

no worries, can do :+1:

tjhop:
Rebased on current upstream main, squashed to a single commit, and updated the commit description/PR summary with some more accurate docs. Lemme know if you have any further thoughts :+1:

ArthurSens:
Thanks! Merging and doing the release :)

promslog/slog.go
```diff
@@ -0,0 +1,182 @@
...
+		key := a.Key
+		switch key {
+		case slog.TimeKey:
+			a.Key = "ts"
+
+			// This timestamp format differs from RFC3339Nano by using .000 instead
+			// of .999999999 which changes the timestamp from 9 variable to 3 fixed
+			// decimals (.130 instead of .130987456).
+			t := a.Value.Time()
+			a.Value = slog.StringValue(t.Format("2006-01-02T15:04:05.000Z07:00"))
```
dswarbrick, line 55: The non-UTC timestamp format here differs from the previous promlog (go-kit) timestamp format, which used `.UTC()`.

dswarbrick: PR created: https://github.com/prometheus/common/pull/696

PR #680, opened 2024-08-21
enable utf-8 validation mode by default

ywwg:
part of https://github.com/prometheus/prometheus/issues/13095

This should be part of Prometheus 3.0, do we need a separate feature branch for the common library so we can schedule when the change is committed?

### DIFF
--- a/model/alert_test.go
+++ b/model/alert_test.go
@@ -100,6 +100,8 @@ func TestAlertValidate(t *testing.T) {
 		},
 	}
 
+	NameValidationScheme = LegacyValidation
+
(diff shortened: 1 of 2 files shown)


beorn7:
Isn't the idea that Prometheus v3 can use UTF-8 by default without this change in common?

We can still have this change here, but we have to communicate to users of this package that they now have to actively opt out of UTF-8 support. I don't think the change here has to be locked to the v3 release in any way.

ywwg:
ahh that makes sense.  Closing for now.

PR #691, opened 2024-09-05
fix(utf8): Fix multiple metric name inside braces validation

fedetorres93:
There's a bug where metrics inside braces in different lines are mistakenly considered as multiple metric names in the same line when there's no TYPE or HELP line. This PR addresses the issue.

Related to https://github.com/prometheus/common/pull/670

### DIFF
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -73,8 +73,10 @@ type TextParser struct {
 	// These tell us if the currently processed line ends on '_count' or
 	// '_sum' respectively and belong to a summary/histogram, representing the sample
 	// count and sum of that summary/histogram.
-	currentIsSummaryCount, currentIsSummarySum                      bool
-	currentIsHistogramCount, currentIsHistogramSum                  bool
+	currentIsSummaryCount, currentIsSummarySum     bool
+	currentIsHistogramCount, currentIsHistogramSum bool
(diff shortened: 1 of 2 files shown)


expfmt/text_parse.go
```diff
@@ -73,9 +73,9 @@ type TextParser struct {
 	// These tell us if the currently processed line ends on '_count' or
 	// '_sum' respectively and belong to a summary/histogram, representing the sample
 	// count and sum of that summary/histogram.
-	currentIsSummaryCount, currentIsSummarySum     bool
-	currentIsHistogramCount, currentIsHistogramSum bool
-	currentMetricIsInsideBraces                    bool
+	currentIsSummaryCount, currentIsSummarySum                      bool
+	currentIsHistogramCount, currentIsHistogramSum                  bool
+	currentMetricIsInsideBraces, currentMetricInsideBracesIsPresent bool
```
bboreham, line 78: Can we get a comment explaining what these mean?

fedetorres93: Addressed in https://github.com/prometheus/common/pull/691/commits/acd0c007ef78f0864cf33cfdd8d0acb7b8a2b61a, thanks for the feedback

ywwg (approved):
ah, good find!

beorn7 (approved):

PR #696, opened 2024-09-20
promslog: use UTC timestamps for go-kit log style

dswarbrick:
promlog initialized its go-kit logger to use UTC timestamps, so for true backwards compatibility, the go-kit log style in promslog should retain that configuration.


tjhop (approved):

SuperQ (approved):
I think we should also do this for normal slog logging as well. We have a UTC policy for logging because we have a UTC policy for the metrics store in the TSDB.

This make sure we always get aligned time when comparing Prometheus logs and metrics.

PR #705, opened 2024-10-08
promslog: Only log basename, not full path

roidelapluie:


gotjosh (approved):
Not sure if this warrants a test but thought I would suggest it.

gotjosh (approved):

PR #706, opened 2024-10-10
WIP: expfmt/text: optionally include empty MetricFamilies

dgrisonnet:
When parsing text format it is sometimes useful to keep all the MetricFamilies around, even the ones that don't have any timeseries. This is the case for the testutil library in client_golang where we need to distinguish registered metrics from erronous inputs.


expfmt/text_create.go
```diff
@@ -76,10 +76,6 @@ var (
 //
 // This method fulfills the type 'prometheus.encoder'.
 func MetricFamilyToText(out io.Writer, in *dto.MetricFamily) (written int, err error) {
-	// Fail-fast checks.
-	if len(in.Metric) == 0 {
-		return 0, fmt.Errorf("MetricFamily has no metrics: %s", in)
-	}
```
dgrisonnet, line 82: Tests are failing because of this. I guess I would need to introduce a new function just to keep MetricFamilies without timeseries around?

bwplotka: ...or variadic option 🤔 or "both", so

NewMFConverter(opts).MetricFamilyToText

then keeping MetricFamilyToText as NewMFConverter(nil).MetricFamilyToText(....)

PR #710, opened 2024-10-28
Add openmetric parser[1]: add type,unit,help parser

jyz0309:
Ref: https://github.com/prometheus/prometheus/issues/8932
Part of #669 
Add openmetrics protocol decoder, for promtool to support check openmetric protocol.
This pr just parse the HELP, TYPE, UNIT as the [comment](https://github.com/prometheus/common/pull/669#issuecomment-2332317520) suggest


expfmt/openmetrics_parse_test.go
```diff
@@ -0,0 +1,307 @@
...
+			},
+		},
+	}
+
+	for i, scenario := range scenarios {
+		out, err := omParser.OpenMetricsToMetricFamilies(strings.NewReader(scenario.in))
+		if err != nil {
+			t.Errorf("%d. error: %s", i, err)
+			continue
+		}
```
ArthurSens, line 190: What do you think about using testify `assert` or `require`? I think it should be more readable, right? We could also add `name` to the scenario struct and use `t.Run(scenario.name...` instead of comments with numbers to make test results more readable.

ArthurSens: We already import testify in our go.mod, so this should be fine

expfmt/openmetrics_parse_test.go
```diff
@@ -0,0 +1,307 @@
...
+		if expected, got := len(scenario.out), len(out); expected != got {
+			t.Errorf(
+				"%d. expected %d MetricFamilies, got %d",
+				i, expected, got,
+			)
+		}
```
ArthurSens, line 196: We could use `require.Len` here

expfmt/openmetrics_parse_test.go
```diff
@@ -0,0 +1,307 @@
...
+func testOpenMetricParseError(t testing.TB) {
+	scenarios := []struct {
+		in  string
+		err string
+	}{
+		// 0:
+		{
+			in: `# TYPE metric counter
+# TYPE metric untyped
+`,
```
ArthurSens, line 225: Let's do the same here! Adding `name` to the scenarios and then using t.Run should provide more context about what each scenario is about :)

expfmt/openmetrics_parse.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return nil
+}
+
+// startComment represents the state where the next byte read from p.buf is the
+// start of a comment (or whitespace leading up to it).
+func (p *OpenMetricsParser) startComment() stateFn {
+	if p.skipBlankTab(); p.err != nil {
+		return nil // Unexpected end of input.
+	}
```
ArthurSens, line 145: I remember that we talked about this in the previous PR but I can't remember the outcome (sorry for being lazy here 😛). Does OpenMetrics allow lines starting with blanks/tabs?

ArthurSens (commented):
Good work here! I can see that this is basically a copy of the current text parser alongside its tests, but I think we can improve things a bit.

Comments are mostly around test readability

expfmt/openmetrics_parse.go
ArthurSens: Hmmm, I've just searched OM's spec and there isn't a single mention of the words `tab` or `whitespace`. Let's keep it like this then, but I wouldn't add tests for this since we don't need to offer strong guarantees about it

expfmt/openmetrics_parse_test.go
jyz0309: has a little confusing about this, because now there's benchmark for the test, so the arg in test function is testing.TB, and arg required by t.Run is *testing.T(`t.Run(name string, f func(t *testing.T))`), the testing.TB has not  implement the interface

PR #713, opened 2024-11-04
Support Application-Layer Protocol negotiation via nextProtos flag in tlsconfig

davidkydd:
Allows fallback based on protocol availability and support for custom protocols securely over TLS.

ptal:
@roidelapluie
@gotjosh
@ArthurSens

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -1069,6 +1069,10 @@ func NewTLSConfigWithContext(ctx context.Context, cfg *TLSConfig, optFuncs ...TL
 		tlsConfig.ServerName = cfg.ServerName
 	}
 
+	if len(cfg.NextProtos) > 0 {
+		tlsConfig.NextProtos = cfg.NextProtos
+	}
+
 	// If a client cert & key is provided then configure TLS config accordingly.
 	if cfg.usingClientCert() && cfg.usingClientKey() {
 		// Verify that client cert and key are valid.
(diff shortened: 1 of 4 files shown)


davidkydd:
we want to use ALPN protocols to identify certain scraping traffic routing through the APIServer, for a similar use case as konnectivity supports it: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/e4d5a2d64d0b5b7a28d73be99c34ae924190f2fa/cmd/agent/app/options/options.go#L117C2-L117C252

Its a generic capability so would likely have other uses - for example there is an open issue unrelated to our scenario here: https://github.com/prometheus/blackbox_exporter/issues/870

phealy:
To clarify David's post a bit - envoy can route traffic differently based on ALPN header and it's a good way to identify traffic or route connections without needing extra IPs or TLS decryption.

phealy:
This has a lot of the same potential cases as being able to set [server_name in the tls_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tls_config), though that can also be used for cert validation.

phealy:
> How is it different than setting enable_http2: false?

Setting http2 to false will change the ALPN header from something like "h2c,h2,http1.1" to just "http1.1". This PR instead lets us add extra things in there - for example, we could set it to "metrics,h2c,h2,http1.1", then use an envoy filter looking for the word "metrics" in the ALPN value to route the traffic on a VIP to the metrics destination instead of the default destination. Since that value is in cleartext in TLSClientHello, it's another nice way to route encrypted traffic without having to decrypt/reencrypt.

phealy:
@roidelapluie Here's a graphic that might make the idea clearer. The basic use case is traffic routing when you're not terminating TLS, based on something other than SNI. We have firewall restrictions that make adding extra public IPs or hostnames difficult, so this is a way to route traffic without needing any of that.

image

krajorama:
Hello from the bug scrub!

This is still on the backlog for the dev summit, having received only 3 votes (we discuss issues with most votes first).

ywwg:
We are closing this bug due to lack of activity and because we are unconvinced this feature is necessary and worth the maintenance burden of adding it.

PR #716, opened 2024-11-07
Provide a way to get UserAgent

mmorel-35:
In Prometheus, the userAgent is calculated the same way in several places:

https://github.com/search?q=repo%3Aprometheus%2Fprometheus%20userAgent&type=code

This exposes a function that can be used instead 

```go
import "github.com/prometheus/common/version"

version.UserAgent()
```

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -90,6 +90,10 @@ func GetTags() string {
 	return computedTags
 }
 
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Thanks!

PR #718, opened 2024-11-07
chore: use testify instead of testing.Fatal

mmorel-35:
This uses testify instead of testing for t.Fatal calls

### DIFF
--- a/assets/go.mod
+++ b/assets/go.mod
@@ -1,3 +1,11 @@
 module github.com/prometheus/common/assets
 
 go 1.21
+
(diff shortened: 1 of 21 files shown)


SuperQ (approved):
Nice

assets/go.mod
```diff
@@ -1,3 +1,11 @@
 module github.com/prometheus/common/assets
 
 go 1.21
+
+require github.com/stretchr/testify v1.9.0
```
SuperQ, line 5: Now that we have requirements here, we should enable dependabot for this directory.

Issue #723, opened 2024-11-15
Change default validation mode to UTF-8


ywwg:
With the release of Prometheus 3.0, we can set the default name validation mode to UTF-8, simplifying a bunch of code

PR #724, opened 2024-11-15
Change default validation scheme to UTF8Validation

ywwg:
fixes https://github.com/prometheus/common/issues/723

### DIFF
--- a/model/alert_test.go
+++ b/model/alert_test.go
@@ -22,6 +22,11 @@ import (
 )
 
 func TestAlertValidate(t *testing.T) {
+	oldScheme := NameValidationScheme
+	NameValidationScheme = LegacyValidation
+	defer func() {
+		NameValidationScheme = oldScheme
+	}()
 	ts := time.Now()
 
 	cases := []struct {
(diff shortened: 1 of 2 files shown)


SuperQ:
Is there any coordination with client_golang we need here? This seems like it may be a transparent breaking change for some systems.

ywwg:
I'll test it out -- probably yes we want to coordinate releases

bwplotka (approved):
Thanks!

Yea let's update client_golang straight away with this and explicitly set this if needed.

bwplotka:
Ok, this breaks in quite many places (client_golang, prometheus, outside users like Google), changing global and defer to change back does not scale really. We might consider fixing 0.62.0 and change all those functions to actually NOT use global cc @ywwg @SuperQ

bwplotka:
The challenge is UnmarshalJSON and so, plus doing this in backward compatible way. Unless we want to add more validation scheme OR the cases where legacy has to be used, it might be indeed easier to break everyone here ):

bwplotka:
Added big warning to release https://github.com/prometheus/common/releases/tag/v0.62.0

PR #725, opened 2024-11-16
Open user-agent definition

mmorel-35:
Most of the time the User-agent is Prometheus/Version but might be changed to Component/Version in certain occasion.
This expose a PrometheusUserAgent which is the most common case and also provide a ComponentUserAgent to support a component customization.

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -95,7 +95,7 @@ func PrometheusUserAgent() string {
 }
 
 func ComponentUserAgent(component string) string {
-	return fmt.Sprintf("%s/%s", component, Version)
+	return component+"/"+Version
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


version/info.go
```diff
@@ -90,8 +90,12 @@ func GetTags() string {
...
 }
 
-func UserAgent() string {
-	return "Prometheus/" + Version
+func PrometheusUserAgent() string {
+	return ComponentUserAgent("Prometheus")
+}
+
+func ComponentUserAgent(component string) string {
+	return fmt.Sprintf("%s/%s", component, Version)
```
SuperQ, line 98: How about this?

```suggestion
	return component+"/"+Version
```

SuperQ (approved):

PR #727, opened 2024-11-21
fix: values escaping bugs

ywwg:
I missed some test cases the first time around. Confirmed that without this fix the test cases fail the round trip.

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 
@@ -270,10 +271,6 @@ func metricNeedsEscaping(m *dto.Metric) bool {
 	return false
 }
 
-const (
-	lowerhex = "0123456789abcdef"
-)
-
 // EscapeName escapes the incoming name according to the provided escaping
 // scheme. Depending on the rules of escaping, this may cause no change in the
 // string that is returned. (Especially NoEscaping, which by definition is a
@@ -308,7 +305,7 @@ func EscapeName(name string, scheme EscapingScheme) string {
 			} else if isValidLegacyRune(b, i) {
 				escaped.WriteRune(b)
 			} else {
-				escaped.WriteRune('_')
+				escaped.WriteString("__")
 			}
 		}
 		return escaped.String()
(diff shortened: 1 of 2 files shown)


model/metric.go
```diff
@@ -391,7 +382,7 @@ func UnescapeName(name string, scheme EscapingScheme) string {
 			var utf8Val uint
 			for j := 0; i < len(escapedName); j++ {
 				// This is too many characters for a utf8 value.
-				if j > 4 {
+				if j > 8 {
```
jesusvazquez, line 385: Where does this come from?

model/metric.go
```diff
@@ -318,21 +315,15 @@ func EscapeName(name string, scheme EscapingScheme) string {
 		}
 		escaped.WriteString("U__")
 		for i, b := range name {
-			if isValidLegacyRune(b, i) {
+			if b == '_' {
+				escaped.WriteString("__")
```
jesusvazquez, line 319: Something rings a bell that this was decided when there is an underscore put two underscores but could we add a reference to where this was agreed on?

jesusvazquez (approved):
I've had a look and since tests are working It gives us some assurances. 

Left a couple questions for my own understanding.

model/metric_test.go
```diff
@@ -269,6 +277,14 @@ func TestEscapeName(t *testing.T) {
 			expectedUnescapedDots: "http.status:sum",
 			expectedValue:         "U__http_2e_status:sum",
 		},
+		{
+			name:                  "name with spaces and emoji",
+			input:                 "label with 😱",
+			expectedUnderscores:   "label_with__",
+			expectedDots:          "label_with__",
+			expectedUnescapedDots: "label_with_",
```
npazosmendez, line 285: I'm confused a bit by this -- `expectedUnescapedDots` is the expected output of un-escpaing `input` with the `DotsEscaping`? What if the input was not a valid escape sequence, like this case? Wouldn't we want to error out instead?

ywwg: the input is valid, and all non-legacy characters become underscores. We never want to produce an error.

However it's true that two non-legacy characters in a row become a double underscore, which the unescaper translates as a single underscore.  I am not sure there is a better alternative here, though. (non-legacy characters become `__`?)

model/metric.go
ywwg: the max UTF codepoint is '\U0010FFFF', therefore the max number of characters is actually 6, not 8.

model/metric_test.go
npazosmendez: I got confused with the test, sorry. I see that `expectedUnescapedDots` is the expected unescaping of the already escaped input. Looks good.

model/metric.go
jesusvazquez: Sorry I dived a bit and found a few constants, have a look at this https://go.dev/play/p/PoHxOOeoyQu

```
The maximum number of bytes needed to represent unicode.MaxRune is: 4
There is also utf8.UTFMAX: : 4
```

Shouldnt this remain as 4?

beorn7: Wait, it should be 7, right? The 4 bytes in the longest possible rune encode as 8 hexadecimal digits, so `j` can legally be between 0 and 7.

beorn7: Sorry for slow thinking. I think I got it now:

`utf8.UTFMAX` is about the bytes needed in UTF-8 encoded form. But what we put into the `U__` encoding is the _code point_, which is different from the encoded byte sequence. The codepoint \U0010FFFF that @ywwg mentioned would be written as `U__10FFFF`. As UTF-8 string, it would take the byte pattern `F4 8F BF BF` (the 4 bytes max length that `utf8.UTFMAX` refers to).

But this means the correct number here is 5, don't you think so?

beorn7 (approved):

model/metric.go
```diff
@@ -308,7 +305,7 @@ func EscapeName(name string, scheme EscapingScheme) string {
 			} else if isValidLegacyRune(b, i) {
 				escaped.WriteRune(b)
 			} else {
-				escaped.WriteRune('_')
+				escaped.WriteString("__")
```
fxrlv, line 308: why was it changed to two underscores?

fxrlv: https://prometheus.io/docs/instrumenting/escaping_schemes/#dots-escaping-dots

> Other non-legacy characters MUST be replaced with single underscores.

PR #728, opened 2024-12-01
promslog: always lowercase log level from CLI

jkroepke:
The `--log.level` argument now relaxes casing by converting all input from end-users to lowercase. 

This allows end-users to set `log.level` to values like `DEBUG`.


SuperQ (approved):
Good idea, thanks!

PR #735, opened 2024-12-04
fix(promslog): always use UTC for time

tjhop:
@superq and I have both called this out in a few places; finally remembering to fix it.

Prometheus has a UTC policy for logging because UTC is also used for metrics and so they should correlate.

Before:
```
=== RUN   TestDynamicLevels/slog_log_style
time=2024-12-02T13:51:08.463-05:00 level=INFO source=slog_test.go:120 msg=info hello=world
time=2024-12-02T13:51:08.463-05:00 level=DEBUG source=slog_test.go:121 msg=debug hello=world
```

After:
```
=== RUN   TestDynamicLevels/slog_log_style
time=2024-12-04T19:51:59.813Z level=INFO source=slog_test.go:120 msg=info hello=world
time=2024-12-04T19:51:59.813Z level=DEBUG source=slog_test.go:121 msg=debug hello=world
```

### DIFF
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -68,13 +68,16 @@ var (
 
 		return a
 	}
-	truncateSourceAttrFunc = func(groups []string, a slog.Attr) slog.Attr {
-		if a.Key != slog.SourceKey {
-			return a
-		}
-
-		if src, ok := a.Value.Any().(*slog.Source); ok {
+	defaultReplaceAttrFunc = func(groups []string, a slog.Attr) slog.Attr {
+		key := a.Key
+		switch key {
+		case slog.TimeKey:
+			t := a.Value.Time()
+			a.Value = slog.TimeValue(t.UTC())
+		case slog.SourceKey:
+			src, _ := a.Value.Any().(*slog.Source)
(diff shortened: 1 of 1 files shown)


SuperQ (approved):

beorn7:
I don't think we should revert the behavior as it is done here.

We had a bunch of feature requests to use the timezone configured in the environment for logging. I closed them as completed once the slog transition was done.

Log format is exempt from the stability guarantees, so we are technically allowed to change it without a major version bump. But we should still only do it as an informed decision.

At the very least, we should give users an option to use the default slog behavior again.

But I would even argue that the current behavior (prior to this PR) is fine. Those people that have seen the light will configure UTC on their servers and get UTC logging. Those that set a local timezone on their server apparently want it that way, so why log in UTC?

I also don't think that using UTC consistently for time series means we have to only use UTC timestamps for logs. The reasoning is quite different for each case.

juliusv:
> But I would even argue that the current behavior (prior to this PR) is fine. Those people that have seen the light will configure UTC on their servers and get UTC logging. Those that set a local timezone on their server apparently want it that way, so why log in UTC?

:+1: seems ok to me. Sometimes when I'm just testing something locally it's actually nicer to have local timestamps, as it's easier for me to tell when something happened.

But I don't mind much either way.

PR #739, opened 2024-12-05
promslog: Use the default timezone (again)

beorn7:
This reverts commit 145b50adb4ece06ea09a5b0abcbedb76bd1b123a.

See discussion on Slack. We should first decide which log format we want before changing it from what is already released in prometheus/prometheus.

UPDATE:

This is the new commit description:

As per the dev-summit discussion, we want to use the default timezone
as selected by slog after all rather than a hardcoded UTC.

(This was accidentally introduced earlier when moving from gokit to
slog, but then switched back to hardcoded UTC in PR https://github.com/prometheus/common/pull/735.)

Environments that have configured UTC anyway won't see any difference,
but those with other timezones configured will see the configured
timezone for the timestamps in their logs. This is in line with how
slog behaves by default.

Example timestamp _before_ this change and also _after_ this change
with UTC configured in the environment via a suitable way (e.g.
setting `TZ=UTC` on Linux):

    time=2025-04-10T12:00:38.179Z

Example timestamp _after_ this change with location Europe/Berlin,
i.e. timezone is CET or CEST:

    time=2025-04-10T14:00:03.120+02:00

Note that the precise delta to UTC is in the timestamp. So this change
will also be transparent to the usual logs processing tools.

### DIFF
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -200,9 +200,8 @@ func defaultReplaceAttr(_ []string, a slog.Attr) slog.Attr {
 	key := a.Key
 	switch key {
 	case slog.TimeKey:
-		if t, ok := a.Value.Any().(time.Time); ok {
-			a.Value = slog.TimeValue(t.UTC())
-		} else {
+		// Note that we do not change the timezone to UTC anymore.
+		if _, ok := a.Value.Any().(time.Time); !ok {
 			// If we can't cast the any from the value to a
 			// time.Time, it means the caller logged
 			// another attribute with a key of `time`.


SuperQ (changes requested):
There is no need to revert this. This is a bug fix for the previous release where we did not correctly make timestamps UTC.

beorn7:
It's a feature, not a bug… :)

But even if we come to the conclusion in the end that it is a bug fix, let's first come to that conclusion before we start cutting actual releases with the changed log format.

See also https://docs.google.com/document/d/1uurQCi5iVufhYHGlBZ8mJMK_freDFKPG0iYBQqJ9fvA/edit?pli=1&tab=t.0#bookmark=id.l370cy8m6ef4 .

SuperQ (approved):
Approved at 2025-03-31 Dev Summit

SuperQ:
I guess we need to do a bit of a rebase revert.

beorn7:
Yeah, other things happened with fine tuning the timestamp format. I'll push a new commit shortly.

beorn7:
Here we are. @SuperQ PHAL.

PR #741, opened 2024-12-24
Making the map a public variable for promtheus-operator

dongjiang1989:
ref: https://github.com/prometheus-operator/prometheus-operator/pull/7003#discussion_r1896566574


dongjiang1989:
cc @simonpasquier

SuperQ (approved):

PR #748, opened 2025-01-16
http_config: Allow customizing TLS config and settings.

bwplotka:
We (Google) would love to modify TLS config for custom authorization settings for our environments. Sometimes it means literally using our internal Go tls.Config instead of the created one by this client. This PR adds a minimal change to support a single option to do that, hope that makes sense!

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -453,13 +453,14 @@ func (a *BasicAuth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type DialContextFunc func(context.Context, string, string) (net.Conn, error)
 
 type httpClientOptions struct {
-	dialContextFunc   DialContextFunc
-	keepAlivesEnabled bool
-	http2Enabled      bool
-	idleConnTimeout   time.Duration
-	userAgent         string
-	host              string
-	secretManager     SecretManager
+	dialContextFunc     DialContextFunc
+	keepAlivesEnabled   bool
+	http2Enabled        bool
+	idleConnTimeout     time.Duration
+	userAgent           string
+	host                string
+	secretManager       SecretManager
+	extendTLSConfigFunc TLSConfigExtension
 }
 
 // HTTPClientOption defines an option that can be applied to the HTTP client.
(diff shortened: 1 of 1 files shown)


config/http_config.go
```diff
@@ -515,6 +516,17 @@ func WithHost(host string) HTTPClientOption {
 	})
 }
 
+// TLSConfigExtension modifies the given tls config and settings.
+type TLSConfigExtension func(*tls.Config, TLSRoundTripperSettings) (*tls.Config, TLSRoundTripperSettings, error)
```
machine424, line 520: maybe we can use the same naming as the existing `dialContextFunc DialContextFunc` (and group them)

bwplotka: group them meaning what?

machine424: Put them next to each other in the struct def. And move the func + type defs up next to the dial ones. For better discoverability.
(using the same naming pattern will help with that as well)

SuperQ: I also think the naming should match, `tlsConfigFunc TLSConfigFunc`.

bwplotka: addressed!

config/http_config.go
```diff
@@ -473,13 +477,22 @@ func (f httpClientOptionFunc) applyToHTTPClientOptions(options *httpClientOption
...
+// The default is `net.Dialer.DialContext`.
 func WithDialContextFunc(fn DialContextFunc) HTTPClientOption {
 	return httpClientOptionFunc(func(opts *httpClientOptions) {
 		opts.dialContextFunc = fn
 	})
 }
 
+// WithNewTLSConfigFunc allows you to override the func that creates the TLS config
+// from the prometheus http config.
+func WithNewTLSConfigFunc(newTLSConfigFunc NewTLSConfigFunc) HTTPClientOption {
```
machine424, line 490: nit (for consistency): we can get rid of the `New/new` as the dial ones don't mention them.

config/http_config.go
```diff
@@ -670,6 +683,14 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
 		return rt, nil
 	}
 
+	if opts.newTLSConfigFunc != nil {
+		tlsConfig, err := opts.newTLSConfigFunc()
+		if err != nil {
+			return nil, err
+		}
+		return newRT(tlsConfig)
```
machine424, line 691: the first version wasn't short circuiting here, any reason why not to follow branches below?

bwplotka: It's because we don't expect custom tlsConfig to be runtime reloadable as it might be built outside of the provided HTTP config. I can keep the old pattern if we want.

bwplotka: Ok, I did exact 1:1: NewTLSConfigWithContext semantics so it's simpler on our side.

config/http_config.go
bwplotka: kindly disagree, this is now consistent with dial, because `dial` is `dialling` and `newTLSConfig` is creating (thus new) config.

SuperQ (approved):

PR #754, opened 2025-01-31
promslog: Make AllowedLevel concurrency safe.

bwplotka:
Needed for https://github.com/prometheus/prometheus/issues/10352

Also I renamed AllowedLevel and AllowedFormat to Level and Format. Default level (and String()) is also now 'info' not empty.

It's a breaking change, but I suspect nobody was using those constructs directly, WDYT?

cc @tjhop

### DIFF
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -28,6 +28,7 @@ import (
 	"time"
 )
 
+// LogStyle represents the common logging formats in the Prometheus ecosystem.
 type LogStyle string
 
 const (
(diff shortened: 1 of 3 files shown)


promslog/slog_test.go
```diff
@@ -43,29 +43,29 @@ func TestDefaultConfig(t *testing.T) {
...
 func TestUnmarshallEmptyLevel(t *testing.T) {
-	l := &AllowedLevel{}
+	l := NewLevel()
 	err := yaml.Unmarshal([]byte(``), l)
 	if err != nil {
 		t.Error(err)
 	}
-	if l.s != "" {
-		t.Errorf("expected empty level, got %s", l.s)
+	if got := l.String(); got != "info" {
```
bwplotka, line 62: I changed the behaviour purposefully. This simplifies (and optimizes) the level code.

I don't know exactly how Unmarshal and String was envisioned to be used in the YAML context. Does this blocks anyone?

promslog/slog.go
```diff
@@ -155,55 +70,60 @@ func (l *AllowedLevel) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 	}
-	l.s = s
 	return nil
 }
 
-// AllowedFormat is a settable identifier for the output format that the logger can have.
-type AllowedFormat struct {
+// Format controls a logging output format.
+// Not concurrency-safe.
+type Format struct {
```
bwplotka, line 114: As mentioned in the description - this is a breaking change, let me know if this is ok.

promslog/slog.go
```diff
@@ -38,115 +39,29 @@ const (
...
 )
 
-// AllowedLevel is a settable identifier for the minimum level a log entry
-// must be have.
-type AllowedLevel struct {
-	s   string
+// Level controls a logging level, with an info default.
+// It wraps slog.LevelVar with string-based level control.
+// Level is safe to be used concurrently.
+type Level struct {
```
bwplotka, line 53: As mentioned in the description - this (rename) is a breaking change, let me know if this is ok.

promslog/slog.go
SuperQ: Yes, this is a breaking change.

Any chance you would keep the old struct around and add a `Deprecated` comment so downstream Go users get a linter warning?

SuperQ: Or, maybe it's better to just drop the struct so it's a compile error.

tjhop:
> It's a breaking change, but I suspect nobody was using those constructs directly, WDYT?

There are at least a few that I know of:

promtheus json file logger:
https://github.com/prometheus/prometheus/blob/main/util/logging/file.go#L48-L53

prom api testing
https://github.com/prometheus/prometheus/blob/main/web/api/v1/api_test.go#L483-L484

blackbox exporter custom probe logging (handler.go and handler_test.go)
https://github.com/prometheus/blackbox_exporter/blob/master/prober/handler.go#L172

Updates shouldn't be difficult, but a few will be needed after the breaking change

tjhop (approved):
I'm on board :+1:

SuperQ (approved):

promslog/slog.go
bwplotka: For deprecated flow we would need to maintain a behaviour which means `promslog.Config` using old thing, we can but until we are v1 and this is not too painful we can break.

amotl:
Hi there. Thanks a stack for your work on this package!

> It's a breaking change, but I suspect nobody was using those constructs directly

Our code apparently used one of the symbols, but we have been able to resolve that speed bump quickly.

- https://github.com/crate/cratedb-prometheus-adapter/pull/212

NB: This is just FYI for people who may run into the same problem.

bwplotka:
Thanks and sorry @amotl for the impact. 

I just started a new discussion about potential v1.0 for this module so we can be more careful with breakages gong forward https://github.com/prometheus/common/issues/771

denysvitali:
Just fyi, this broke a lot of stuff we use internally at our company too 🙃

And it looks like we were not alone: VictoriaMetrics had to pin the previous version because of this:
https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8905

PR #755, opened 2025-01-31
otlp: Move otlptranslator package from prom/prom to here

ywwg:
For eventual import by otel collector


ArthurSens:
Nice, thanks for starting this!

I believe we agreed on just moving the "name translator" and not the remotewrite translator. I can't remember exactly why we came to this agreement though 🤔

ywwg:
If that was an agreement between you and me, I think that was based on a misunderstanding I had about the code arrangement.  Based on the diffing I've done, the translators are pretty close together between prom and otel-col.  I think what I'll do is try to work ahead a few PRs and confirm that otel will work when linked against this code. If not then we can avoid rolling things back

ArthurSens:
> If that was an agreement between you and me

No, it was within the otel-prom SIG call. It was probably captured in [this meeting note's entry](https://docs.google.com/document/d/19bnXziPn2MZ9wO6684UoI4D-LCjGL5bTJkGhux29bx8/edit?tab=t.0#bookmark=id.u09pe6fysknc), but, if after digging deeper into the code it might sound too difficult, we can re-think that decision

ywwg:
ok yes I think that stemmed from my misunderstanding.  I now think it's possible for us to combine them.  I will continue working and let you know how it turns out

PR #759, opened 2025-02-11
Test validation with dots

m1k1o:
Since in [v0.62.0](https://github.com/prometheus/common/releases/tag/v0.62.0) was introduced a new validation, that was advised as a breaking change, it indeed it was transparent breaking change for me.

I am using OTEL that is exporting metrics via prometheus. They contain dots in the name e.g. `module.submodule.metric_name`. Up until this change, they were changed to `module_submodule_metric_name` and everything worked properly.

Since this breaking change, they do contain the dot. However, only when requesting metrics with `escaping=allow-utf-8` what is prometheus doing by default.

It was tricky to find the root of this behavior change, since `curl 127.0.0.1:9000/metrics` yields correct results, but `curl -H 'Accept: text/plain;escaping=allow-utf-8' 127.0.0.1:9000/metrics` adds dots to the name.

I don't know if this change was intended, if I should  change my metrics everywhere to include the dot or not. In any case, I just added tests to validate with dots so the expected behavior is defined in tests.


ywwg:
> yields correct results, but `curl -H 'Accept: text/plain;escaping=allow-utf-8' 127.0.0.1:9000/metrics` adds dots to the name

I would reword this because the use of "correct" and "adds dots" are misleading:  curling without setting the Accept headers to indicate that your endpoint can process UTF-8 names yields results that match the behavior you expected because the library is escaping the names as it did in previous versions.  However with the Accept header that indicates support for utf-8, the library does not escape the names and instead allows them through as-is. Dots are not being "added" anywhere, they are no longer being escaped to underscores.

This behavior is already tested in numerous places but if it helps to have this one as well that's ok by me.

ywwg (approved):

ywwg:
for your situation, you probably want to set the `metric_name_validation_scheme` config option to "legacy" which will tell prometheus to escape the names.

m1k1o:
@ywwg setting `metric_name_escaping_scheme` to  `underscores` fixes my issue. but still having those tests would be good for future development. i think it could be merged, or?

m1k1o:
Ping beorn7.

ywwg (approved):

PR #760, opened 2025-02-13
feat: add an inline OM created timestamp syntax

Maniktherana:
this PR aims to update the created timestamp syntax as outlined in the following proposal: https://github.com/prometheus/proposals/pull/43

### DIFF
--- a/expfmt/openmetrics_create.go
+++ b/expfmt/openmetrics_create.go
@@ -22,9 +22,8 @@ import (
 	"strconv"
 	"strings"
 
-	"google.golang.org/protobuf/types/known/timestamppb"
-
 	"github.com/prometheus/common/model"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	dto "github.com/prometheus/client_model/go"
 )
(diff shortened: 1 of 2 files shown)


ArthurSens (commented):
Nice, thanks for starting this!

I think we may want a different approach though, OpenMetrics 1.0 is not going away, so we need to keep the _created lines for users of that version of the protocol.

What we want is a separate implementation for 2.0.

Maniktherana:
> Nice, thanks for starting this!
> 
> I think we may want a different approach though, OpenMetrics 1.0 is not going away, so we need to keep the _created lines for users of that version of the protocol.
> 
> What we want is a separate implementation for 2.0.

understood, how can we manage both implementations together? Do we use a feature flag for this?

ArthurSens:
Feedback from our call:

Clients who use prometheus/common to encode data use `NewEncoder` and it returns a different encoder based on Content Type.

We can try two options and see what works best:
* Create a totally new encoder MetricFamilyToOpenMetrics2.0
or
* The existing MetricFamilyToOpenMetrics encoder could have a new encoder option that decides its version

PR #765, opened 2025-03-04
model: Clarify the purpose of model.NameValidationScheme

ywwg:
There is a lot of confusion around this global variable. Clarify that it is meant to indicate that a project is aware of UTF-8 support, and that those projects should have their own flags to control validation mode.


model/metric.go
```diff
@@ -27,13 +27,25 @@ import (
...
+	// when appropriate.
+	//
+	// A more appropriate name for this variable would have been UTF8Aware or
+	// UTF8Capable. The authors regret the misleading name choice and gesture
+	// towards a whiteboard on which the two hardest problems in computer science
+	// are listed.
+	//
+	// This value can be set to LegacyValidation during startup if a binary is not
+	// UTF-8-aware. To avoid need for locking, this value should be set once,
+	// ideally in an init(), before multiple goroutines are started.
```
bwplotka, line 48: Thanks!

I would suggest stronger words around deprecation. Then I'm not sure if the "perhaps it should be named" would be useful here. First of all, this is only true, because Prometheus abused this variable, prometheus/common only uses this for validation, so the name fits well. However, perhaps it's nice to admit the mistake, the main reason why it existed in the past and the fact we don't plan to change the value anymore.

WDYT?

```suggestion
	// NameValidationScheme determines the global default method of the name validation
	// to be used by all calls to IsValidMetricName() and LabelName IsValid().
	//
	// Deprecated: This variable should not be used and maybe be removed in the far future. If you
	// wish to stick to the legacy name validation use `IsValidLegacyMetricName()` and
	// `LabelName.IsValidLegacy()` methods instead. This variable is here as an escape hatch
	// for emergency cases, given the recent change from `LegacyValidation` to `UTF8Validation`,
	// e.g., to delay UTF-8 migrations in time or aid in debugging unforeseen results of the change.
	// In such a case, a temporary assignment to `LegacyValidation` value in the `init()` function
	// in your main.go or so, could be considered.
	//
	// Historically we opted for a global variable for feature gating different validation
	// schemes in operations that were not otherwise easily adjustable (e.g. Labels yaml
	// unmarshaling). That could have been a mistake, a separate Labels structure or package
	// might have been a better choice. Given the change was mode and many upgraded the
	// common already, we live this as-is with this warning and learning for the future.
```

bwplotka (commented):
Thanks!

I would suggest stronger words around deprecation, plus some historical context. WDYT?

model/metric.go
```diff
@@ -27,13 +27,22 @@ import (
...
-	// all calls to IsValidMetricName() and LabelName IsValid(). Setting UTF-8
-	// mode in isolation from other components that don't support UTF-8 may result
-	// in bugs or other undefined behavior. This value can be set to
-	// LegacyValidation during startup if a binary is not UTF-8-aware binaries. To
-	// avoid need for locking, this value should be set once, ideally in an
-	// init(), before multiple goroutines are started.
+	// NameValidationScheme determines the global default method of the name validation
+	// to be used by all calls to IsValidMetricName() and LabelName IsValid().
+	//
+	// Deprecated: This variable should not be used and maybe be removed in the far future. If you
```
bwplotka, line 33:
```suggestion
	// Deprecated: This variable should not be used and might be removed in the far future. If you
```

bwplotka (approved):
Thanks!

model/metric.go
```diff
@@ -27,13 +27,22 @@ import (
...
+	// `LabelName.IsValidLegacy()` methods instead. This variable is here as an escape hatch
+	// for emergency cases, given the recent change from `LegacyValidation` to `UTF8Validation`,
+	// e.g., to delay UTF-8 migrations in time or aid in debugging unforeseen results of the change.
+	// In such a case, a temporary assignment to `LegacyValidation` value in the `init()` function
+	// in your main.go or so, could be considered.
+	//
+	// Historically we opted for a global variable for feature gating different validation
+	// schemes in operations that were not otherwise easily adjustable (e.g. Labels yaml
+	// unmarshaling). That could have been a mistake, a separate Labels structure or package
+	// might have been a better choice. Given the change was mode and many upgraded the
```
grobinson-grafana, line 44:
```suggestion
	// might have been a better choice. Given the change was made and many upgraded the
```

model/metric.go
```diff
@@ -27,13 +27,25 @@ import (
...
+	// given the recent change from `LegacyValidation` to `UTF8Validation`, e.g.,
+	// to delay UTF-8 migrations in time or aid in debugging unforeseen results of
+	// the change. In such a case, a temporary assignment to `LegacyValidation`
+	// value in the `init()` function in your main.go or so, could be considered.
+	//
+	// Historically we opted for a global variable for feature gating different
+	// validation schemes in operations that were not otherwise easily adjustable
+	// (e.g. Labels yaml unmarshaling). That could have been a mistake, a separate
+	// Labels structure or package might have been a better choice. Given the
+	// change was mode and many upgraded the common already, we live this as-is
```
grobinson-grafana, line 47:
```suggestion
	// change was made and many upgraded the common already, we live this as-is
```

bwplotka (approved):

PR #770, opened 2025-03-13
Update Go

SuperQ:
* Update Go to support 1.24.
* Update minimum Go version to 1.23.0.
* Bump Go modules.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus/common
 
-go 1.21
+go 1.23.0
```
ArthurSens, line 3: Would this force golang to set their minimum go version to 1.23 as well? I believe we're still in 1.22 over there 🤔

SuperQ: Yes, unfortunately, the upstream Go maintainers (`golang.org/x/oauth2`) have set this as a minimum requirement.

There's now a upstream Go bot auto-updating this: https://github.com/golang/oauth2/commit/3f78298beea38fb76a3fbca33e3056f4b7eb5502

We'll have to reconsider our support policy for client_golang versions. :crying_cat_face:

SuperQ:
Ping client_golang maintainers.

Upstream Go is now bumping minimum Go versions in `golang.org/x/` packages to only support the latest 2 releases. This forces us to bump our support here in order to pull in updates that include security fixes.

This is blocking some updates that are making vulnerability scanners complain. Even if there are no actual issues it would be good to move forward with updating our Go version support in common/client_golang to match.

ArthurSens (approved):

PR #772, opened 2025-03-19
Define code owners for otlptranslator

aknuds1:
If OTLP translation code should be moved from Prometheus to the `otlptranslator` package in this repo, `otlptranslator` should have the same CODEOWNERS as in Prometheus.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+otlptranslator @aknuds1 @jesusvazquez


jesusvazquez (approved):
@ArthurSens might want to add himself too but maybe not blocking. 

Looks good to me.

ArthurSens:
Not sure if needed since we received feedback to move it to a separate repo. We will deprecate and remove this package, aren't we?

aknuds1:
Yep, sounds as if we're moving that way @ArthurSens.

PR #781, opened 2025-04-16
feat(http_config): support JWT token auth as alternative to client secret (RFC 7523 3.1)

jkroepke:
This PR implements JWT Authentication for Authorization Grant ([RFC 7523 3.1](https://datatracker.ietf.org/doc/html/rfc7523#section-3.1)) as alternative to static client credentials.

As I know, this auth flow is used by STACKIT Service Accounts (Key Flow).

I tried to not break any exists method signatures to remain backwards compatibility. 

---

Note: This PR does not provide support for JWTs in [Client Authentication Processing](https://datatracker.ietf.org/doc/html/rfc7523#section-3.2), since it not supported by x/oauth2 yet. See https://github.com/golang/oauth2/pull/745

As I know, JWTs in Client Authentication Processing is used by [Microsoft Entra ID (Client Certificates)](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow#second-case-access-token-request-with-a-certificate).


config/http_config.go
```diff
@@ -241,8 +246,22 @@ type OAuth2 struct {
...
+
+	ClientCertificateKeyID   string `yaml:"client_certificate_key_id" json:"client_certificate_key_id"`
+	ClientCertificateKey     Secret `yaml:"client_certificate_key" json:"client_certificate_key"`
+	ClientCertificateKeyFile string `yaml:"client_certificate_key_file" json:"client_certificate_key_file"`
+	// ClientCertificateKeyRef is the name of the secret within the secret manager to use as the client
+	// secret.
+	ClientCertificateKeyRef string `yaml:"client_certificate_key_ref" json:"client_certificate_key_ref"`
+	// GrantType is the OAuth2 grant type to use. It can be one of
+	// "client_credentials" or "urn:ietf:params:oauth:grant-type:jwt-bearer" (RFC 7523).
+	GrantType string `yaml:"grant_type" json:"grant_type"`
```
SuperQ, line 258: This is now a required field. But it appears to default to `client_credentials`.

Issue #745, opened 2025-01-10
bug: promslog panics if a user sets a log attribute with a key of `source` 


tjhop:
This issue is for tracking purposes, @SuperQ and I are discussing things further in slack. 

---

In our slog implementation, we always instruct slog to emit source code info of the calling log statement:
https://github.com/prometheus/common/blob/main/promslog/slog.go#L183

We also do some default formatting of the resulting source code position info, both in the default implementation and the legacy go-kit styled implementation:
https://github.com/prometheus/common/blob/main/promslog/slog.go#L77-L79
https://github.com/prometheus/common/blob/main/promslog/slog.go#L54-L63

As mentioned in slog's docs, the `AddSource` field adds a key-value pair onto the logger keyed by `SourceKey`, which is a constant value of `source`:
https://pkg.go.dev/log/slog#pkg-constants

We key on that `SourceKey` in our `ReplaceAttr` funcs that we use for formatting log lines, and when an attribute is found with a key equal to `SourceKey`, it extracts the `any` from the value and casts it to an `*slog.Source`. We do not check the error from the type cast, as slog always adds source info as an attribute with a key of `SourceKey` and value of `slog.Source`. 

This implementation is naive, as users may add their own attributes with a key of `source`, causing promslog to panic:
https://github.com/SuperQ/chrony_exporter/issues/107

Putting aside the usefulness/validity of duplicate keys in a structured log message, this is quite the landmine for users to run into, and promslog should handle it more gracefully.

PR #783, opened 2025-05-01
refactor(promslog): make `NewNopLogger()` wrapper around `New()`

tjhop:
While discussing the fix for prometheus/prometheus#16466, it was pointed
out that our `promslog.NewNopLogger()` convenience function would
benefit from using the same logic used when initializing loggers with
`promslog.New()`. By refactoring NewNopLogger to wrap New, it inherits
that same setup logic.

### DIFF
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -266,10 +266,5 @@ func New(config *Config) *slog.Logger {
 // NewNopLogger is a convenience function to return an slog.Logger that writes
 // to io.Discard.
 func NewNopLogger() *slog.Logger {
-	// Set log level to debug, in case there are code paths that are only
-	// triggered when debug logging is enabled.
-	level := NewLevel()
-	_ = level.Set("debug")
-
-	return New(&Config{Level: level, Writer: io.Discard})
+	return New(&Config{Writer: io.Discard})
 }
--- a/promslog/slog.go
+++ b/promslog/slog.go
@@ -266,5 +266,10 @@ func New(config *Config) *slog.Logger {
 // NewNopLogger is a convenience function to return an slog.Logger that writes
 // to io.Discard.
 func NewNopLogger() *slog.Logger {
-	return slog.New(slog.NewTextHandler(io.Discard, nil))
+	// Set log level to debug, in case there are code paths that are only
+	// triggered when debug logging is enabled.
+	level := NewLevel()
+	_ = level.Set("debug")
+
+	return New(&Config{Level: level, Writer: io.Discard})
 }


tjhop:
cc: @machine424 @SuperQ

machine424:
Thanks for this.

Thinking about it more for the Prometheus use case, others may argue that the logger is no longer truly a “no-op”, it does more things now, and the added overhead may not be acceptable in some scenarios.

If that’s a concern, perhaps we can have that under a new `NewDiscardLogger`, `NewSilentLogger` or something to avoid that overhead. In Prometheus, we’d use that everywhere (we could even employ forbidigo or something similar to block NewNopLogger usage, as for Prometheus itself we always initialize a new real logger) and clearly communicate this to “Prometheus as a library” users. They could still override it on their side with `NewNopLogger` (if they use common) or any other no-op logger if they need.

cc @bboreham @dimitarvdimitrov as they worked on related stuff according to https://github.com/prometheus/prometheus/pull/15993

bboreham:
I see the genesis of this was for testing.  I don't think we want to turn on debug logging for nop-logger in non-test situations.

Also I'm going to change the title to reflect this change.

tjhop:
Agreed, I don't find the debug logging to be a requirement here, personally. It was a suggestion from the linked issue, the original patch I suggested just updated the function to wrap New()

tjhop:
I think it's reasonable to do within the context of the existing noop logger. This is the prometheus project's opinionated library, and I think aligning the initialization of the constructors of New() and NewNopLogger() is reasonable. It still accomplishes the same goal that a library user has, which is a logger that doesn't do anything. 

I don't disagree with you on the semantics though -- the promslog's NewNopLogger() was arguably never a true noop logger, as it always just writes to io.Discard, whereas a proper noop logger should likely be a custom handler that just returns nil in the Handle() func. It was created as a convenience function and named like go-kit/log's NewNopLogger(), which was heavily used through the prometheus ecosystem before we adopted log/slog. 

> It's not ideal that when you need the Debug logs the most you find them panicing, it would just add to the frustration.

Fully agreed, but just to be clear, the level isn't relevant to this particular bug; #745 and #746 have more details, but the original formatting code I wrote for log formatting was naive and didn't handle duplicate keys well. This bug could've triggered at any log level. I'll still vote to remove the debug level and update the PR; it can always be added in the future if needed

machine424 (approved):
thanks for the context.
with `if !l.Enabled(ctx, level) {` early returns, we'd still have code that only run at debug level, but let's start with this, if others are ok with it. We can always reconsider later.

machine424:
by the way, I had given this version a try on Prometheus, and only a test `TestNewGroup` which compare loggers would need to be adapted.

SuperQ (approved):

PR #792, opened 2025-06-03
Fix panic in `tlsRoundTripper` when CA file is absent

ndk:
https://github.com/prometheus/prometheus/issues/16622#issuecomment-2935997383

Prevents a nil-pointer panic in `tlsRoundTripper` that occurred when cert_file/key_file were set but no ca_file, and the CA reload failed. The code now returns a descriptive error instead of crashing.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -1377,6 +1377,9 @@ func (t *tlsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	// using GetClientCertificate.
 	tlsConfig := t.tlsConfig.Clone()
 	if !updateRootCA(tlsConfig, caData) {
+		if t.settings.CA == nil {
+			return nil, errors.New("unable to use specified CA cert: none configured")
+		}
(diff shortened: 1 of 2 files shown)


cxdy:
We got around this by setting the `ca_cert` in our `remote_write.tls_config` and in our scrape jobs, but this would be really nice to have!

krajorama (approved):
Hello from the bug scrub. Seems like a straightforward guard statement. Merging.

machine424:
This fixes the panic but `RoundTrip` would still fail.
`t.settings.CA` being empty is possible/legit and https://github.com/prometheus/common/pull/707 explicitly allowed that.
I think `!updateRootCA(tlsConfig, caData)` should be skipped if `t.settings.CA` is nil.
And we'll need to adjust TestTLSRoundTripper_NoCAConfigured as well.

I have https://github.com/prometheus/prometheus/pull/18727 that shows that the certs are never rotated. (because `RoundTrip()` keeps failing)

cc @roidelapluie @krajorama

PR #793, opened 2025-06-12
feat: Support negative duration in new function ParseDurationAllowNegative

iamhalje:
This PR extend the `ParseDuration` function to support negative durations.

negative durations need be useful in Loki alerts templates where time offsets such as '-5m' are common in range calculations and comparisons.

### Changes:
- update `ParseDuration` to handle negative values like `-1h`, `-30m`, etc.
- preserves compatibility with existing valid duration strings.
- includes error checks for unit order and duration overflows.

related to promethes/template changes for Loki alerting - see [prometheus/prometheus#16669](https://github.com/prometheus/prometheus/pull/16619)

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -201,6 +201,7 @@ var unitMap = map[string]struct {
 
 // ParseDuration parses a string into a time.Duration, assuming that a year
 // always has 365d, a week always has 7d, and a day always has 24h.
+// Negative durations are not supported.
 func ParseDuration(s string) (Duration, error) {
 	switch s {
 	case "0":
@@ -257,9 +258,7 @@ func ParseDuration(s string) (Duration, error) {
 	return Duration(dur), nil
 }
 
-// ParseDurationAllowNegative parses a string into a time.Duration, assuming that a year
-// always has 365d, a week always has 7d, and a day always has 24h.
-// Supporting negative durations as well.
+// ParseDurationAllowNegative is like ParseDuration but also accepts negative durations.
 func ParseDurationAllowNegative(s string) (Duration, error) {
 	switch s {
 	case "0":
--- a/model/time.go
+++ b/model/time.go
@@ -260,69 +260,20 @@ func ParseDuration(s string) (Duration, error) {
 
 // ParseDurationAllowNegative is like ParseDuration but also accepts negative durations.
 func ParseDurationAllowNegative(s string) (Duration, error) {
-	switch s {
-	case "0":
-		// Allow 0 without a unit.
-		return 0, nil
-	case "":
+	if s == "" {
 		return 0, errors.New("empty duration string")
 	}
 
-	orig := s
-	var dur uint64
-	lastUnitPos := 0
-
-	negative := false
-	if s[0] == '-' {
-		negative = true
-		s = s[1:]
+	if s[0] != '-' {
+		return ParseDuration(s)
 	}
 
-	for s != "" {
-		if !isdigit(s[0]) {
-			return 0, fmt.Errorf("not a valid duration string: %q", orig)
-		}
-		// Consume [0-9]*
-		i := 0
-		for ; i < len(s) && isdigit(s[i]); i++ {
-		}
-		v, err := strconv.ParseUint(s[:i], 10, 0)
-		if err != nil {
-			return 0, fmt.Errorf("not a valid duration string: %q", orig)
-		}
-		s = s[i:]
-
-		// Consume unit.
-		for i = 0; i < len(s) && !isdigit(s[i]); i++ {
-		}
-		if i == 0 {
-			return 0, fmt.Errorf("not a valid duration string: %q", orig)
-		}
-		u := s[:i]
-		s = s[i:]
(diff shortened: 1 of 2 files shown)


model/time.go
```diff
@@ -214,6 +214,13 @@ func ParseDuration(s string) (Duration, error) {
 	var dur uint64
 	lastUnitPos := 0
 
+	// Check if duration is negative.
+	neg := false
```
beorn7: Maybe let's call this `negative` and then get rid of all the other code comments below ("Return negative duration", "Handle negative duration") as the code becomes self-explanatory.

beorn7 (changes requested):
Looks good in general. I just have a nit about redundant code comments and overly short variable names.

However, there is also a more general concern: My assumption was this function is just used for templating, but it is used for many other purposes, too. It is not unlikely that those other purposes rely on the return never being negative.

Let's play safe for now and better introduce a new function that allows negative durations (e.g. `ParseDurationAllowNegative` or something). It will be easy to wire it to `parseDuration` in the templating.

WDYT?

model/time.go
hautestacky: called `negative`, and deleted comments

iamhalje:
@beorn7 actually has similar question before, i agree. what would you say now? mean what think about `String` function and the tests?

model/time.go
```diff
@@ -253,6 +253,76 @@ func ParseDuration(s string) (Duration, error) {
 			return 0, errors.New("duration out of range")
 		}
 	}
+
+	return Duration(dur), nil
+}
+
+// ParseDurationAllowNegative parses a string into a time.Duration, assuming that a year
+// always has 365d, a week always has 7d, and a day always has 24h.
+// Supporting negative durations as well.
```
beorn7: I think it's fine to not repeat the whole doc comment from `ParseDuration` but do something like the following:

```suggestion
// ParseDurationAllowNegative works like ParseDuration but it allows
// negative durations as well.
```

And maybe, for clarity, add to the original `ParseDuration` doc comment that negative values will cause an error. (I was confused myself about that.)

model/time.go
```diff
@@ -253,6 +253,76 @@ func ParseDuration(s string) (Duration, error) {
...
+func ParseDurationAllowNegative(s string) (Duration, error) {
```
beorn7, line 262: How about implementing the whole function in this way:

```Go
func ParseDurationAllowNegative(s string) (Duration, error) {
	if s[0] != '-' {
		return ParseDuration(s)
	}
	d, err := ParseDuration(s[1:])
	return -d, err
}
```

model/time.go
```diff
@@ -265,6 +335,11 @@ func (d Duration) String() string {
 		return "0s"
 	}
 
+	negative := ms < 0
+	if negative {
+		ms = -ms
+	}
```
beorn7: Slightly longer here, but it saves logic below:

```suggestion
	sign := ""
	if ms < 0 {
		sign = "-"
		ms *= -1
	}
```

model/time.go
```diff
@@ -286,6 +361,10 @@ func (d Duration) String() string {
 	f("s", 1000, false)
 	f("ms", 1, false)
 
+	if negative {
+		return "-" + r
+	}
+
 	return r
```
beorn7: This goes along with above suggestion:

```suggestion
	return sign + r
```

iamhalje:
@beorn7 i really like everything, thanks, and i also improved TestParseDuration function in tests

model/time.go
```diff
@@ -253,18 +254,43 @@ func ParseDuration(s string) (Duration, error) {
...
 		}
 	}
+
 	return Duration(dur), nil
 }
 
+// ParseDurationAllowNegative is like ParseDuration but also accepts negative durations.
+func ParseDurationAllowNegative(s string) (Duration, error) {
+	if s == "" {
+		return 0, errors.New("empty duration string")
```
beorn7: Good catch that we have to check for empty string before checking for the sign.

But I would still leave the error handling to the original function. See my next comment.

model/time.go
```diff
@@ -253,18 +254,43 @@ func ParseDuration(s string) (Duration, error) {
...
+	}
+
+	if s[0] != '-' {
```
beorn7: Let's add the condition in L263 here:

```suggestion
	if s == "" || s[0] != '-' {
```

model/time.go
```diff
@@ -253,18 +254,43 @@ func ParseDuration(s string) (Duration, error) {
...
+		return ParseDuration(s)
+	}
+
+	d, err := ParseDuration(s[1:])
+	if err != nil {
+		return 0, err
+	}
```
beorn7: I think we don't need this. `ParseDuration` will already return `0, err` in case of an error, so returning `-0, err` is equivalent.

iamhalje:
@beorn7 LGTY?

beorn7 (approved):
Looks great. I'll also tag a new release so that you can use this change in https://github.com/prometheus/prometheus/pull/16619

PR #794, opened 2025-06-17
remove gopkg.in/yaml.v2

miparnisari:
https://github.com/go-yaml/yaml/tree/v3 was archived on April 1, 2025.

I am replacing it with https://github.com/kubernetes-sigs/yaml.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -21,7 +21,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"gopkg.in/yaml.v2"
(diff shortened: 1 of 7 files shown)


boekkooi-impossiblecloud:
Good day,
I was just wondering why one would use https://github.com/kubernetes-sigs/yaml over https://github.com/yaml/go-yaml :smile: 
A related discussion also happened over at https://github.com/stretchr/testify/issues/1724 but they didn't look into `github.com/kubernetes-sigs/yaml` so I was just wondering :smile:

PR #797, opened 2025-06-27
feat(secrets): Introduce new package for dynamic secret management

hsmatulis:
Draft PR adding secrets management core library to be used with prometheus and alertmanager. See the [proposa herel](https://github.com/prometheus/proposals/blob/main/proposals/0047-secret_providers.md))

### DIFF
--- a/secrets/field.go
+++ b/secrets/field.go
@@ -1,0 +1,115 @@
+package secrets
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+// SecretField is a field containing a secret.
+type SecretField struct {
+	provider Provider
+	manager  **Manager
+}
+
+func (s SecretField) String() string {
+	return fmt.Sprintf("SecretField{Provider: %s}", s.provider.Name())
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for SecretField.
+func (s SecretField) MarshalYAML() (interface{}, error) {
+	if s.provider.Name() == "inline" && s.manager != nil && (*s.manager).MarshalInlineSecrets {
+		return s.Get(), nil
+	}
+	out := make(map[string]interface{})
+	out[s.provider.Name()] = s.provider
+	return out, nil
+}
+
+// MarshalJSON implements the json.Marshaler interface for SecretField.
+func (s SecretField) MarshalJSON() ([]byte, error) {
+	data, err := s.MarshalYAML()
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(data)
+}
+
+// providerBase is used to extract the type of the provider.
+type providerBase = map[string]map[string]interface{}
+
+func (s *SecretField) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var plainSecret string
+	if err := unmarshal(&plainSecret); err == nil {
+		s.provider = &InlineProvider{
+			secret: plainSecret,
+		}
+	}
+
+	var base providerBase
+	if err := unmarshal(&base); err != nil {
+		return err
+	}
+
+	if len(base) != 1 {
+		return fmt.Errorf("secret must contain exactly one provider type, but found %d.", len(base))
+	}
+
+	var name string
+	var providerConfig map[string]interface{}
+	for providerType, data := range base {
+		name = providerType
+		providerConfig = data
+		break
+	}
+
+	concreteProvider, err := Providers.Get(name)
+	if err != nil {
+		return err
+	}
+
+	configBytes, err := yaml.Marshal(providerConfig)
+	if err != nil {
+		return fmt.Errorf("failed to re-marshal config for %s provider: %w", name, err)
+	}
+
+	if err := yaml.Unmarshal(configBytes, &concreteProvider); err != nil {
+		return fmt.Errorf("failed to unmarshal into %s provider: %w", name, err)
+	}
+
+	s.provider = concreteProvider
+	return nil
+}
+
+// SetSecretValidation registers an optional validation function for the secret.
+//
+// When the secret manager fetches a new version of the secret, it will not
+// be used immediately if there is a validator. Instead, the manager will
+// hold the new secret in a pending state and call the provided Validate
+// with it until it returns true, there is an explicit refresh request,
+// there is a time out, or the old secret was never valid.
+func (s *SecretField) SetSecretValidation(validator SecretValidator) {
+	if *s.manager == nil {
+		panic("secret fields can only be used when a manager is set")
+	}
+	(*s.manager).setSecretValidation(s, validator)
+}
+
+func (s *SecretField) Get() string {
+	if *s.manager == nil {
+		panic("secret fields can only be used when a manager is set")
+	}
+	return (*s.manager).get(s)
+}
+
+func (s *SecretField) TriggerRefresh() {
+	if *s.manager == nil {
+		panic("secret fields can only be used when a manager is set")
+	}
+	(*s.manager).triggerRefresh(s)
+}
+
+func (s *SecretField) providerId() string {
+	return fmt.Sprintf("%s/%s", s.provider.Name(), s.provider.Key())
+}
--- a/secrets/manager.go
+++ b/secrets/manager.go
@@ -1,0 +1,245 @@
+package secrets
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+const (
+	// fetchTimeout governs the maximum time a single fetch attempt can take.
+	fetchTimeout = 5 * time.Minute
+	// fetchInitialBackoff is the initial backoff duration for refetching a secret after a failure.
+	fetchInitialBackoff = 1 * time.Second
+	// fetchMaxBackoff is the maximum backoff duration for retrying a failed fetch.
+	fetchMaxBackoff = 2 * time.Minute
+
+	// defaultValidateInterval is the default interval at which to check for pending secrets that need validation.
+	defaultValidateInterval = 10 * time.Second
+	// validationTimeout governs the maximum time a single validation attempt can take.
+	validationTimeout = 30 * time.Second
+	// validationInitialBackoff is the initial backoff duration for re-validating a secret after a failure.
+	validationInitialBackoff = 1 * time.Second
+	// validationMaxBackoff is the maximum backoff duration for retrying a failed validation.
+	validationMaxBackoff = 30 * time.Second
+	// validationMaxRetries is the maximum number of retries for a failed validation.
+	validationMaxRetries = 10
+)
+
+type Manager struct {
+	MarshalInlineSecrets bool
+	mtx                  sync.RWMutex
+	secrets              map[string]*managedSecret
+	refreshRequested     chan struct{}
+	cancel               context.CancelFunc
+}
+
+type managedSecret struct {
+	mtx              sync.RWMutex
+	pendingSecret    string
+	provider         Provider
+	fetched          time.Time
+	fetchInProgress  bool
+	refreshInterval  time.Duration
+	refreshRequested bool
+	fields           map[*SecretField]*managedSecretField
+}
+
+type managedSecretField struct {
+	validator SecretValidator
+	verified  bool
+	secret    string
+}
+
+func NewManager(config interface{}) *Manager {
+	// TODO: populate secrets from config using resolve.go
+	return &Manager{
+		secrets: make(map[string]*managedSecret),
+	}
+}
+
+func (m *Manager) Start(ctx context.Context) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	go func() {
+		m.fetchSecretsLoop(ctx)
+		cancel()
+	}()
+
+	m.cancel = cancel
+}
+
+func (m *Manager) Stop() {
+	m.cancel()
+	// TODO: wait for output
+}
+
+func (m *Manager) setSecretValidation(s *SecretField, validator SecretValidator) {
+	m.mtx.RLock()
+	secret := m.secrets[s.providerId()]
+	m.mtx.RUnlock()
+	secret.mtx.Lock()
+	secret.fields[s].validator = validator
+	secret.mtx.Unlock()
+}
+
+func (m *Manager) get(s *SecretField) string {
+	if inline, ok := s.provider.(*InlineProvider); ok {
+		return inline.secret
+	}
+	m.mtx.RLock()
+	secret := m.secrets[s.providerId()]
+	m.mtx.RUnlock()
+	secret.mtx.RLock()
+	defer secret.mtx.RUnlock()
+	return secret.fields[s].secret
+}
+
+func (m *Manager) triggerRefresh(s *SecretField) {
+	m.mtx.RLock()
+	secret := m.secrets[s.providerId()]
+	m.mtx.RUnlock()
+	secret.mtx.Lock()
+	defer secret.mtx.Unlock()
+	secret.refreshRequested = true
+	secret.fields[s].verified = false
+	secret.fields[s].secret = secret.pendingSecret
+	m.refreshRequested <- struct{}{}
+}
+
+// fetchSecretsLoop is a long-running goroutine that periodically fetches secrets.
+func (m *Manager) fetchSecretsLoop(ctx context.Context) {
+	timer := time.NewTimer(5 * time.Minute)
+	defer timer.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+		case <-m.refreshRequested:
+			if !timer.Stop() {
+				<-timer.C
+			}
+		}
+		m.mtx.RLock()
+		// Create a list of secrets to check to avoid holding the lock during fetch operations.
+		secretsToCheck := make([]*managedSecret, 0, len(m.secrets))
+		for _, secret := range m.secrets {
+			secretsToCheck = append(secretsToCheck, secret)
+		}
+		m.mtx.RUnlock()
+
+		waitTime := 5 * time.Minute
+
+		for _, ms := range secretsToCheck {
+			ms.mtx.Lock()
+
+			refreshNeeded := ms.refreshRequested || time.Since(ms.fetched) > ms.refreshInterval
+			timeToRefresh := time.Until(ms.fetched.Add(ms.refreshInterval))
+
+			if ms.fetchInProgress {
+				ms.mtx.Unlock()
+				continue
+			}
+
+			if !refreshNeeded {
+				waitTime = min(timeToRefresh, waitTime)
+				ms.mtx.Unlock()
+				continue
+			}
+			ms.fetchInProgress = true
+			ms.mtx.Unlock()
+
+			go m.fetchAndStoreSecret(ctx, ms)
+		}
+		timer.Reset(waitTime)
+
+	}
+}
+
+// fetchAndStoreSecret performs a single secret fetch, including retry logic with exponential backoff.
+// It is robust against hangs in the underlying provider's FetchSecret method.
+func (m *Manager) fetchAndStoreSecret(ctx context.Context, ms *managedSecret) {
+	var newSecret string
+	var err error
+	ms.mtx.Lock()
+	provider := ms.provider
+	ms.mtx.Unlock()
+
+	backoff := fetchInitialBackoff
+	for {
+		fetchCtx, cancel := context.WithTimeout(ctx, fetchTimeout)
+
+		newSecret, err = provider.FetchSecret(fetchCtx)
+		cancel()
+
+		if err == nil {
+			break // Success
+		}
+
+		select {
+		case <-time.After(backoff):
+			backoff = min(fetchMaxBackoff, backoff*2)
+		case <-ctx.Done():
+			return
+		}
+
+	}
+
+	ms.mtx.Lock()
+	defer ms.mtx.Unlock()
+
+	ms.pendingSecret = newSecret
+	ms.fetched = time.Now()
+	ms.fetchInProgress = false
+	ms.refreshRequested = false
+
+	for _, field := range ms.fields {
+		// If a was not verified before, we can swap it immediately
+		if field.verified == false {
+			field.secret = newSecret
+		}
+		field.verified = false
+		go m.validateAndStoreField(ctx, ms, field, newSecret)
+	}
+}
+
+// validateAndStoreField performs validation for a single field, including retry logic.
+func (m *Manager) validateAndStoreField(ctx context.Context, ms *managedSecret, field *managedSecretField, pendingSecret string) {
+	var isValid bool
+
+	ms.mtx.RLock()
+	validator := field.validator
+	ms.mtx.RUnlock()
+
+	backoff := validationInitialBackoff
+	for i := range validationMaxRetries {
+
+		validateCtx, cancel := context.WithTimeout(ctx, validationTimeout)
+		isValid = validator == nil || validator.Validate(validateCtx, pendingSecret)
+		cancel()
+
+		if isValid {
+			break // Success
+		}
+
+		if i < validationMaxRetries-1 {
+			select {
+			case <-time.After(backoff):
+				backoff = min(validationMaxBackoff, backoff*2)
+			case <-ctx.Done():
+				return
+			}
+		}
+	}
+
+	ms.mtx.Lock()
+	defer ms.mtx.Unlock()
+
+	if ms.pendingSecret != pendingSecret {
+		return
+	}
+
+	field.secret = pendingSecret
+	field.verified = true
+}
--- a/secrets/provider.go
+++ b/secrets/provider.go
@@ -1,0 +1,26 @@
+package secrets
+
+import (
+	"context"
+)
+
+type Provider interface {
+	// FetchSecret retrieves the secret value.
+	FetchSecret(ctx context.Context) (string, error)
+
+	// Name returns the provider's name (e.g., "inline").
+	Name() string
+
+	// Key returns a stable identifier for the secret being fetched.
+	// It is combined with Name() to deduplicate fetch calls.
+	Key() string
+}
+
+// SecretValidator allows for validating a new secret before it is
+// rotated into active use. If invalid, the old secret will be used
+// while it is still considered valid and has not expired.
+// This interface is optional, to prevent monitoring gaps if the
+// system being scraped hasn't had its secret refreshed yet.
+type SecretValidator interface {
+	Validate(ctx context.Context, secret string) bool
+}
--- a/secrets/resolve.go
+++ b/secrets/resolve.go
@@ -1,0 +1,90 @@
+package secrets
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+)
+
+var secretFieldType = reflect.TypeOf(SecretField{})
+
+type SecretPaths map[string]SecretField
+
+type walkItem struct {
+	path []string
+	val  reflect.Value
+}
+
+func getSecretFields(v interface{}) (SecretPaths, error) {
+	results := make(SecretPaths)
+	if v == nil {
+		return results, nil
+	}
+	visited := make(map[uintptr]bool)
+	queue := []walkItem{{path: nil, val: reflect.ValueOf(v)}}
+
+	for len(queue) > 0 {
+		currentItem := queue[0]
+		queue = queue[1:]
+
+		path := currentItem.path
+		val := currentItem.val
+		if len(path) > 50 {
+			return nil, fmt.Errorf("path traversal exceeded maximum depth (current depth: %d)", len(path))
+		}
+
+		if val.Type() == secretFieldType {
+			path := strings.Join(path, ".")
+			secret, ok := val.Interface().(SecretField)
+			if !ok {
+				return nil, fmt.Errorf("path '%s': internal error: matched SecretField type but failed type assertion", path)
+			}
+			results[path] = secret
+			continue
+		}
+		queue = process(path, val, visited, queue)
+	}
+	return results, nil
+}
+
+func process(path []string, val reflect.Value, visited map[uintptr]bool, queue []walkItem) []walkItem {
+	if !val.IsValid() {
+		return queue
+	}
+	switch val.Kind() {
+	case reflect.Ptr:
+		if val.IsNil() || visited[val.Pointer()] {
+			return queue
+		}
+		visited[val.Pointer()] = true
+		return append(queue, walkItem{path: path, val: val.Elem()})
+	case reflect.Interface:
+		return append(queue, walkItem{path: path, val: val.Elem()})
+	case reflect.Struct:
+		for i := 0; i < val.NumField(); i++ {
+			newPath := append(path, val.Type().Field(i).Name)
+			field := val.Field(i)
+			if field.CanInterface() {
+				queue = append(queue, walkItem{path: newPath, val: field})
+			}
+		}
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < val.Len(); i++ {
+			newPath := append(path, fmt.Sprintf("[%d]", i))
+			queue = append(queue, walkItem{path: newPath, val: val.Index(i)})
+		}
+	case reflect.Map:
+		keys := val.MapKeys()
+		sort.Slice(keys, func(i, j int) bool {
+			return fmt.Sprintf("%v", keys[i].Interface()) < fmt.Sprintf("%v", keys[j].Interface())
+		})
+		for _, key := range keys {
+			keyPath := append(path, fmt.Sprintf("[%v:key]", key.Interface()))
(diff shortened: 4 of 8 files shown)


secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"go.yaml.in/yaml/v2"
+)
+
+// SecretField is a field containing a secret.
+type SecretField struct {
```
bwplotka: readability nit: We have opportunity to "eat" `secret` prefix, since it's a `secrets` package, so everyone will use it generally as `secrets.SecretField`. So we could do `secrets.Field` if we want. To consider, not a blocker.

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+	provider  Provider
+	manager   *Manager
+	validator SecretValidator
+	settings  SecretFieldSettings
+}
+
+type SecretFieldSettings struct {
+	RefreshInterval time.Duration `yaml:"refreshInterval,omitempty"`
+	Default         string        `yaml:"default,omitempty"`
```
bwplotka: Is this used? I don't see any tests for it. What this default use case should look like?

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+// SecretField is a field containing a secret.
+type SecretField struct {
+	provider  Provider
+	manager   *Manager
+	validator SecretValidator
+	settings  SecretFieldSettings
+}
+
+type SecretFieldSettings struct {
+	RefreshInterval time.Duration `yaml:"refreshInterval,omitempty"`
```
bwplotka: question: Hm, good question if refresh is really relevant to all providers... E.g it's not for inline and.. also not for file based? Refreshing in general is a key concept here so we at least needs a good commentary in the code for this field on how it suppose to be implemented and consumed?

secrets/resolve_test.go
```diff
@@ -0,0 +1,384 @@
...
+			// Check for expected error.
+			if tc.errContains != "" {
+				if gotErr == nil {
+					t.Fatalf("Expected error containing '%s', but got no error", tc.errContains)
+				}
+				if !strings.Contains(gotErr.Error(), tc.errContains) {
+					t.Errorf("Expected error containing '%s', but got: %v", tc.errContains, gotErr)
+				}
+				return
+			} else if gotErr != nil {
```
bwplotka: Generally no need for "else" if you return above.

secrets/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+## How to Use
+
+Using the `secrets` package involves three main steps: defining your configuration struct, initializing the secret manager, and accessing the secret values.
+
+### 1. Define Your Configuration Struct
+
+In your configuration struct, use the `secrets.SecretField` type for any fields that should contain secrets.
+
+```go
```
bwplotka: Sustainability suggestion (non-blocking):

This is well documented, thanks!

However, it might be even better if we would replace the "How to use" section with a single or set of buildable examples as per https://go.dev/blog/examples .. and commentary in the key public code structures.

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+// MarshalJSON implements the json.Marshaler interface for SecretField.
+func (s SecretField) MarshalJSON() ([]byte, error) {
+	data, err := s.MarshalYAML()
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(data)
+}
+
+type mapType = map[string]interface{}
```
bwplotka: nit: `interface{}` is deprecated in Go, `any` is prefered (same thing, new name).
```suggestion
type mapType = map[string]any
```

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+
+// splitProviderAndSettings separates provider-specific configuration from the generic SecretField settings.
+func splitProviderAndSettings(baseMap mapType) (baseProvider Provider, providerData interface{}, settingsMap mapType, err error) {
+	settingsMap = make(mapType)
+	providerName := ""
+
+	for k, v := range baseMap {
+		// Check if the key corresponds to a registered provider.
+		if p, _ := Providers.Get(k); p != nil {
```
bwplotka: Using globals is generally not ideal - sooner or later we hit blockers  e.g. in testing or libA importing common/secrets and libB importing common/secrets for totally different configs and libB wants to ensure different set of providers are unrelated configs, so can we at least:

1. Use global in few, well documented places, and rest of code is expecting global things to be locally passed as arguments if possible. For example `splitProviderAndSettings(providers X, baseMap mapType)` is trivial to do here and will save us from problems later on.
2. Document limitations e.g. the fact there can be only one set of providers for all configs in a single binary (perhaps fine for now).

In separate thread we could discuss if globals are even needed in common (it's a library, if ever we do globals, doing them on a caller side would be much safer e.g. only on Prometheus side).

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+
+	"go.yaml.in/yaml/v2"
+)
+
+// SecretField is a field containing a secret.
+type SecretField struct {
+	provider  Provider
+	manager   *Manager
+	validator SecretValidator
```
bwplotka: I wonder if we don't hit [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here.  I understand the intention, but do we want Prometheus code to do extra fake network calls only to check if secret works? Do we do that in Prometheus for every changed secret in a file path? 🤔 

If not we could add this feature later and make this PR a bit smaller?

secrets/provider.go
```diff
@@ -0,0 +1,69 @@
...
+// rotated into active use. If invalid, the old secret will be used
+// while it is still considered valid and has not expired.
+// This interface is optional, to prevent monitoring gaps if the
+// system being scraped hasn't had its secret refreshed yet.
+type SecretValidator interface {
+	Validate(ctx context.Context, secret string) bool
+	Settings() ValidationSettings
+}
+
+// ValidationSettings holds configurable parameters for secret validation.
```
bwplotka: Hm.. this is even bigger YAGNI, because it assumes all validators will need timeouts and backoffs. I wonder if we could:

1. Scope this down and remove this feature for now
2. At least decompose validation and especially timeout based validation to separate structs not even in this package (composability).

secrets/field_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.yaml.in/yaml/v2"
+)
+
+func TestSecretField_UnmarshalYAML(t *testing.T) {
```
bwplotka: Love it (both tests and YAML surface usability, so how users can construct YAML here 👍🏽 )

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+// with it until it returns true, there is an explicit refresh request,
+// there is a time out, or the old secret was never valid.
+func (s *SecretField) SetSecretValidation(validator SecretValidator) {
+	s.validator = validator
+	if s.manager != nil {
+		s.manager.setSecretValidation(s, validator)
+	}
+}
+
+func (s *SecretField) Get() string {
```
bwplotka: Nice! This make caller unaware of errors which can be both great and bad, depending what we do. I think it would be great to test out in this state 👍🏽 

One alternative to make things leaner on SecretField (decompose responsibilities and state), would be if SecretField would allow storing (atomically) and getting secret value string. This way we could have common package global free and SecretField manager free (manager could be responsible to set correct things in config).

It could be even provider agnostic/free combined with https://github.com/prometheus/common/pull/797/files#r2414589581 idea (and avoid globals in this package)

secrets/resolve.go
```diff
@@ -0,0 +1,112 @@
...
+var secretFieldType = reflect.TypeOf(SecretField{})
+
+type secretPaths map[string]*SecretField
+
+type walkItem struct {
+	path []string
+	val  reflect.Value
+}
+
+func getSecretFields(v interface{}) (secretPaths, error) {
```
bwplotka: Love it, epic! A go idiomatic way to find secret fields automatically and do something about it 👍🏽

secrets/manager.go
```diff
@@ -0,0 +1,413 @@
...
+		secrets: make(map[*SecretField]*managedSecret),
+	}
+	manager.registerMetrics(r)
+	for path, field := range paths {
+		manager.registerSecret(path, field)
+	}
+	return manager, nil
+}
+
+func (m *Manager) registerMetrics(r prometheus.Registerer) {
```
bwplotka, line 129: Thanks for metrics, super important!

secrets/manager.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+	m.mtx.RLock()
+	secret := m.secrets[s]
+	m.mtx.RUnlock()
+	secret.mtx.RLock()
+	defer secret.mtx.RUnlock()
+	return secret.secret
+}
+
+func (m *Manager) triggerRefresh(s *SecretField) {
```
bwplotka: note to myself: Manager code needs deeper review, quite many locks 🙈 prone to errors, but perhaps needed.

secrets/field.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to re-marshal config: %w", err)
+	}
+	if err := yaml.Unmarshal(bytes, target); err != nil {
+		return fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+	return nil
+}
+
+func (s *SecretField) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
bwplotka: Have we considered doing plainSecret thingy, but then unmarshal to `any` or `map[string]any` and let manager.ApplyConfig do the final parsing? This would make SecretField leaner and manager/provider agnostic (together with setter/getter for secret values https://github.com/prometheus/common/pull/797/files#r2414554264)

secrets/field.go
hsmatulis: Done!

secrets/field.go
hsmatulis: Not used, removed it. Thanks!

secrets/field.go
hsmatulis: Hmm I guess it is not relevant to inline but could be relevant to file. Although we might want  to think about listening to filesystem changes for a file based provider instead, but I found that to complicate the API quite a bit...

secrets/field.go
hsmatulis: You are right, this hits YAGNI for sure. Removed it

secrets/provider.go
hsmatulis: Removed as well

secrets/manager.go
hsmatulis: yeah it was very complicated before with the verify logic, I think it should be a bit simpler now.. Still need some cleanup there, but should be more readable

secrets/manager.go
```diff
@@ -0,0 +1,333 @@
...
+			return nil, err
+		}
+	}
+	return manager, nil
+}
+
+func (m *Manager) registerMetrics(r prometheus.Registerer) {
+	labels := []string{"provider", "secret_id"}
+
+	m.lastSuccessfulFetch = prometheus.NewGaugeVec(
```
bwplotka:
```suggestion
	m.lastSuccessfulFetch = promauto.With(r).NewGaugeVec(
```

Then we can remove r.MustRegisterer

secrets/field.go
```diff
@@ -0,0 +1,168 @@
...
+	"time"
+
+	"go.yaml.in/yaml/v2"
+)
+
+// Field is a field containing a secret.
+type Field struct {
+	providerName   string
+	providerConfig ProviderConfig
+	manager        *Manager
```
bwplotka: idea during 1:1, it would be nice to remove bi-direction and have lazy refresh perhaps

bwplotka: Alternatively just use interface for Refresh()

secrets/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Concepts
+
+The package is built around a few core concepts:
+
+  * `SecretField`: A type used in configuration structs to represent a field that holds a secret. It handles the logic for unmarshaling from different secret sources, and the API for accessing secrets.
+  * `ProviderConfig`: An interface for configuring a secret provider. It includes methods to create a new `Provider` instance and to clone the configuration.
+  * `ProviderConfigId`: An interface for uniquely identifying a `ProviderConfig` instance.
+  * `Provider`: An interface for fetching secrets from a specific source (e.g., inline string, file on disk). The package comes with built-in providers, and new ones can be registered.
+  * `Manager`: A component that discovers all `SecretField` instances within a configuration struct, manages their lifecycle, and handles periodic refreshing of secrets.
```
bwplotka: Can we put those into comments? Especially for library, unlikely anyone will notice README. In go it's either `doc.go` or code structure for godocs.

bwplotka: https://go.dev/blog/godoc

bwplotka: https://go.dev/blog/examples

bwplotka (commented):
Looks solid, thanks!

Still some few comments to address, but overall looks close to be merged!

secrets/resolve_test.go
hsmatulis: thanks! fixed

secrets/field.go
hsmatulis: Fixed! Thanks

secrets/field.go
hsmatulis: Fixed passing in global as arg, but will get rid of the global with the field re-write w/o manager

secrets/field_test.go
hsmatulis: Thanks!:)

secrets/manager.go
hsmatulis: Done, thanks!

secrets/resolve.go
hsmatulis: Thanks! I hope it will be ergonomic

secrets/field.go
hsmatulis: Done!

secrets/manager.go
hsmatulis: Can't be lacking metrics in prometheus!

secrets/field.go
hsmatulis: done!

secrets/field.go
hsmatulis: done! No interface, lazy refresh

secrets/README.md
hsmatulis: done!

secrets/README.md
hsmatulis: done!

secrets/example_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.yaml.in/yaml/v2"
+
+	"github.com/prometheus/common/secrets"
+)
+
+func Example() {
```
bwplotka, line 28: Nice!

nit: We likely want some name for this example to make this wrong, but maybe it works like that too? Happy to merge and learn.

```suggestion
func ExampleSecretsUse() {
```

secrets/example_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		if ready {
+			break
+		}
+		// Prevent a busy loop while waiting for I/O
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	// Access the secret values.
+	apiKey := cfg.APIKey.Get()
```
bwplotka: very cool use!

secrets/README.md
```diff
@@ -0,0 +1,18 @@
+# Secret Management
```
bwplotka, line 1: Nice!

One thing to discuss/put in README.

How it relates to https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/config/config.go#L28

What would be the plan forward? Deprecate `config.Secret`?

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.yaml.in/yaml/v2"
+)
+
+// PolicyFunc returns true if secrets should be printed (insecure),
+// and false if they should be scrubbed (secure).
+type PolicyFunc func() bool
```
bwplotka, line 29: suggestion: We have epic opportunity to deglobalize this and have `Field.redact bool` field with perhaps a DAG method to set it on or all off per config. WDYT?

Also if we do bool, we could just do simpler global as in https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/config/config.go#L33C5-L33C23

bwplotka (commented):
Solid work, did early pass, will go with more review next week, thanks! 💪🏽

secrets/example_test.go
bwplotka: I guess it somehow works (I tried locally).

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+		return "", nil, nil, fmt.Errorf("no valid secret provider found in configuration:\n%s", string(yamlBytes))
+	}
+
+	return providerName, providerData, settingsMap, nil
+}
+
+// convertConfig takes a yaml-parsed any and unmarshals it into a typed struct.
+// It achieves this by first marshalling the input to YAML and then unmarshalling
+// it into the target struct.
+func convertConfig[T any](source any, target T) error {
```
bwplotka: Perhaps more readable?

```suggestion
func unmarshalTo[T any](source any, target T) error {
```

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+	bytes, err := yaml.Marshal(source)
```
bwplotka, line 210: This is interesting, I thought `any` is internally just bytes of unmarshalled blob, sounds like it's not so simple?

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, err
+	}
+	return json.Marshal(data)
+}
+
+type mapType = map[string]any
+
+// splitProviderAndSettings separates provider-specific configuration from the generic SecretField settings.
+func splitProviderAndSettings(provReg *ProviderRegistry, baseMap mapType) (providerName string, providerData interface{}, settingsMap mapType, err error) {
+	settingsMap = make(mapType)
```
bwplotka: nit: We typically in Go just inline map instantiation, no need for `mapType`

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+		// Check if the key corresponds to a registered provider.
+		if _, err := provReg.Get(k); err == nil {
+			if providerName != "" {
+				// A provider has already been found, which is an error.
+				return "", nil, nil, fmt.Errorf("secret must contain exactly one provider type, but multiple were found: %s, %s", providerName, k)
+			}
+			providerName = k
+			providerData = v
+		} else {
+			// If it's not a provider key, treat it as a setting.
```
bwplotka: For readability (was confused reading this code)

```suggestion
			// If it's not a provider key, treat it as a setting. We will detect missing provider later.
```

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+			settingsMap[k] = v
```
bwplotka: Does this mean settings fields can conflict with provider name. Is that a problem? Do we need settings? Example does  not show it. Can we apply YAGNI to descope (add it later when needed?)

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+			value: plainSecret,
+		}, nil
+	}
+
+	var baseMap mapType
+	if err := convertConfig(s.rawConfig, &baseMap); err != nil {
+		return nil, err
+	}
+
+	providerName, providerConfigData, settingsMap, err := splitProviderAndSettings(reg, baseMap)
```
bwplotka: nit: Wouldn't this be simpler if `splitProviderAndSettings` was returning `providerConfig`?

secrets/field.go
bwplotka: I wouldn't also mind if we kill `FieldSettings` and duplicate refresh and other logic in the provider. We can share common things once we have at least two complex providers.. (:

bwplotka: I guess to avoid/handle dups one could try to unmarshal `FieldSettings` here? Anyway just a bit pain to add more fields, you never know if it conflicts with dynamic provider someone might implement in fork, perhaps easier to inline with provider?

secrets/field.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+type FieldSettings struct {
+	RefreshInterval time.Duration `yaml:"refreshInterval,omitempty"`
+}
+
+func (fs *fieldState) id() string {
+	fs.mutex.Lock()
+	defer fs.mutex.Unlock()
+	providerID := fs.path
```
bwplotka: It's has nothing to do with provider (?) was a bit confused.

```suggestion
	id := fs.path
```

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+	labels := prometheus.Labels{
+		"provider":  state.providerName,
+		"secret_id": state.id(),
+	}
+
+	if state.settings.RefreshInterval == 0 {
+		state.settings.RefreshInterval = defaultRefreshInterval
+	}
+
+	if _, exists := m.secrets[state.id()]; !exists {
```
bwplotka: Any chances to guard this map from a single manager used on totally different unmarshalled configurations? Or even different config types? 

Perhaps save a pointer of passed config or something? We can add TODO, too.

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	state, err := s.parseRawConfig(m.providers, path)
+	if err != nil {
+		return err
+	}
+
+	labels := prometheus.Labels{
+		"provider":  state.providerName,
```
bwplotka: Why we need provider? Isn't ID already giving you that?

bwplotka: Ah, it's for metrics. That's ok for now, but it's a bit of duplicated info, maybe `path`, `provider` would be a better way or even without provider. We can change later.

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+		m.secrets[state.id()] = ms
+		m.secretState.With(labels).Set(stateInitializing)
+		m.fetchSuccessTotal.With(labels).Add(0)
+		m.fetchFailuresTotal.With(labels).Add(0)
+	}
+	m.secrets[state.id()].mtx.Lock()
+	defer m.secrets[state.id()].mtx.Unlock()
+
+	m.secrets[state.id()].refreshInterval = min(
```
bwplotka: Can we add comment? not sure why we pick min of old and updated interval (how I read this now)

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+		if err != nil {
+			return err
+		}
+		ms := &managedSecret{
+			refreshInterval: state.settings.RefreshInterval,
+			metricLabels:    labels,
+			provider:        provider,
+		}
+		m.secrets[state.id()] = ms
+		m.secretState.With(labels).Set(stateInitializing)
```
bwplotka, line 217: Another way would be to get `m.secretState.With(labels)` (and other metrics) into managedSecret.metrics for direct access. It's reusable, we can then kill `managedSecret.labels`

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+		provider, err := state.config.NewProvider()
+		if err != nil {
+			return err
+		}
+		ms := &managedSecret{
+			refreshInterval: state.settings.RefreshInterval,
+			metricLabels:    labels,
+			provider:        provider,
+		}
+		m.secrets[state.id()] = ms
```
bwplotka: Can we do `id()` once and save local var?

secrets/manager.go
```diff
@@ -0,0 +1,367 @@
...
+		m.secretState.With(labels).Set(stateInitializing)
+		m.fetchSuccessTotal.With(labels).Add(0)
+		m.fetchFailuresTotal.With(labels).Add(0)
```
bwplotka, line 219: return early?

secrets/example_test.go
```diff
@@ -0,0 +1,100 @@
...
+`, passwordFile.Name()))
+
+	var cfg MyConfig
+	if err := yaml.Unmarshal(configData, &cfg); err != nil {
+		panic(fmt.Errorf("error unmarshaling config: %w", err))
+	}
+
+	// Create a secret manager. This discovers and manages all Fields in cfg.
+	// The manager will handle refreshing secrets in the background.
+	manager, err := secrets.NewManager(promRegisterer, secrets.Providers, &cfg)
```
bwplotka: question: What's the process for refreshing secrets ON reconfiguration (very common in Prometheus -- `ApplyConfig`). It sounds like with this abstraction we need to recreate manager? (`secrets.NewManager` has to be re-run?)

secrets/example_test.go
bwplotka: suggestion: Although now I noticed that in order to use `Get() string` you need to do `manager.SecretsReady` 🤔 ? Why not combining all into `Get`?

secrets/field.go
bwplotka: Again, comment would be nice. I am worried it's YAGNI - we don't know if refresh in such a form would be even useful for other providers (we are guessing). Perhaps moving this to inside provider config is safer?

hsmatulis: Good challenge, that was unclear. It’s actually essential for the file provider (to support file rotations) and any future network providers with expiring secrets (like Vault).

Since the Manager owns the polling loop defining the interval here allows us to control scheduling centrally without forcing every provider to implement their own timer logic. I've added a clarifying comment as requested

secrets/manager.go
hsmatulis: Cleaned up manager code

secrets/field.go
hsmatulis: Renamed, thanks!

secrets/field.go
hsmatulis: Done!

secrets/field.go
hsmatulis: Done! Thanks

secrets/field.go
hsmatulis: I've moved these to their own struct to prevent collisions, so now it'd be like this:

```
file:
  path: /path/to/secret
 
options:  # Explicitly reserved in the registry
  refresh_interval: 5m
  ```

File: secrets/field.go (hunk 28 above)
Comment (hsmatulis) [reply secrets/field.go]: Done

File: secrets/field.go (hunk 29 above)
Comment (hsmatulis) [reply secrets/field.go]: Fixed!

File: secrets/manager.go (hunk 33 above)
Comment (hsmatulis) [reply secrets/manager.go:217]: Would this still work if we need to drop labels? Reference code

```
		m.lastSuccessfulFetch.Delete(secret.metricLabels)
		m.secretState.Delete(secret.metricLabels)
		m.fetchSuccessTotal.Delete(secret.metricLabels)
		m.fetchFailuresTotal.Delete(secret.metricLabels)
		m.fetchDuration.Delete(secret.metricLabels)
```

File: secrets/manager.go (hunk 34 above)
Comment (hsmatulis) [reply secrets/manager.go]: done!

File: secrets/manager.go (hunk 35 above)
Comment (hsmatulis) [reply secrets/manager.go:219]: Can't return early anymore as we do an optimistic fetch on startup

File: secrets/example_test.go (hunk 36 above)
Comment (hsmatulis) [reply secrets/example_test.go]: The ApplyConfig method is PopulateConfig. It has a special name as, unlike other apply config methods, it will mutate the config for others

File: secrets/manager.go (hunk 31 above)
Comment (hsmatulis) [reply secrets/manager.go]: Changed it to provider & path

File: secrets/manager.go (hunk 30 above)
Comment (hsmatulis) [reply secrets/manager.go]: The ID already has the config type baked in no matter what, so different types would never collide. I switched to a pointer for the default deduplication ID

File: secrets/manager.go (hunk 32 above)
Comment (hsmatulis) [reply secrets/manager.go]: Added!

File: secrets/field.go (hunk 24 above)
Comment (hsmatulis) [reply secrets/field.go:210]: Any is either a string or a map[string]any. Added some comments

File: secrets/README.md (hunk 21 above)
Comment (hsmatulis) [reply secrets/README.md:1]: Added!

File: secrets/field.go (hunk 22 above)
Comment (hsmatulis) [reply secrets/field.go:29]: Resolved per our offline convo: this design is so that the current lib can follow `config.MarshalSecretValue` and avoid cyclical dependencies

File: secrets/manager.go [hunk 37]
```diff
@@ -0,0 +1,437 @@
...
+	}
+	if err := errors.Join(errs...); err != nil {
+		return err
+	}
+	for secretID := range droppedSecrets {
+		secret := m.secrets[secretID]
+		secret.mtx.Lock()
+		if secret.fetchCancel != nil {
+			secret.fetchCancel()
+		}
```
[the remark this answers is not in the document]
hsmatulis: It is atomic due to the mutex guard

secrets/field.go
```diff
@@ -0,0 +1,307 @@
...
+// SetVisibilityPolicy sets the global function that determines if secrets
+// should be printed or scrubbed.
+//
+// This is designed to be called exactly ONCE, typically by the
+// prometheus/common/config package during initialization.
+func SetVisibilityPolicy(p PolicyFunc) {
+	if !policyInitialized.CompareAndSwap(false, true) {
+		panic("prometheus/common/secrets: duplicate initialization of secret visibility policy.\n" +
+			"\tReason: The secret scrubbing policy is a global singleton that must be consistent across the application.\n" +
+			"\tExpected Owner: Only 'prometheus/common/config' should initialize this policy.\n",
```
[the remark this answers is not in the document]
hsmatulis: fixed

secrets/mock.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+
+	// Release if the test requires it, to signal fetch has started.
+	if mp.releaseChan != nil {
+		close(mp.releaseChan)
+	}
+
+	mp.mtx.RLock()
+	defer mp.mtx.RUnlock()
```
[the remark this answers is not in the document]
hsmatulis: done

secrets/provider.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// ProviderConfigID is an interface for uniquely identifying a ProviderConfig
+// instance.
+//
+// The ID method should return a string that is unique for each provider
+// configuration. This is used by the Manager to identify and manage secrets.
+type ProviderConfigID interface {
+	ProviderConfig
+	// ID returns a string that is equal iff the two provider configs output the same thing.
```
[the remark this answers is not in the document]
hsmatulis: fixed

secrets/manager.go
```diff
@@ -0,0 +1,437 @@
...
+	var startFetch, endFetch time.Time
+	for {
+		startFetch = time.Now()
+		fetchCtx, cancel := context.WithTimeout(ctx, fetchTimeout)
+		endFetch = time.Now()
+
+		newSecret, err = provider.FetchSecret(fetchCtx)
+		cancel()
+
+		if err == nil {
```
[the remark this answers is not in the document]
hsmatulis: fixed

secrets/resolve_test.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	return normalized
+}
+
+type SimpleStruct struct {
+	Secret Field
+	Day    string
+}
+
+type PrtStruct struct {
```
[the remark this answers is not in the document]
hsmatulis: done

secrets/manager.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+	manager := &Manager{
+		secrets:  make(map[string]*managedSecretProvider),
+		refreshC: make(chan struct{}, 1),
+		logger:   logger,
+		// blocks waiting for up to newFetchTimeout.
+		registerFetchTimeout: newFetchTimeout,
+		providers:            providers,
+	}
+	promslog.NewNopLogger()
```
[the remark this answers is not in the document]
hsmatulis: done

secrets/field.go
```diff
@@ -0,0 +1,338 @@
...
+		providerName: providerName,
+		config:       providerConfig,
+		settings:     settings,
+		value:        settings.DefaultValue,
+	}, nil
+}
+
+func (s *Field) IsNil() bool {
+	return s.rawConfig == nil
+}
```
bwplotka, line 285:
```suggestion
func (s *Field) IsEmpty() bool {
	return s.IsZero()
}
```

secrets/field.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (s *Field) Set(providerName string, config ProviderConfig) error {
```
bwplotka, line 287:
```suggestion
// SetConfig is setting ...
func (s *Field) SetConfig(providerName string, config ProviderConfig) error {
```

PR #798, opened 2025-07-01
fix(promslog): always print time.Duration values as go duration strings

tjhop:
Addresses: prometheus/prometheus#16766

As brought up in the linked issue, it seems that upstream slog handles
time.Duration values in a sub-optimal way in the JSON handler. To
improve handling of duration values in a consistent way across the
ecosystem, we should ensure that duration values are handled the same
way across all supported log formats (json and logfmt).

This change ensures that any key containing a value that is a
time.Duration is reformatted as a boring string value, explicitly
formatted as a Go duration string (ie, "1d2h3m") by calling the
`String()` method on the duration.

Example test output pre/post patch:

Before:
```
~/go/src/github.com/prometheus/common (main [ U ]) -> go test -v -race ./promslog -run TestDurationValues
=== RUN   TestDurationValues
time=2025-06-30T22:11:01.682-04:00 level=INFO source=slog_test.go:108 msg="duration testing" duration_raw=1m30s duration_string=1m30s

{"time":"2025-06-30T22:11:01.683153372-04:00","level":"INFO","source":"slog_test.go:123","msg":"duration testing","duration_raw":90000000000,"duration_string":"1m30s"}

    slog_test.go:128:
                Error Trace:    /home/tjhop/go/src/github.com/prometheus/common/promslog/slog_test.go:128
                Error:          "{\"time\":\"2025-06-30T22:11:01.683153372-04:00\",\"level\":\"INFO\",\"source\":\"slog_test.go:123\",\"msg\":\"duration testing\",\"duration_raw\":90000000000,\"duration_string\":\"1m30s\"}\n" should not contain "\"duration_raw\":90000000000"
                Test:           TestDurationValues
                Messages:       Expected duration to be output as Go duration string "1m30s", got "90000000000"
--- FAIL: TestDurationValues (0.00s)
FAIL
FAIL    github.com/prometheus/common/promslog   0.018s
FAIL
```

After:

```
~/go/src/github.com/prometheus/common (main [ U ]) -> go test -v -race ./promslog -run TestDurationValues
=== RUN   TestDurationValues
time=2025-06-30T22:13:03.714-04:00 level=INFO source=slog_test.go:108 msg="duration testing" duration_raw=1m30s duration_string=1m30s

{"time":"2025-06-30T22:13:03.714880745-04:00","level":"INFO","source":"slog_test.go:123","msg":"duration testing","duration_raw":"1m30s","duration_string":"1m30s"}

--- PASS: TestDurationValues (0.00s)
PASS
ok      github.com/prometheus/common/promslog   1.014s
```

### DIFF
--- a/promslog/slog_test.go
+++ b/promslog/slog_test.go
@@ -103,8 +103,8 @@ func TestDurationValues(t *testing.T) {
 		logFormat string
 		want      string
 	}{
-		"logfmt_duration_testing": {want: "duration_raw=1m30s", logFormat: "logfmt"},
-		"json_duration_testing":   {want: "\"duration_raw\":\"1m30s\"", logFormat: "json"},
+		"logfmt_duration_testing": {want: "duration_raw=1m30s duration_string=1m30s", logFormat: "logfmt"},
+		"json_duration_testing":   {want: "\"duration_raw\":\"1m30s\",\"duration_string\":\"1m30s\"", logFormat: "json"},
 	}
 
 	for name, tc := range tests {
--- a/promslog/slog_test.go
+++ b/promslog/slog_test.go
@@ -98,36 +98,30 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
 func TestDurationValues(t *testing.T) {
 	var (
 		buf    bytes.Buffer
-		output string
 		dur, _ = time.ParseDuration("1m30s")
-		config = &Config{Writer: &buf}
+		config = &Config{
+			Writer: &buf,
+			Format: NewFormat(),
+		}
 	)
 
-	// Test logfmt.
-	logger := New(config)
-	logger.Info("duration testing", slog.Duration("duration_raw", dur), slog.String("duration_string", dur.String()))
-	output = buf.String()
-	// Print logs for humans to see, if needed.
-	fmt.Println(output)
+	tests := map[string]struct {
+		want      string
+		logFormat string
+	}{
+		"logfmt_duration_testing": {want: "duration_raw=1m30s", logFormat: "logfmt"},
+		"json_duration_testing":   {want: "\"duration_raw\":\"1m30s\"", logFormat: "json"},
+	}
 
-	require.NotContainsf(t, output, "duration_raw=\"90000000000\"", "Expected duration to be output as Go duration string \"1m30s\", got \"90000000000\"")
-
-	buf.Reset()
-
-	// Test json.
-	jsonFmt := NewFormat()
-	_ = jsonFmt.Set("json")
-	config.Format = jsonFmt
-
-	logger = New(config)
-	logger.Info("duration testing", "duration_raw", dur, "duration_string", dur.String())
-	output = buf.String()
-	// Print logs for humans to see, if needed.
-	fmt.Println(output)
-
-	require.NotContainsf(t, output, "\"duration_raw\":90000000000", "Expected duration to be output as Go duration string \"1m30s\", got \"90000000000\"")
-
-	buf.Reset()
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			buf.Reset()
+			_ = config.Format.Set(tc.logFormat)
+			logger := New(config)
+			logger.Info("duration testing", "duration_raw", dur, "duration_string", dur.String())
+			require.Contains(t, buf.String(), tc.want)
+		})
+	}
 }
 
 func TestDynamicLevels(t *testing.T) {
--- a/promslog/slog_test.go
+++ b/promslog/slog_test.go
@@ -96,18 +96,12 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
 }
 
 func TestDurationValues(t *testing.T) {
-	var (
-		buf    bytes.Buffer
-		dur, _ = time.ParseDuration("1m30s")
-		config = &Config{
-			Writer: &buf,
-			Format: NewFormat(),
-		}
-	)
+	dur, err := time.ParseDuration("1m30s")
+	require.NoError(t, err)
 
 	tests := map[string]struct {
-		want      string
 		logFormat string
+		want      string
 	}{
 		"logfmt_duration_testing": {want: "duration_raw=1m30s", logFormat: "logfmt"},
 		"json_duration_testing":   {want: "\"duration_raw\":\"1m30s\"", logFormat: "json"},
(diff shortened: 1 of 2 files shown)


promslog/slog_test.go
```diff
@@ -94,6 +95,41 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
...
+	)
+
+	// Test logfmt.
+	logger := New(config)
+	logger.Info("duration testing", slog.Duration("duration_raw", dur), slog.String("duration_string", dur.String()))
+	output = buf.String()
+	// Print logs for humans to see, if needed.
+	fmt.Println(output)
+
+	require.NotContainsf(t, output, "duration_raw=\"90000000000\"", "Expected duration to be output as Go duration string \"1m30s\", got \"90000000000\"")
```
machine424: can't we just `require.Equal()`? as the output should be known and stable.
also no need for the message, `require` should take care of that

promslog/slog_test.go
```diff
@@ -94,6 +95,41 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
...
+		dur, _ = time.ParseDuration("1m30s")
+		config = &Config{Writer: &buf}
+	)
+
+	// Test logfmt.
+	logger := New(config)
+	logger.Info("duration testing", slog.Duration("duration_raw", dur), slog.String("duration_string", dur.String()))
+	output = buf.String()
+	// Print logs for humans to see, if needed.
+	fmt.Println(output)
```
machine424: no need I think, `require` should print it when sth is wrong.

promslog/slog_test.go
```diff
@@ -94,6 +95,41 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
...
+	var (
+		buf    bytes.Buffer
+		output string
+		dur, _ = time.ParseDuration("1m30s")
+		config = &Config{Writer: &buf}
+	)
+
+	// Test logfmt.
+	logger := New(config)
+	logger.Info("duration testing", slog.Duration("duration_raw", dur), slog.String("duration_string", dur.String()))
```
machine424: couldn't we just `logger.Info("duration testing", "duration_raw", dur, "duration_string", dur.String())`? as that's how it'd be usually used.

promslog/slog_test.go
```diff
@@ -94,6 +95,41 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
...
+	output = buf.String()
+	// Print logs for humans to see, if needed.
+	fmt.Println(output)
+
+	require.NotContainsf(t, output, "duration_raw=\"90000000000\"", "Expected duration to be output as Go duration string \"1m30s\", got \"90000000000\"")
+
+	buf.Reset()
+
+	// Test json.
```
machine424: we can use test cases here where the config is customized, I think we can set format there as well...

promslog/slog_test.go
tjhop: Sure, I'm fine with that

promslog/slog_test.go
tjhop: True, we just do it for other tests currently, so it was a consistency thing -- certainly not required

promslog/slog_test.go
tjhop: We can indeed and that would be equivalent. Considering this is a test suite and we're testing behavior for specific types, it seemed appropriate to explicitly add these values in a type-deterministic way rather than letting slog's sugar logger calls do the type conversion stuff behind the scenes. 

Opinions?

promslog/slog_test.go
tjhop: That's fair, I can clean it up a bit :+1:

promslog/slog_test.go
machine424: I'd go with testing as much of the logic as possible under conditions that closely mimic real‑world usage.

promslog/slog_test.go
tjhop: Woops, forgot. The timestamp in the log entry makes a string equality match on the full log line trickier because you need to either strip the timestamp out or mock it (that's why most of the other tests also use contains/notcontains). I can definitely rework it to use `require.Contains()` instead so the logic is more straightforward for this test, though

machine424 (approved):
lgtm, thanks.

Some nits to isolate the subtests and test both fields:

```diff
diff --git a/promslog/slog_test.go b/promslog/slog_test.go
index 2efb82d..5ffe3cf 100644
--- a/promslog/slog_test.go
+++ b/promslog/slog_test.go
@@ -96,27 +96,25 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
 }
 
 func TestDurationValues(t *testing.T) {
-       var (
-               buf    bytes.Buffer
-               dur, _ = time.ParseDuration("1m30s")
-               config = &Config{
-                       Writer: &buf,
-                       Format: NewFormat(),
-               }
-       )
+       dur, err := time.ParseDuration("1m30s")
+       require.NoError(t, err)
 
        tests := map[string]struct {
-               want      string
                logFormat string
+               want      string
        }{
-               "logfmt_duration_testing": {want: "duration_raw=1m30s", logFormat: "logfmt"},
-               "json_duration_testing":   {want: "\"duration_raw\":\"1m30s\"", logFormat: "json"},
+               "logfmt_duration_testing": {logFormat: "logfmt", want: "duration_raw=1m30s duration_string=1m30s"},
+               "json_duration_testing":   {logFormat: "json", want: "\"duration_raw\":\"1m30s\",\"duration_string\":\"1m30s\""},
        }
 
        for name, tc := range tests {
                t.Run(name, func(t *testing.T) {
-                       buf.Reset()
-                       _ = config.Format.Set(tc.logFormat)
+                       var buf bytes.Buffer
+                       config := &Config{
+                               Writer: &buf,
+                               Format: NewFormat(),
+                       }
+                       require.NoError(t, config.Format.Set(tc.logFormat))
                        logger := New(config)
                        logger.Info("duration testing", "duration_raw", dur, "duration_string", dur.String())
                        require.Contains(t, buf.String(), tc.want)
```

promslog/slog_test.go
```diff
@@ -94,6 +95,33 @@ func getLogEntryLevelCounts(s string, re *regexp.Regexp) map[string]int {
...
 
+func TestDurationValues(t *testing.T) {
+	dur, err := time.ParseDuration("1m30s")
+	require.NoError(t, err)
+
+	tests := map[string]struct {
+		logFormat string
+		want      string
+	}{
+		"logfmt_duration_testing": {want: "duration_raw=1m30s", logFormat: "logfmt"},
```
machine424: nit: if we could also add `duration_string ...`

PR #799, opened 2025-07-01
model.ValidationScheme: Support encoding as YAML

aknuds1:
Modify `model.ValidationScheme` to support encoding as YAML. To facilitate this, I add a default value for the type: `UnsetValidation`. I also add a `String` method to the type, for use in error messages etc (exemplified in `IsValidMetricName`). 

The motivation is to consolidate Prometheus around `model.ValidationScheme`, dropping the duplicated representations `config.LegacyValidationConfig` and `config.UTF8ValidationConfig`. This will in turn facilitate moving Prometheus (and e.g. Grafana Mimir) off the deprecated `model.NameValidationScheme` global.

See https://github.com/prometheus/prometheus/pull/16806 for reference, where I implement the consolidation around `model.ValidationScheme`.

## Default value changing to `UnsetValidation`

From my testing, without changing the default value to `UnsetValidation`, you'll get `LegacyValidation` when YAML unmarshalling an empty string. This means you can't tell whether the user has set the validation scheme in a configuration file or not.

### DIFF
--- a/model/metric_test.go
+++ b/model/metric_test.go
@@ -107,7 +107,7 @@ func TestValidationScheme_String(t *testing.T) {
 		{
 			name:   "Unset",
 			scheme: UnsetValidation,
-			want:   `unset`,
+			want:   "unset",
 		},
 		{
 			name:   "Legacy",
@@ -165,7 +165,7 @@ func TestValidationScheme_UnmarshalYAML(t *testing.T) {
 	}{
 		{
 			name:  "Unset empty input",
-			input: ``,
+			input: "",
 			want:  UnsetValidation,
 		},
 		{
@@ -175,17 +175,17 @@ func TestValidationScheme_UnmarshalYAML(t *testing.T) {
 		},
 		{
 			name:  "Legacy",
-			input: `legacy`,
+			input: "legacy",
 			want:  LegacyValidation,
 		},
 		{
 			name:  "UTF8",
-			input: `utf8`,
+			input: "utf8",
 			want:  UTF8Validation,
 		},
 		{
 			name:      "Invalid",
-			input:     `invalid`,
+			input:     "invalid",
 			wantError: errors.New(`unrecognized ValidationScheme: "invalid"`),
 		},
 	} {


model/metric.go
```diff
@@ -62,16 +63,72 @@ var (
...
+	case UTF8Validation:
+		return "utf8"
+	default:
+		panic(fmt.Errorf("unhandled ValidationScheme: %d", s))
+	}
+}
+
+// MarshalYAML implements the yaml.Marshaler interface.
+func (s ValidationScheme) MarshalYAML() (any, error) {
+	switch s {
```
juliusmh, line 101: Legacy, and UTF8 cases can be simplified to:

```go
return s.String(), nil
```

aknuds1: Thanks, done.

model/metric.go
```diff
@@ -185,7 +240,7 @@ func IsValidMetricName(n LabelValue) bool {
 		}
 		return utf8.ValidString(string(n))
 	default:
-		panic(fmt.Sprintf("Invalid name validation scheme requested: %d", NameValidationScheme))
+		panic(fmt.Sprintf("Invalid name validation scheme requested: %s", NameValidationScheme.String()))
```
juliusmh, line 243: nit: since we implement `fmt.Stringer` anyways

```suggestion
		panic(fmt.Sprintf("Invalid name validation scheme requested: %v", NameValidationScheme))
```

aknuds1: I don't understand the change to `%v` though, `%s` is more specific and thus better IMO. There's no real benefit to dropping the `.String()` call, it just becomes implicit.

model/metric.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	dto "github.com/prometheus/client_model/go"
 	"google.golang.org/protobuf/proto"
+	"gopkg.in/yaml.v2"
```
jesusvazquez, line 27: v3 just came out in case you'd like to upgrade https://gopkg.in/yaml.v3

model/metric.go
```diff
@@ -62,16 +63,70 @@ var (
...
 const (
+	// UnsetValidation represents an undefined ValidationScheme.
+	// Should not be used in practice.
+	UnsetValidation ValidationScheme = iota
+
 	// LegacyValidation is a setting that requires that all metric and label names
 	// conform to the original Prometheus character requirements described by
 	// MetricNameRE and LabelNameRE.
-	LegacyValidation ValidationScheme = iota
+	LegacyValidation
```
jesusvazquez, line 73: Did you check if the change of ordering here is going to cause any issues?

model/metric.go
```diff
@@ -62,16 +63,70 @@ var (
...
+
+// MarshalYAML implements the yaml.Marshaler interface.
+func (s ValidationScheme) MarshalYAML() (any, error) {
+	switch s {
+	case UnsetValidation:
+		return "", nil
+	case LegacyValidation, UTF8Validation:
+		return s.String(), nil
+	default:
+		panic(fmt.Errorf("unhandled ValidationScheme: %d", s))
```
jesusvazquez, line 107: Why not return an error here?

aknuds1: One _could_ return an error here, but I reasoned that a panic makes sense as it's a code path that should never happen.

model/metric.go
aknuds1: It's not really a change of ordering is it, but a change of default, right? It's a breaking change if anyone depends on `LegacyValidation` being the default, but that's a bad idea since you can't tell whether the enum is set or not. I'm not sure how to figure out if this will break anyone, but I can try to search for cases in Prometheus and Mimir.

jesusvazquez: > but that's a bad idea since you can't tell whether the enum is set or not. 

I see that now,. Well if you cant find a dependency in mimir or prom to the current default then I'm happy with this change.

aknuds1: Thanks @jesusvazquez - I will try to look a bit.

model/metric.go
aknuds1: Prometheus still uses v2 at least. It would be a larger lift and shift to upgrade.

jesusvazquez: that sounds fair then

model/metric.go
aknuds1: I don't see anything breaking in Prometheus/Mimir from this change at least.

I also did some more testing, and found that YAML decoding depends on `UnsetValidation` being the default enum value. The reason being that when unmarshalling from an empty string, `ValidationScheme.UnmarshalYAML` isn't called, so you always get the default value. If `LegacyValidation` remains the default, you can't tell whether the user has configured the validation scheme or not in YAML.

model/metric.go
bwplotka: Just? :p it's long time ago and it's already deprecated and archived.

See https://github.com/prometheus/proposals/issues/49

aknuds1: I left a [comment](https://github.com/prometheus/proposals/issues/49#issuecomment-3031250580) on your proposal @bwplotka about possibly transitioning to https://github.com/yaml/go-yaml.

PR #804, opened 2025-07-14
Add build tag to build without global name validation scheme

aknuds1:
Add build tag `localvalidationscheme`, to support building without global metric/label name validation scheme (`model.NameValidationScheme`). This is to support building projects such as Prometheus and Grafana Mimir with configurable metric/label name validation scheme, without changing the API unless said build tag is used.


ywwg:
How long do you think we'll need the two build tags?  It can be brittle to have two code paths like this, even if the divergence is small.  I know other projects still need time to update their code for the new APIs but we'll want to try to set a deadline.

Usually a build tag is exposed somewhere in the Makefile, and that's a good place to mark it experimental.  We'll also need README and other doc updates to explain what's going on.

aknuds1:
I've discussed with @ywwg in [Slack](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1752495869863399).

README.md
```diff
@@ -15,3 +15,16 @@ any stability guarantees for external usage.
...
+The libraries in this Go module share a notion of metric and label name validation scheme.
+There are two different schemes to choose from:
+* `model.LegacyValidation` => Metric and label names have to conform to the original Prometheus character requirements
+* `model.UTF8Validation` => Metric and label names are only required to be valid UTF-8 strings
+
+The active name validation scheme is normally implicitly controlled via the global variable `model.NameValidationScheme`.
+It's used by functions such as `model.IsValidMetricName` and `model.LabelName.IsValid`.
+_However_, if building with the _experimental_ build tag `localvalidationscheme`, the `model.NameValidationScheme` global is removed, and the API changes to accept the name validation scheme as an explicit parameter.
+`model.NameValidationScheme` is deprecated, and at some point, the API currently controlled by the build tag `localvalidationscheme` becomes standard.
+For the time being, the `localvalidationscheme` build tag is experimental and the API enabled by it may change.
```
ywwg:
```suggestion
`model.NameValidationScheme` is deprecated, and at some point, the API currently controlled by the build tag `localvalidationscheme` becomes standard.
All users of this library will be expected to move to the new API, and there will be an announcement in the changelog when the global is removed.
For the time being, the `localvalidationscheme` build tag is experimental and the API enabled by it may change.
```

ywwg:
Would it be possible to use a new function name for the API rather than rely on a build tag?  Since the function signature has changed, all of the call sites in mimir will need to change already anyway, right?

aknuds1:
Closing, since we've taken a different approach.

PR #806, opened 2025-07-28
Add `ValidationScheme` methods `IsValidMetricName` and `IsValidLabelName`

aknuds1:
Add `ValidationScheme` method `IsValidMetricName` to replace `model.IsValidMetricName` and `IsValidLegacyMetricName` , and `ValidationScheme` method `IsValidMetricName` to replace `model.LabelName.IsValid` and `model.LabelName.IsValidLegacy`. The replaces functions are marked as deprecated, so to phase out dependency on the `model.NameValidationScheme` global.

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -144,9 +144,9 @@ func (s ValidationScheme) IsValidMetricName(metricName string) bool {
 		if len(metricName) == 0 {
 			return false
 		}
-		return utf8.ValidString(string(metricName))
+		return utf8.ValidString(metricName)
 	default:
-		panic(fmt.Sprintf("Invalid metricName validation scheme requested: %s", s.String()))
+		panic(fmt.Sprintf("Invalid name validation scheme requested: %s", s.String()))
 	}
 }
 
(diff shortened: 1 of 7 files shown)


model/metric.go
```diff
@@ -127,6 +127,53 @@ func (s *ValidationScheme) UnmarshalYAML(unmarshal func(any) error) error {
...
+// IsValidLabelName returns whether labelName is valid according to s.
+func (s ValidationScheme) IsValidLabelName(labelName string) bool {
+	if len(labelName) == 0 {
+		return false
+	}
+	switch s {
+	case LegacyValidation:
+		if len(labelName) == 0 {
+			return false
+		}
```
ywwg: I think the copilot note was correct? I also see a duplicate zero length check

ywwg (commented):
thanks for this! just one nit that copilot found -- will approve when you mark this as ready for review

model/metric.go
aknuds1: Hm... let me check.

aknuds1: Yes, it was correct in fact. I ignored it since the suggestion looked out of wack. I pushed a fix.

PR #807, opened 2025-07-31
 ValidationScheme implements pflag.Value and json.Marshaler/Unmarshaler interfaces

juliusmh:


model/metric.go
```diff
@@ -77,10 +78,12 @@ const (
 	UTF8Validation
 )
 
-var (
-	_ yaml.Marshaler = UnsetValidation
-	_ fmt.Stringer   = UnsetValidation
```
aknuds1, line 82: Why are you removing the `fmt.Stringer` assertion?

juliusmh: Ahh good catch. I ment to check for pflag.Value because it already contains String(). But I don't want to add the dependency just for this check.

aknuds1: It's probably a better idea to keep the `fmt.Stringer` assertion, for explicitness if nothing else.

model/metric_test.go
```diff
@@ -203,6 +204,94 @@ func TestValidationScheme_UnmarshalYAML(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var got ValidationScheme
+			err := json.Unmarshal([]byte(tc.input), &got)
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, tc.want, ValidationScheme(got))
```
aknuds1: Superfluous cast.
```suggestion
			require.Equal(t, tc.want, got)
```

model/metric_test.go
```diff
@@ -203,6 +204,94 @@ func TestValidationScheme_UnmarshalYAML(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var got ValidationScheme
+			err := got.Set(tc.input)
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, tc.want, ValidationScheme(got))
```
aknuds1: Superfluous cast.
```suggestion
			require.Equal(t, tc.want, got)
```

aknuds1 (approved):

PR #808, opened 2025-07-31
Decoder: Remove use of global name validation and add validation

ywwg:
This change removes the use of the global variable to determine name validation, using the provided format instead.  

Also:

* adds validity checks to the text parser, which previously did not check for name validity.
* Returns error if the caller tries to create a decoder for prototext format, which is not supported.  

Signed-off-by: Owen Williams <owen.williams@grafana.com>

### DIFF
--- a/expfmt/decode.go
+++ b/expfmt/decode.go
@@ -70,19 +70,34 @@ func ResponseFormat(h http.Header) Format {
 	return FmtUnknown
 }
 
-// NewDecoder returns a new decoder based on the given input format.
-// If the input format does not imply otherwise, a text format decoder is returned.
+// NewDecoder returns a new decoder based on the given input format. Metric
+// names are validated based on the provided Format -- if the format requires
+// escaping, raditional Prometheues validity checking is used. Otherwise, names
+// are checked for UTF-8 validity. Supported formats include delimited protobuf
+// and Prometheus text format.  For historical reasons, this decoder fallbacks
+// to classic text decoding for any other format. This decoder does not fully
+// support OpenMetrics although it may often succeed due to the similarities
+// between the formats. This decoder may not support the latest features of
+// Prometheus text format and is not intended for high-performance applications.
+// See: https://github.com/prometheus/common/issues/812
 func NewDecoder(r io.Reader, format Format) Decoder {
+	scheme := model.LegacyValidation
+	if format.ToEscapingScheme() == model.NoEscaping {
+		scheme = model.UTF8Validation
+	}
 	switch format.FormatType() {
 	case TypeProtoDelim:
-		return &protoDecoder{r: bufio.NewReader(r)}
+		return &protoDecoder{r: bufio.NewReader(r), s: scheme}
+	case TypeProtoText, TypeProtoCompact:
+		return &errDecoder{err: fmt.Errorf("format %s not supported for decoding", format)}
 	}
-	return &textDecoder{r: r}
+	return &textDecoder{r: r, s: scheme}
 }
 
 // protoDecoder implements the Decoder interface for protocol buffers.
 type protoDecoder struct {
 	r protodelim.Reader
+	s model.ValidationScheme
 }
 
 // Decode implements the Decoder interface.
@@ -93,8 +108,7 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 	if err := opts.UnmarshalFrom(d.r, v); err != nil {
 		return err
 	}
-	//nolint:staticcheck // model.IsValidMetricName is deprecated.
-	if !model.IsValidMetricName(model.LabelValue(v.GetName())) {
+	if !d.s.IsValidMetricName(v.GetName()) {
 		return fmt.Errorf("invalid metric name %q", v.GetName())
 	}
 	for _, m := range v.GetMetric() {
@@ -108,8 +122,7 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 			if !model.LabelValue(l.GetValue()).IsValid() {
 				return fmt.Errorf("invalid label value %q", l.GetValue())
 			}
-			//nolint:staticcheck // model.LabelName.IsValid is deprecated.
-			if !model.LabelName(l.GetName()).IsValid() {
+			if !d.s.IsValidLabelName(l.GetName()) {
 				return fmt.Errorf("invalid label name %q", l.GetName())
 			}
 		}
@@ -117,10 +130,20 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 	return nil
 }
 
+// errDecoder is an error-state decoder that always returns the same error.
+type errDecoder struct {
+	err error
+}
+
+func (d *errDecoder) Decode(v *dto.MetricFamily) error {
+	return d.err
+}
+
 // textDecoder implements the Decoder interface for the text protocol.
 type textDecoder struct {
 	r    io.Reader
 	fams map[string]*dto.MetricFamily
+	s    model.ValidationScheme
 	err  error
 }
 
@@ -128,7 +151,7 @@ type textDecoder struct {
 func (d *textDecoder) Decode(v *dto.MetricFamily) error {
 	if d.err == nil {
 		// Read all metrics in one shot.
-		var p TextParser
+		p := TextParser{scheme: d.s}
 		d.fams, d.err = p.TextToMetricFamilies(d.r)
 		// If we don't get an error, store io.EOF for the end.
 		if d.err == nil {
--- a/expfmt/text_parse.go
+++ b/expfmt/text_parse.go
@@ -78,6 +78,9 @@ type TextParser struct {
 	// These indicate if the metric name from the current line being parsed is inside
 	// braces and if that metric name was found respectively.
 	currentMetricIsInsideBraces, currentMetricInsideBracesIsPresent bool
+	// scheme sets the desired ValidationScheme for names. Defaults to the invalid
+	// UnsetValidation.
+	scheme model.ValidationScheme
 }
 
 // TextToMetricFamilies reads 'in' as the simple and flat text-based exchange
@@ -126,6 +129,7 @@ func (p *TextParser) TextToMetricFamilies(in io.Reader) (map[string]*dto.MetricF
 
 func (p *TextParser) reset(in io.Reader) {
 	p.metricFamiliesByName = map[string]*dto.MetricFamily{}
+	p.currentLabelPairs = nil
 	if p.buf == nil {
 		p.buf = bufio.NewReader(in)
 	} else {
@@ -216,6 +220,9 @@ func (p *TextParser) startComment() stateFn {
 		return nil
 	}
 	p.setOrCreateCurrentMF()
+	if p.err != nil {
+		return nil
+	}
 	if p.skipBlankTab(); p.err != nil {
 		return nil // Unexpected end of input.
 	}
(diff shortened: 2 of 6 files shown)


expfmt/decode.go
```diff
@@ -72,12 +74,17 @@ func ResponseFormat(h http.Header) Format {
 
 // NewDecoder returns a new decoder based on the given input format.
 // If the input format does not imply otherwise, a text format decoder is returned.
-func NewDecoder(r io.Reader, format Format) Decoder {
+func NewDecoder(r io.Reader, format Format) (Decoder, error) {
 	switch format.FormatType() {
 	case TypeProtoDelim:
-		return &protoDecoder{r: bufio.NewReader(r)}
+		return &protoDecoder{r: bufio.NewReader(r)}, nil
+	case TypeProtoText, TypeProtoCompact:
```
bwplotka, line 91: I would suggest not add this until it becomes an official scrape protocol format, unless someone asked for it and there's a strong justification. Users should not depend on prototext and ideally we can avoid extra code to maintain. Did anyone ask for this?

And if we really want this, let's do this in separate PR to ensure the relevant fixes are atomically committed (for easier reverts, investigations etc).

expfmt/decode.go
```diff
@@ -72,12 +74,17 @@ func ResponseFormat(h http.Header) Format {
 
 // NewDecoder returns a new decoder based on the given input format.
 // If the input format does not imply otherwise, a text format decoder is returned.
-func NewDecoder(r io.Reader, format Format) Decoder {
+func NewDecoder(r io.Reader, format Format) (Decoder, error) {
```
bwplotka: This is scary, let's be careful here, ideally, in common, we don't break API and if we do we do this is, we do this in a dedicated PR and with tests 😱 

It's a signature and (hidden) logic breaking change on a popular API that will affect thousands indirectly of downstream projects and libraries. Signature because the new error handling (1), and hidden logic (2) breaking change, because for OpenMetrics format we suddenly return error and not attempt text like we did in the past. For majority of the existing users. the previous implementation, while broken, was "predictably" broken and we don't hear input to change it now, do we?

For (2), are we sure that no OpenMetrics attempt of parsing with text parser worked? I'm not, I believe for some basic metrics it could work just fine and users depended on it. Instead we should probably, long term should add automatic OM decoding here (no urgency though - users should use textparse from Prometheus ideally).

For (1), we don't need error handling if we skip (2) change. Another idea would be to create `errDecoder` and error on Decode which is what OpenMetrics input would cause to text parse in *some* cases. Still this would cause (2) breaking change, so I would just keep `NewDecoder` as is and only try to add safer parsers ONLY if there's strong use case.

If we really need safe decoder factory, we could add new `NewSafeDecoder(io.Reader, Format) (Decoder, error)`, but then ensure default (unknown Format) returns error too.

expfmt/decode.go
```diff
@@ -72,12 +74,17 @@ func ResponseFormat(h http.Header) Format {
...
 	switch format.FormatType() {
 	case TypeProtoDelim:
-		return &protoDecoder{r: bufio.NewReader(r)}
+		return &protoDecoder{r: bufio.NewReader(r)}, nil
+	case TypeProtoText, TypeProtoCompact:
+		return &prototextDecoder{r: r}, nil
 	}
-	return &textDecoder{r: r}
+	if format.FormatType() == TypeOpenMetrics {
+		return nil, errors.New("cannot decode openmetrics")
```
bwplotka: just to reiterate - this is a hidden breaking change I think

expfmt/decode.go
```diff
@@ -115,6 +122,44 @@ func (d *protoDecoder) Decode(v *dto.MetricFamily) error {
 	return nil
 }
 
+// prototextDecoder implements the Decoder interface for protocol buffers that
+// are encoded in text format.
+type prototextDecoder struct {
```
bwplotka: again, let's add in separate PR if we have and ideally, don't add if we don't need to 🙈

expfmt/encode.go
```diff
@@ -153,7 +153,7 @@ func NewEncoder(w io.Writer, format Format, options ...EncoderOption) Encoder {
 	case TypeProtoDelim:
 		return encoderCloser{
 			encode: func(v *dto.MetricFamily) error {
-				_, err := protodelim.MarshalTo(w, v)
+				_, err := protodelim.MarshalTo(w, model.EscapeMetricFamily(v, escapingScheme))
```
bwplotka: That's the main important bug fix here and it deserves as separate PR please 🙏🏽 🙈

expfmt/text_parse.go
```diff
@@ -574,7 +574,14 @@ func (p *TextParser) readingType() stateFn {
 	if p.readTokenUntilNewline(false); p.err != nil {
 		return nil // Unexpected end of input.
 	}
-	metricType, ok := dto.MetricType_value[strings.ToUpper(p.currentToken.String())]
+	typString := strings.ToUpper(p.currentToken.String())
+	// OpenMetrics uses UNKNOWN for untyped metrics instead of UNTYPED, so we add
```
bwplotka: Hm, this PR was erroring when somebody tries to use OM with text parse, but we still try to handle OM here? 🤔 What am I missing? Also a candidate for a separate PR that enables OM decoding, if we want

bwplotka (changes requested):
Thanks for jumping into it!

Let's fix proto + escaping, thanks for tackling this, it's important.

However, given we had some issues/bugs recently and generally it's not trivial to develop with common because many stable libs and projects use this, let's perhaps:
* Make separate PRs for separate fixes so it's easier to manager and ensure we didn't miss anything.
* Be careful with breaking changes, this PR proposes 2 breaking changes I believe we could avoid.
* Don't try to avoid introducing logic that was not asked for (unless I don't have context) -- ideally we keep the maintenance surface minimal

Hope that makes sense 🙈  WDYT?

expfmt/decode.go
ywwg: the common library supports encoding in these formats, so shouldn't it also support decoding them?  I don't quite understand the logic behind the current one-way design of the library.

expfmt/text_parse.go
ywwg: This was a bit of a stupid hack to try to return a better error message when someone does try to do this.  I do agree it's in conflict with refusing to decode openmetrics

expfmt/decode.go
ywwg: @beorn7 looked at Prometheus, and it does not use the common library to do parsing.  I doubt anyone actually uses this decoder for openmetrics because it is broken for unknown/untyped metrics. I would have to look closer but I would bet it also doesn't support other openmetrics idioms like different quoting characters.

I think if we don't change the return signature, we should add a clear warning in the docstring that the decoder does not support openmetrics.

ywwg:
@bwplotka thanks for reviewing this! I agree this PR does too many things and for now we should focus on fixing the escaping bug.  But I also think we might want to sit down and decide what the common library should be expected to do: if it can generate output, and has a decoder, should it not support reading that output? I believe we will find more bugs because right now the output is barely tested.

ywwg:
Notes from slack discussion:

* merge separate escaping fix PR
* add textproto support
* add comments to Decode that it does not support OpenMetrics totally, also that the fastest/best experience is the prometheus parsers.
* do not change the newdecoder function signature

expfmt/decode.go
ywwg: My new suggestion is, how about we mark the prototext types as deprecated, so if people try to use them, they get a warning?

expfmt/text_parse.go
```diff
@@ -126,6 +129,7 @@ func (p *TextParser) TextToMetricFamilies(in io.Reader) (map[string]*dto.MetricF
 
 func (p *TextParser) reset(in io.Reader) {
 	p.metricFamiliesByName = map[string]*dto.MetricFamily{}
+	p.currentLabelPairs = nil
```
ywwg, line 132: the fact that this member was not reset does not bode well for the correctness of the whole parser

expfmt/text_parse.go
```diff
@@ -341,6 +358,12 @@ func (p *TextParser) startLabelName() stateFn {
 	p.currentLabelPair = &dto.LabelPair{Name: proto.String(p.currentToken.String())}
 	if p.currentLabelPair.GetName() == string(model.MetricNameLabel) {
 		p.parseError(fmt.Sprintf("label name %q is reserved", model.MetricNameLabel))
+		p.currentLabelPairs = nil
+		return nil
+	}
+	if !p.scheme.IsValidLabelName(p.currentLabelPair.GetName()) {
```
ywwg, line 364: these checks were not done at all previously

expfmt/decode.go
```diff
@@ -70,19 +71,34 @@ func ResponseFormat(h http.Header) Format {
 	return FmtUnknown
 }
 
-// NewDecoder returns a new decoder based on the given input format.
-// If the input format does not imply otherwise, a text format decoder is returned.
+// NewDecoder returns a new decoder based on the given input format. Supported
+// formats include delimited protobuf, text protos, and Prometheus text format.
+// This decoder does not fully support OpenMetrics although it may often succeed
```
bwplotka: Should we add issue about OM and link in some TODO here?

As per https://cloud-native.slack.com/archives/C01AUBA4PFE/p1755075286782209?thread_ts=1753991269.132859&cid=C01AUBA4PFE discussion

expfmt/decode.go
bwplotka: sounds like the last comment is obsolete, resolving discussion.

expfmt/decode.go
```diff
@@ -70,19 +71,34 @@ func ResponseFormat(h http.Header) Format {
...
+	scheme := model.LegacyValidation
+	if format.ToEscapingScheme() == model.NoEscaping {
+		scheme = model.UTF8Validation
+	}
 	switch format.FormatType() {
 	case TypeProtoDelim:
-		return &protoDecoder{r: bufio.NewReader(r)}
+		return &protoDecoder{r: bufio.NewReader(r), s: scheme}
+	case TypeProtoText, TypeProtoCompact:
+		return &prototextDecoder{r: r, s: scheme}
```
bwplotka: As @beorn7 mentioned, prototext was used for debugging purposes on encoding side. If that's the case should we return `errDecoder{err :err}` instead or fallback to text?

expfmt/decode.go
```diff
@@ -70,19 +71,34 @@ func ResponseFormat(h http.Header) Format {
 	return FmtUnknown
 }
 
-// NewDecoder returns a new decoder based on the given input format.
-// If the input format does not imply otherwise, a text format decoder is returned.
+// NewDecoder returns a new decoder based on the given input format. Supported
+// formats include delimited protobuf, text protos, and Prometheus text format.
```
bwplotka:
```suggestion
// formats include delimited protobuf, text protos, and Prometheus text format.
// For historical reasons, this decoder fallbacks to classic text decoding for any other format.
```

expfmt/decode.go
ywwg: done

expfmt/decode.go
ywwg: falling  back to text will just result in weird error messages. I would prefer to return a concrete error if we'd rather not support prototext decoding.  (Is that what you're saying?)

ywwg:
reworded / renamed again, because this PR keeps morphing.  I think it still does something useful?

expfmt/decode.go
```diff
@@ -70,19 +70,32 @@ func ResponseFormat(h http.Header) Format {
...
```
bwplotka:
```suggestion
// formats include delimited protobuf and Prometheus text format.
```

expfmt/text_parse.go
bwplotka: Yes. It also proves it's not heavily used (:

bwplotka (approved):
Thanks!!! LGTM, one nit only.

PR #809, opened 2025-08-01
Fix delimited proto not escaped correctly

thampiotr:
I have encountered an issue that is blocking me from upgrading Prometheus to 3.x where despite receiving a content type header of `"application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited; escaping=underscores"` I get `.` in metric and label names. 

I've noticed there is possibly a missing call to `EscapeMetricFamily` for delimited proto. 

I also noticed the test `TestEscapedEncode` did not assert that actual escaping is done, so fixed that too.

### DIFF
--- a/expfmt/encode_test.go
+++ b/expfmt/encode_test.go
@@ -18,6 +18,8 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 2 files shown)


expfmt/encode_test.go
```diff
@@ -309,86 +310,78 @@ foo_metric 1.234
...
 
-	err := delimEncoder.Encode(metric)
-	if err != nil {
-		t.Errorf("unexpected error during encode: %s", err.Error())
+	verifyCorrectlyEscaped := func(t *testing.T, out string, format Format) {
+		require.NotContainsf(t, out, `"foo.metric"`, "format incorrectly escaped: %s", format)
+		require.Containsf(t, out, `foo_metric`, "format incorrectly escaped: %s", format)
+		require.NotContainsf(t, out, `"dotted.label.name"`, "format incorrectly escaped: %s", format)
+		require.Containsf(t, out, `dotted_label_name`, "format incorrectly escaped: %s", format)
+		require.Containsf(t, out, `my.label.value`, "format incorrectly escaped: %s", format)
```
ywwg: I would prefer that we test the actual output (parsed proto, etc) instead of grepping the strings, but this is ok for now.  Can you add a todo to make the test more robust?  We will also want to test the other escaping methods, round tripping, etc.

thampiotr: I can convert this into a table test and verify the entire output if that works :)

thampiotr: Updated the tests, PTAL.

bwplotka (approved):
Thanks!

ywwg:
I was getting this weird tab/space problem when I was working on my own PR and could not figure out where it was coming from. I ended up just changing the test to not get tripped up on it.  I think there is some bizarre nondeterminism in the prototext output

ywwg (approved):
looks great, thanks!

thampiotr:
> I was getting this weird tab/space problem when I was working on my own PR and could not figure out where it was coming from. I ended up just changing the test to not get tripped up on it. I think there is some bizarre nondeterminism in the prototext output

The protobuf text representation is intentionally nondeterministic as per https://github.com/protocolbuffers/protobuf-go/commit/582ab3de426ef0758666e018b422dd20390f7f26, there is some context and workarounds here: https://github.com/golang/protobuf/issues/1082

In case of this one test, we're really interested in escaping, so I can achieve this by looking at strings in the output, which I reverted to now. 

More broadly the `common` module may want to have a strategy for verifying stability of the outputs. The issue linked above has some workarounds that could be used, but this is out of scope of this PR.

PR #811, opened 2025-08-05
expfmt: throw error when metric's name not provided instead panicing

Alhanaqtah:
This PR enhances extfmt behavior when incorrect metrics privided by user — metrics without name. Now ```TextParser``` throws error "invalid metrics name" instead of panicing ([#3374](https://github.com/prometheus/node_exporter/issues/3374))

@roidelapluie

### DIFF
--- a/expfmt/text_parse_test.go
+++ b/expfmt/text_parse_test.go
@@ -1340,7 +1340,7 @@ request_duration_microseconds_count 2693
 {host="local", dir="alpha"} 1834194837
 {host="remote", dir="beta"} 133638016
 `,
-			err: `text format parsing error in line 4: invalid metric name`,
+			errUTF8: `text format parsing error in line 4: invalid metric name`,
 		},
 	}
 	for i, scenario := range scenarios {


Alhanaqtah:
@roidelapluie could you review PR, please?)

ywwg:
This change looks good to me.  Can you add a test to exercise this case so we can catch regressions? And can you sign your commit to pass the DCO check? Thanks!

ywwg:
Hi, I would like to do a release of prom/common and this would be a nice fix to include, do you think you can add the test and fix DCO soon?

Alhanaqtah:
> Hi, I would like to do a release of prom/common and this would be a nice fix to include, do you think you can add the test and fix DCO soon?

Hi. I'm currently working on test. But I came across the fact that when the testcase is launched alone, it is successful, but if you run all the testcases, then my testcase crashes with an error. Could it be that the state in the parser is preserved between calls of ```TextToMetricFamilies```?

Alhanaqtah:
> Hi, I would like to do a release of prom/common and this would be a nice fix to include, do you think you can add the test and fix DCO soon?

I've done everything. You can merge it)

Alhanaqtah:
@ywwg Could you merge PR, please?)

expfmt/text_parse_test.go
```diff
@@ -934,6 +934,16 @@ line"="bla"} 3.14
 `,
 			err: `text format parsing error in line 2: label name "new" contains unescaped new-line`,
 		},
+		// 40: Metric's name missing.
+		{
+			in: `
+# HELP backupmonitor_size The size of the given backup.
+# TYPE backupmonitor_size counter
+{host="local", dir="alpha"} 1834194837
+{host="remote", dir="beta"} 133638016
```
jan--f, line 1341:
```suggestion
# HELP metric a metric.
# TYPE metric counter
{label="bla"} 3.14
```
Just to be consistent with previous test cases.

jan--f (changes requested):
So yes this is a corner case previously not tested for, thanks for fixing this.
Can you please rebase on current `main`?

expfmt/text_parse_test.go
Alhanaqtah: I came across the fact that the error is not displayed at the values as in the previous metrics, so I decided to use these values:
```
# HELP backupmonitor_size The size of the given backup.
# TYPE backupmonitor_size counter
{host="local", dir="alpha"} 1834194837
{host="remote", dir="beta"} 133638016
```

expfmt/text_parse.go
```diff
@@ -109,7 +109,7 @@ func NewTextParser(nameValidationScheme model.ValidationScheme) TextParser {
 // Summaries and histograms are rather special beasts. You would probably not
 // use them in the simple text format anyway. This method can deal with
 // summaries and histograms if they are presented in exactly the way the
-// text.Create function creates them.
+// text. Create function creates them.
```
jan--f, line 112: This looks like an accident?

expfmt/text_parse_test.go
jan--f: Not sure I follow. The request is make the test case consistent with what we use in the other test cases. The naming itself should have no effect on the case being tested.

PR #813, opened 2025-08-13
docs: fix typo in expfmt.Negotiate

wmcram:
The `Negotiate` function in the `expfmt` package was referencing a function `NegotiateOpenMetrics` which probably should be `NegotiateIncludingOpenMetrics`. I also noticed that the test referencing this function had the same issue; this commit makes the naming consistent everywhere.

Signed-off-by: Wil Cram [wcram@wisc.edu](mailto:wcram@wisc.edu)


wmcram:
Since the project is Apache licensed and this pr adds no new functionality, it looks like we should be fine.

ArthurSens (approved):

ArthurSens:
> noop lgtm. I don't know if we have a CLA that needs to be signed?

We have DCO checks, that should be enough :)

PR #817, opened 2025-08-25
Sync .golangci.yml with prometheus/prometheus

aknuds1:
Sync .golangci.yml with the much more extensive one in prometheus/prometheus. The motivation is that I noticed that superfluous type casting wasn't caught by the linter. I'm temporarily disabling all rules that are currently triggered by discrepancies in the code base, with the exception of the `gci` formatter. I fix `gci` complaints in the offending *.go files.


bwplotka (approved):
Thanks!

PR #822, opened 2025-08-26
Enable the fatcontext linter

aknuds1:
Enable the fatcontext linter and fix issues.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -78,7 +78,7 @@ func (r *Router) handle(handlerName string, h http.HandlerFunc) httprouter.Handl
 		defer cancel()
 
 		for _, p := range params {
-			ctx = context.WithValue(ctx, param(p.Key), p.Value)
+			ctx = context.WithValue(ctx, param(p.Key), p.Value) //nolint:fatcontext
 		}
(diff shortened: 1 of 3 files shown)


route/route.go
```diff
@@ -78,7 +78,7 @@ func (r *Router) handle(handlerName string, h http.HandlerFunc) httprouter.Handl
 		defer cancel()
 
 		for _, p := range params {
-			ctx = context.WithValue(ctx, param(p.Key), p.Value)
+			ctx = context.WithValue(ctx, param(p.Key), p.Value) //nolint:fatcontext
```
ywwg, line 81: if we turn this on, only to ignore it, is it worth using?

aknuds1: Fair point. It's useless in these particular instances, I couldn't find any better way of doing it. It probably has real use cases since it's enabled on Prometheus?

ywwg: being in sync with prometheus seems like a good enough reason

aknuds1: Ack, someone might otherwise try to reintroduce the same linter in the future when syncing with Prometheus.

PR #825, opened 2025-08-27
Use go.uber.org/atomic instead of sync/atomic

aknuds1:
Use `go.uber.org/atomic` instead of `sync/atomic`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -66,8 +66,8 @@ linters:
       rules:
         main:
           deny:
-            #- pkg: "sync/atomic"
-              #desc: "Use go.uber.org/atomic instead of sync/atomic"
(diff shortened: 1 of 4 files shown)


SuperQ:
Can we turn off these useless copilot "reviews"?

SuperQ (approved):

aknuds1:
> Can we turn off these useless copilot "reviews"?

@SuperQ They're not enabled by default - I tend to request them myself, just to catch in case I make some stupid mistake. I can stop requesting them if you don't like them :)

SuperQ:
I'd like to disable it completely. It's not reliable. It's more I now need to look at when reviewing a PR.

"AI" is not ready for this kind of work.

PR #832, opened 2025-09-03
Retract v1.20.3

### LINKED ISSUE #831: Strange tag behaviour on v1.20.3
v1.20.3 is not a tag of the common module. It appears to be a tag of the client module that is in the common git repo?

```
go: downloading github.com/prometheus/common v1.20.3
go: github.com/prometheus/common@upgrade (v1.20.3) requires github.com/prometheus/common@v1.20.3: parsing go.mod:
       module declares its path as: github.com/prometheus/client_golang
               but was required as: github.com/prometheus/common
```

I think the linting tools I use wants to upgrade github.com/prometheus/common to v1.20.3, this isn't strictly a problem with prometheus.

The linting tool I'm using forces us to use the latest released version of code. But its believing that the latest version of prometheus common is v1.20.3 because it was cut yesterday, and a bigger semver version than 0.66.0.

roidelapluie:
Fixes #831

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,4 +37,42 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-retract v0.50.0 // Critical bug in counter suffixes, please read issue https://github.com/prometheus/common/issues/605
+retract (
+	v1.20.99 // This tag is needed to retract accidental tags below, but is retracted directly.
+	v1.20.3 // Tags pushed accidentally, see https://github.com/prometheus/common/issues/831
+	v1.20.3
+	v1.20.2
+	v1.20.1
+	v1.20.0
+	v1.19.1
+	v1.19.0
+	v1.18.0
+	v1.17.0
+	v1.16.0
+	v1.15.1
+	v1.15.0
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -37,4 +37,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-retract v0.50.0 // Critical bug in counter suffixes, please read issue https://github.com/prometheus/common/issues/605
+retract (
+	v1.20.3 // Tag pushed accidentally, see https://github.com/prometheus/common/issues/831
+	v0.50.0 // Critical bug in counter suffixes, please read issue https://github.com/prometheus/common/issues/605
+)
```
ArthurSens: Looks like tons of other tags were pushed accidentally. Once we retract 1.20.3, will 1.20.2 become the latest one?

roidelapluie: Addressed

SuperQ:
I guess if we ever want to make v1, we'll need to start with v1.21.0. :sob:

SuperQ (approved):

PR #833, opened 2025-09-03
Create CHANGELOG.md for easier communication of library changes, especially possible breaking changes.

ywwg:


RELEASE.md
```diff
@@ -11,7 +11,16 @@ With those steps done, you can proceed to cut a release.
 
 ## How to cut an individual release
 
-There is no automated process for cutting a release in `prometheus/common`. A manual release using GitHub's release feature via [this link](https://github.com/prometheus/prometheus/releases/new) is the best way to go. The tag name must be prefixed with a `v` e.g. `v0.53.0` and then you can use the "Generate release notes" button to generate the release note automagically ✨. No need to create a discussion or mark it a pre-release, please do mark it as the latest release if needed.
+There is no automated process for cutting a release in `prometheus/common`.
+The primary trigger for announcing a release is pushing a new version tag.
```
ArthurSens, line 15: Actually, I think we don't want to do that. We shouldn't be pushing tags manually; GitHub UI is good enough for that, and it ensures we don't push tags unintentionally 🤔

RELEASE.md
```diff
@@ -11,7 +11,16 @@ With those steps done, you can proceed to cut a release.
...
+
+NOTE: As soon as the tag is pushed, many downstream projects will automatically create pull requests to update their dependency of prom/common.
+Make sure the release is ready to go, with an updated changelog including notices of any breaking changes, before pushing a new tag.
```
ArthurSens: Interesting, I think this is the root cause of our problems yesterday. Creating a release from the UI, without a pre-existing tag, will create this tag for us at the same time as the release is done.

I believe we shouldn't change this

RELEASE.md
ywwg: I'll clarify -- I just mean that the creation of the tag triggers releases, whether it's pushed or created by the UI

RELEASE.md
ywwg: the root cause was I used a command that pushes all tags, and I assumed that my tags would be in sync with upstream.  I did not consider that I would have local tags that had been deleted on the remote from an earlier cleanup

ywwg:
We won't know for sure until we try doing the next release, but I bet it'll work.  I copied the format of the prometheus/prometheus changelog, which seems to work

SuperQ (commented):
If we're going to go with `CHANGELOG` release workflow, we should also include a `VERSION` file.

VERSION
```diff
@@ -0,0 +1 @@
+0.66.0
```
aknuds1:
```suggestion
0.66.1
```

aknuds1 (changes requested):
VERSION is out of date.

PR #835, opened 2025-09-04
Revert "Use github.com/grafana/regexp instead of regexp"

### LINKED ISSUE #836: Revert dependency on `github.com/grafana/regexp`
We should revert dependency on the more performant `github.com/grafana/regexp` instead of the standard `regexp` package. It was pointed out by @dims that we end up forcing this dependency on projects using this library, and @bboreham reinforced that the performance improvements are not important enough to force a dependency on this library.

### From Slack:

@dims:
> [dependency] Need for grafana/regexp? :thread:
> [...]
> https://github.com/prometheus/common/commit/149efd6d704a94178a9123536aabadcf01a6b09a was recently added that introduced the new package. given that repo was for something missing in go1.22, do we need to switch over?
> [...]
> Can we confirm why we need it?

@aknuds1:
> I switched prometheus/common to grafana/regexp only to harmonize with prometheus/prometheus. It's encoded in prometheus/prometheus/.golangci.yml to use grafana/regexp, so I ported that rule over to prometheus/common. I just know that grafana/regexp has performance optimizations.
> Do you mean that the optimizations in grafana/regexp are now upstream?

@dims:

> yep, Bryan can you please confirm?

@bboreham:

> not all of them
> But we should only put that dependency in programs like Prometheus, not in libraries.
> Other programs using prometheus/common can make their own choices.
> [I made a slight mental error above.
> For total correctness: I was thinking projects downstream can override in go.mod, but this is not true for stdlib packages.]
> However I still think it's best to leave this dependency out of prometheus/common, unless it is shown to make a massive performance difference.

aknuds1:
Revert my change to use the more performant grafana/regexp, as this library shouldn't control this choice. Fixes #836.
This reverts commit 149efd6d704a94178a9123536aabadcf01a6b09a.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -72,8 +72,6 @@ linters:
               desc: "Use github.com/go-kit/log instead of github.com/go-kit/kit/log"
             - pkg: "io/ioutil"
               desc: "Use corresponding 'os' or 'io' functions instead."
-            - pkg: "regexp"
-              desc: "Use github.com/grafana/regexp instead of regexp"
             - pkg: "github.com/pkg/errors"
               desc: "Use 'errors' or 'fmt' instead of github.com/pkg/errors"
             - pkg: "gzip"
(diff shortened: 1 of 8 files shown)


aknuds1:
@SuperQ OK will do.

SuperQ (approved):

PR #838, opened 2025-09-04
Revert "Use go.uber.org/atomic instead of sync/atomic (#825)"

### LINKED ISSUE #837: Revert dependency on go.uber.org/atomic
We should revert dependency on `go.uber.org/atomic` instead of the standard `sync/atomic` package. It's a follow-up to #836, and based on the same logic we should avoid forcing the dependency on `go.uber.org/atomic` also. See #836 for context.

### From Slack:

@aknuds1:

> Given the same logic, should I revert also my [PR](https://github.com/prometheus/common/pull/825) switching us over to go.uber.org/atomic dims?

@dims:

> yes please!

aknuds1:
Revert my change to use the more performant `go.uber.org/atomic` instead of `sync/atomic`, as this library shouldn't control this choice. Fixes #837.

This reverts commit 6d40fe16ed5f8f21fae8f9e53fd8b54eb08c2d44.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -66,8 +66,6 @@ linters:
       rules:
         main:
           deny:
-            - pkg: "sync/atomic"
-              desc: "Use go.uber.org/atomic instead of sync/atomic"
             - pkg: "github.com/go-kit/kit/log"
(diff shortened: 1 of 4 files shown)


SuperQ (approved):
Thanks!

PR #839, opened 2025-09-06
Move interface assertions to a test file

msiegen:
This is the only dependency on YAML in the model package. Moving it to the tests avoids pulling a dependency into the main module of applications that don't otherwise use YAML.


roidelapluie:
Thanks!

PR #843, opened 2025-09-30
expfmt: Add support for float histograms and gauge histograms

### LINKED ISSUE #430: expfmt: need to support gauge histogram metric type and float histogram
The various `MetricFamilyTo...` functions are currently ignorant of the gauge histogram type recently added to the proto spec. Similarly, they ignore the float fields that got added to support a histogram where the counts are all floats.

Note that this is independent of the new native histograms. (The latter also support float histograms and gauge histograms, but so do the conventional histograms.)

beorn7:
Fixes #430.


expfmt/openmetrics_create.go
```diff
@@ -325,14 +327,20 @@ func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...E
...
 			if metric.Histogram == nil {
 				return written, fmt.Errorf(
 					"expected histogram in metric %s %s", compliantName, metric,
 				)
 			}
 			infSeen := false
 			for _, b := range metric.Histogram.Bucket {
+				if b.GetCumulativeCountFloat() > 0 {
+					return written, fmt.Errorf(
+						"float histogram %s %s not supported in OpenMetrics",
```
krajorama: Let's be a bit more generic to be consistent with other errors and avoid having to say OpenMetrics version as well.

```suggestion
						"unexpected float histogram bucket count in metric %s %s",
```

expfmt/openmetrics_create.go
```diff
@@ -368,6 +379,12 @@ func MetricFamilyToOpenMetrics(out io.Writer, in *dto.MetricFamily, options ...E
 			if err != nil {
 				return
 			}
+			if metric.Histogram.GetSampleCountFloat() > 0 {
+				return written, fmt.Errorf(
+					"float histogram %s %s not supported in OpenMetrics",
```
krajorama: Let's be a bit more generic to be consistent with other errors and avoid having to say OpenMetrics version as well.

```suggestion
					"unexpected float histogram count in metric %s %s",
```

expfmt/openmetrics_create.go
beorn7: Not sure if that nails it. The float histogram is not necessarily unexpected. It's fine in the data model, just that OM v1 cannot represent it.

We could say "current format does not support float histogram bucket count in metric %s %s" or maybe "OpenMetrics v1.0 does not support float histogram bucket count in metric %s %s". WDYT?

krajorama: In that case I'm ok with saying
```suggestion
						"float histogram %s %s not supported in OpenMetrics v1.0",
```

beorn7: OK, let's still put the `metric %s %s` at the end because it can be quite verbose (as I have seen during debugging).

beorn7:
How about this?

krajorama (approved):
LGTM

Issue #850, opened 2025-10-07
0.67.0 causes a file collision


Sanduhr32:
go get -u
go: downloading github.com/prometheus/common v0.67.0
go: create zip: version/info.go: case-insensitive file name collision: "VERSION" and "version"
version/info_test.go: case-insensitive file name collision: "VERSION" and "version"
go: create zip: version/info.go: case-insensitive file name collision: "VERSION" and "version"
version/info_test.go: case-insensitive file name collision: "VERSION" and "version"
go: create zip: version/info.go: case-insensitive file name collision: "VERSION" and "version"
version/info_test.go: case-insensitive file name collision: "VERSION" and "version"
go: create zip: version/info.go: case-insensitive file name collision: "VERSION" and "version"
version/info_test.go: case-insensitive file name collision: "VERSION" and "version"
go: create zip: version/info.go: case-insensitive file name collision: "VERSION" and "version"
version/info_test.go: case-insensitive file name collision: "VERSION" and "version"

PR #853, opened 2025-10-07
Fix Go case-insensitive file name collision

SuperQ:
Remove the `VERSION` file to avoid "case-insensitive file name collision" error importing this package.

Fixes: https://github.com/prometheus/common/issues/850


beorn7 (changes requested):
RELEASE.md needs an update because it refers to VERSION.

beorn7 (approved):

PR #862, opened 2025-10-30
Support JWT Profile for Authorization Grant (RFC 7523 3.1)

JorTurFer:
This PR implements JWT Authentication for Authorization Grant ([RFC 7523 3.1](https://datatracker.ietf.org/doc/html/rfc7523#section-3.1)) as alternative to static client credentials.

As usage example, this auth flow is used by STACKIT Service Accounts (Key Flow).

Note: This PR does not provide support for JWTs in [Client Authentication Processing](https://datatracker.ietf.org/doc/html/rfc7523#section-3.2), but I can include it if you think that it's worth or I can add it as another PR following this

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -254,10 +254,10 @@ type OAuth2 struct {
 	// "client_credentials" or "urn:ietf:params:oauth:grant-type:jwt-bearer" (RFC 7523).
 	// Default value is "client_credentials"
 	GrantType string `yaml:"grant_type" json:"grant_type"`
-	// SignatureAltgorithm is the RSA algorithm used to sign JWT token. Only used if
+	// SignatureAlgorithm is the RSA algorithm used to sign JWT token. Only used if
 	// GrantType is set to "urn:ietf:params:oauth:grant-type:jwt-bearer".
 	// Default value is RS256 and valid values RS256, RS384, RS512
-	SignatureAltgorithm string `yaml:"signature_algorithm,omitempty" json:"signature_algorithm,omitempty"`
+	SignatureAlgorithm string `yaml:"signature_algorithm,omitempty" json:"signature_algorithm,omitempty"`
 	// Iss is the OAuth client identifier used when communicating with
 	// the configured OAuth provider. Default value is client_id. Only used if
 	// GrantType is set to "urn:ietf:params:oauth:grant-type:jwt-bearer".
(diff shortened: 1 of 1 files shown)


JorTurFer:
Nice questions! I didn't think about the radius blast of adding new dependencies at core.

The problem that I see is that `golang.org/x/oauth2/jwt` 2 important features:
- Multiple signing algorithms
- `jti`

Although RS256 is the most common signing alg, there are auth providers which don't support it because they require RS512 as minimum (and eventually, there can be servers requesting other alg than RS but I guess that this is more overthinking because RS is accepted almost everywhere)

About `jti`, it's optional in the RFC but is also says 
> The authorization server MAY
> ensure that JWTs are not replayed by maintaining the set of used
> "jti" values for the length of time for which the JWT would be
> considered valid based on the applicable "exp" instant.

So it's also on authorization server the option to enforce the value if it's verified. Said this, `jti` could provided as custom claim because it's used to ensure that not access tokens are signed at the same time for the same assertion, so `golang.org/x/oauth2/jwt` implementation should be fine using static `jti`.

> Is it really necessary to define custom algorithms, or could we just rely on RS256 as golang.org/x/oauth2 does?

The biggest problem that I see is the signature methods hardcoded to RS256. RS384 and RS512 are the same method with different lengths and for enterprise authorization servers, increasing the length is mandatory.

About `uuid` dep, I used the same as [prometheus is already using](https://github.com/prometheus/prometheus/blob/84d2007a0830dddc77c63b62d32e6e430932d349/go.mod#L37) to avoid using multiple deps for same things

JorTurFer:
I didn't notice it but prometheus is also using [`github.com/golang-jwt/jwt/v5`](https://github.com/prometheus/prometheus/blob/84d2007a0830dddc77c63b62d32e6e430932d349/go.mod#L155) indirectly, so the dependency is already there (I know that this doesn't mean that it's already everywhere)

config/http_config.go
```diff
@@ -237,12 +243,38 @@ type OAuth2 struct {
...
+	// secret.
+	ClientCertificateKeyRef string `yaml:"client_certificate_key_ref" json:"client_certificate_key_ref"`
+	// GrantType is the OAuth2 grant type to use. It can be one of
+	// "client_credentials" or "urn:ietf:params:oauth:grant-type:jwt-bearer" (RFC 7523).
+	// Default value is "client_credentials"
+	GrantType string `yaml:"grant_type" json:"grant_type"`
+	// SignatureAltgorithm is the RSA algorithm used to sign JWT token. Only used if
+	// GrantType is set to "urn:ietf:params:oauth:grant-type:jwt-bearer".
+	// Default value is RS256 and valid values RS256, RS384, RS512
+	SignatureAltgorithm string `yaml:"signature_algorithm,omitempty" json:"signature_algorithm,omitempty"`
```
jkroepke: typo

SignatureAltgorithm -> SignatureAlgorithm

JorTurFer: nice catch!

JorTurFer:
> @JorTurFer Could you confirm that refreshing works on your side as well?

image

I can confirm it too, the refresh works as expected

jkroepke (approved):
LGTM

PR #877, opened 2026-01-27
Remove logic adding unit to metrics name

vesari:
Following [this discussion](https://github.com/prometheus/client_golang/pull/1392#discussion_r2682447946) regarding OM unit support in client_golang, this PR removes the optional unit flag and the related logic that would automatically change the metric name if the UNIT field was populated but the unit not already included in the metric name. This is needed to complete the afore mentioned client_golang [PR](https://github.com/prometheus/client_golang/pull/1392). cc @bwplotka .

### DIFF
--- a/expfmt/encode_test.go
+++ b/expfmt/encode_test.go
@@ -261,10 +261,10 @@ foo_metric 1.234
 foo_metric 1.234
 `,
 		},
-		// 7: Simple Counter FmtOpenMetrics_0_0_1
+		// 7: Simple Counter FmtOpenMetrics_1_0_0
 		{
 			metric: metric1,
-			format: FmtOpenMetrics_0_0_1,
+			format: FmtOpenMetrics_1_0_0,
 			expOut: `# TYPE foo_metric unknown
 # UNIT foo_metric seconds
 foo_metric 1.234


expfmt/encode_test.go
```diff
@@ -256,24 +257,16 @@ foo_metric 1.234
...
-# UNIT foo_metric_seconds seconds
-foo_metric_seconds 1.234
-`,
-		},
-		// 8: Simple Counter FmtOpenMetrics_1_0_0 unit opted out
+		// 7: Simple Counter FmtOpenMetrics_0_0_1
 		{
 			metric: metric1,
-			format: FmtOpenMetrics_1_0_0,
+			format: FmtOpenMetrics_0_0_1,
```
bwplotka: copy pasta?

expfmt/openmetrics_create.go
```diff
@@ -51,17 +50,6 @@ func WithCreatedLines() EncoderOption {
 	}
 }
 
-// WithUnit is an EncoderOption enabling a set unit to be written to the output
-// and to be added to the metric name, if it's not there already, as a suffix.
-// Without opting in this way, the unit will not be added to the metric name and,
-// on top of that, the unit will not be passed onto the output, even if it
-// were declared in the *dto.MetricFamily struct, i.e. even if in.Unit !=nil.
-func WithUnit() EncoderOption {
```
bwplotka, line 59: The only potential question is around compatibility for this. For client_golang we should be fine, we don't surface this option to users.

But for potentially downstream common users this might be explictly breaking (compile error). I don't know about any evidences of wide usage, also it could be a damaging option (renaming metrics), so I vote we kill it.

bwplotka (approved):

PR #888, opened 2026-03-26
version: Include race build tag

tcard:
When gathering info, tags are extracted directly from the `-tags` build setting. However, this doesn't include the `race` build tag that is implicitly set when building with `-race`.

With this change, we explicitly check the presence of the `race` build tag with conditional compilation, and include it in the tags list if present.

@roidelapluie @gotjosh


version/info.go
```diff
@@ -141,7 +144,11 @@ func computeRevision() (string, string) {
 			}
 		}
 		if v.Key == "-tags" {
-			tags = v.Value
+			if race {
+				tags = v.Value + ",race"
+			} else {
+				tags = v.Value
+			}
```
beorn7: I believe there is an easier way of doing this.

If the race detector is enabled, the `buildInfo.Settings` contain an entry `-race=true`. So we simply need something like
```go
		if v.Key == "-race" && v.Value == "true" {
```
with the appropriate flow later to assemble the right `tags` string, as we cannot expect a certain order of the `-tags` key and the `-race` key.

tcard: Thanks, done at bd03eb9.

roidelapluie:
i do not think that race is a build tag, should it be in its own category (and not shown when false)?

beorn7:
> i do not think that race is a build tag, should it be in its own category (and not shown when false)?

It is essentially treated as a build tag when using `//go:build !race`, but I guess you are right. It should be easy to add it as a separate line.

However, I'm not sure if we should really not show anything if race==false. Obviously, the extra line will be noise for many users, but on the other hand, consistency is important.

beorn7:
Useless information: Original meaning of "race-built": https://en.wikipedia.org/wiki/Race-built_galleon

beorn7:
Maybe the extra line in the template should be explicit (to avoid confusion), something like `race detector enabled: [true,false]`

PR #889, opened 2026-03-29
model: fix EscapeName incorrectly preserving colon in label names

### LINKED ISSUE #18380: `EscapeMetricFamily` incorrectly preserves `:` in label names
### What did you do?

When escaping label names, `EscapeMetricFamily` and `EscapeName` treat `:` (colon) as a
legacy-valid character and leave it unescaped. This is correct for **metric names**, but
wrong for **label names**. The Prometheus data model has always reserved `:` for metric
names only; it has never been valid in label names.

The result: a label named `app:instance-id` is rendered as `app:instance_id` under
`UnderscoreEscaping` (the default) instead of the correct `app_instance_id`.

---

## How to reproduce

```go
package main

import (
    "fmt"
    "log"
    "net/http"

    "github.com/prometheus/client_golang/prometheus/promhttp"
)

func main() {
    // Direct escaping — shows the bug in model.EscapeName
    fmt.Println(model.EscapeName("app:instance-id", model.UnderscoreEscaping))
    // Got:      "app:instance_id"   (colon preserved — BUG)
    // Expected: "app_instance_id"   (colon should be escaped for a label name)

    // HTTP endpoint — shows the bug in EscapeMetricFamily via promhttp
    counter := prometheus.NewCounterVec(
        prometheus.CounterOpts{Name: "requests_total", Help: "Total requests."},
        []string{"app:instance-id"},
    )
    prometheus.MustRegister(counter)
    counter.WithLabelValues("my-server").Inc()

    http.Handle("/metrics", promhttp.Handler())
    log.Fatal(http.ListenAndServe(":9090", nil))
}
```

```
$ curl -s http://localhost:9090/metrics | grep requests_total{
requests_total{app:instance_id="my-server"} 1
#              ^^^--- colon preserved, hyphen correctly escaped to _

$ curl -s -H 'Accept: text/plain; version=0.0.4; escaping=allow-utf-8' \
       http://localhost:9090/metrics | grep requests_total{
requests_total{"app:instance-id"="my-server"} 1
#               ^^^^^^^^^^^^^^^^^^--- verbatim original (correct for allow-utf-8)
```

---

## Affected escaping schemes

All schemes that go through `EscapeName` are affected wherever `isValidLegacyRune` decides
whether to leave a character in place:

| Scheme | Input `app:instance-id` | Got | Expected |
|---|---|---|---|
| `underscores` (default) | `app:instance-id` | `app:instance_id` | `app_instance_id` |
| `dots` | `app:instance-id` | `app:instance__id` | `app__instance__id` |
| `values` | `app:instance-id` | `U__app:instance_2d_id` | `U__app_3a_instance_2d_id` |
| `allow-utf-8` | `app:instance-id` | `"app:instance-id"` | `"app:instance-id"` ✓ |

`allow-utf-8` (`NoEscaping`) is not affected because it returns the name verbatim by
design.

---

## Comparison with the Python client

`prometheus_client` 0.24.1 gets this right. It defines two separate rune validators and
two separate escape entry points:

```python
# openmetrics/exposition.py:263
def _is_legacy_metric_rune(b, i):
    return _is_legacy_labelname_rune(b, i) or b == ':'   # colon OK for metric names

def _is_legacy_labelname_rune(b, i):
    return ('a' <= b <= 'z') or ('A' <= b <= 'Z') or b == '_' or \
           (b.isdigit() and i > 0)                        # NO colon

def escape_metric_name(s, escaping=UNDERSCORES): ...      # uses _is_legacy_metric_rune
def escape_label_name(s, escaping=UNDERSCORES):  ...      # uses _is_legacy_labelname_rune
```

Side-by-side output for `UnderscoreEscaping`:

| Input | Python (label) | Go (label) | Agree? |
|---|---|---|---|
| `app:instance-id` | `app_instance_id` | `app:instance_id` | **No** |
| `http.status:sum` | `http_status_sum` | `http_status:sum` | **No** |
| `my-metric-name` | `my_metric_name` | `my_metric_name` | Yes |
| `my_metric:name` | `my_metric_name` | `my_metric_name` | Yes |
| `my.metric.name` | `my_metric_name` | `my_metric_name` | Yes |

### What did you expect to see?

A label named `app:instance-id` escaped as `app_instance_id` under `UnderscoreEscaping`
— colon and hyphen both replaced with `_`, matching the legacy label name character set
`[a-zA-Z_][a-zA-Z0-9_]*`.

### What did you see instead? Under which circumstances?

`app:instance_id` — the colon is preserved, only the hyphen is replaced. This happens
with any escaping scheme other than `allow-utf-8`, whenever a label name contains a colon.

shankalpaLamichhane:
Fixes prometheus/prometheus#18380

Colons are only valid in metric names, not label names. EscapeName was using the same isValidLegacyRune for both contexts, causing colons to be preserved when escaping label names.

Added ValidationContext parameter to isValidLegacyRune and EscapeName so callers can specify metric or label context. EscapeMetricFamily updated to pass the correct context at each call site.

@roidelapluie @gotjosh

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -86,6 +86,15 @@ var _ interface {
 	fmt.Stringer
 } = new(ValidationScheme)
 
+// ValidationContext determines whether a name being validated or escaped
+// is a metric name or a label name. Colons are only valid in metric names.
+type ValidationContext int
+
+const (
+	ContextMetric ValidationContext = iota
+	ContextLabel
+)
+
 // String returns the string representation of s.
 func (s ValidationScheme) String() string {
 	switch s {
@@ -165,7 +174,7 @@ func (s ValidationScheme) IsValidMetricName(metricName string) bool {
 			return false
 		}
 		for i, b := range metricName {
-			if !isValidLegacyRune(b, i) {
+			if !isValidLegacyRune(b, i, ContextMetric) {
 				return false
 			}
 		}
@@ -353,7 +362,7 @@ func EscapeMetricFamily(v *dto.MetricFamily, scheme EscapingScheme) *dto.MetricF
 	if v.Name == nil || IsValidLegacyMetricName(v.GetName()) {
 		out.Name = v.Name
 	} else {
-		out.Name = proto.String(EscapeName(v.GetName(), scheme))
+		out.Name = proto.String(EscapeName(v.GetName(), scheme, ContextMetric))
 	}
 	for _, m := range v.Metric {
 		if !metricNeedsEscaping(m) {
@@ -378,7 +387,7 @@ func EscapeMetricFamily(v *dto.MetricFamily, scheme EscapingScheme) *dto.MetricF
 				}
 				escaped.Label = append(escaped.Label, &dto.LabelPair{
 					Name:  proto.String(MetricNameLabel),
-					Value: proto.String(EscapeName(l.GetValue(), scheme)),
+					Value: proto.String(EscapeName(l.GetValue(), scheme, ContextMetric)),
 				})
 				continue
 			}
@@ -387,7 +396,7 @@ func EscapeMetricFamily(v *dto.MetricFamily, scheme EscapingScheme) *dto.MetricF
 				continue
 			}
 			escaped.Label = append(escaped.Label, &dto.LabelPair{
-				Name:  proto.String(EscapeName(l.GetName(), scheme)),
+				Name:  proto.String(EscapeName(l.GetName(), scheme, ContextLabel)),
 				Value: l.Value,
 			})
 		}
@@ -412,7 +421,7 @@ func metricNeedsEscaping(m *dto.Metric) bool {
 // scheme. Depending on the rules of escaping, this may cause no change in the
 // string that is returned. (Especially NoEscaping, which by definition is a
 // noop). This function does not do any validation of the name.
-func EscapeName(name string, scheme EscapingScheme) string {
+func EscapeName(name string, scheme EscapingScheme, ctx ValidationContext) string {
 	if len(name) == 0 {
 		return name
 	}
@@ -421,11 +430,14 @@ func EscapeName(name string, scheme EscapingScheme) string {
 	case NoEscaping:
 		return name
 	case UnderscoreEscaping:
-		if IsValidLegacyMetricName(name) {
+		if ctx == ContextMetric && IsValidLegacyMetricName(name) {
+			return name
+		}
+		if ctx == ContextLabel && LegacyValidation.IsValidLabelName(name) {
 			return name
 		}
 		for i, b := range name {
-			if isValidLegacyRune(b, i) {
+			if isValidLegacyRune(b, i, ctx) {
 				escaped.WriteRune(b)
 			} else {
 				escaped.WriteRune('_')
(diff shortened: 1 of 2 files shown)


bwplotka:
Hello from the Bug Scrub

Please not mark this issue as `Fixes` (it will auto close the issue) - there's likely more to happen (e.g. import) to fully fix the issue.

model/metric.go
ywwg, line 384: these calls also need fixing

model/metric.go
```diff
@@ -555,8 +567,16 @@ func UnescapeName(name string, scheme EscapingScheme) string {
...
 }
 
-func isValidLegacyRune(b rune, i int) bool {
-	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || b == ':' || (b >= '0' && b <= '9' && i > 0)
+func isValidLegacyRune(r rune, i int, ctx ValidationContext) bool {
+	if (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || r == '_' || (r >= '0' && r <= '9' && i > 0) {
+		return true
+	}
+	// Colons are reserved for metric names (e.g. recording rules) only.
+	// They have never been valid in label names.
```
ywwg, line 575: No need to talk about history in comments

```suggestion
	// Colons are reserved for metric names (e.g. recording rules) only.
```

model/metric.go
```diff
@@ -555,8 +567,16 @@ func UnescapeName(name string, scheme EscapingScheme) string {
 	}
 }
 
-func isValidLegacyRune(b rune, i int) bool {
-	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || b == ':' || (b >= '0' && b <= '9' && i > 0)
+func isValidLegacyRune(r rune, i int, ctx ValidationContext) bool {
```
ywwg, line 570: `ctx` is go idiomatic for a context.Context, and it's confusing to use it for this simple enum value. this is more of a NameType than a validation context.

model/metric.go
```diff
@@ -86,6 +86,15 @@ var _ interface {
...
 
+// ValidationContext determines whether a name being validated or escaped
+// is a metric name or a label name. Colons are only valid in metric names.
+type ValidationContext int
+
+const (
+	ContextMetric ValidationContext = iota
+	ContextLabel
+)
+
```
ywwg, line 97: "context" is not quite the right word here, we want something that describes a concrete object, not the context that it exists in.  How about:

```suggestion
// NameClass determines whether a name being validated or escaped
// is a metric name or a label name.
type NameClass int

const (
	MetricClass NameClass = iota
	LabelClass
)

```

A few more options:  AttributeClass, NameType, ...

model/metric_test.go
```diff
@@ -386,6 +386,72 @@ func TestValidationScheme_IsMetricNameValid(t *testing.T) {
 	}
 }
 
+// TestEscapeNameLabelContext verifies that colons are correctly escaped when
+// EscapeName is called with ContextLabel.
+// Colons are reserved for metric names only and must never be preserved in escaped label names.
```
ywwg, line 391: Not strictly true. We do preserve colons in label names when we are doing NoEscaping.  (whereas quote marks *must* be escaped even during "NoEscaping"...)

```suggestion
// Colons are legacy valid in metric names, but not label names.
```

ywwg (changes requested):
Thank you for putting this together.  I made some notes -- please make sure to find all the places we are incorrectly calling IsValidLegacyMetricName on labels and make sure the tests are covering those cases as well.

shankalpaLamichhane:
Sorry for the long silence. I haven’t been able to get back to this.
I’m going to close this PR for now. I’ll open a new one later when I have time to properly address the feedback.
Thanks for the review!

Issue #893, opened 2026-04-07
expfmt: Support OpenMetrics 2.0


dashpole:
OpenMetrics 2.0 recently released an experimental release candidate: https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/.

We should add support for it in the expfmt encoder so that it is supported in prometheus/client_golang.

@ywwg @bwplotka @krajorama

PR #894, opened 2026-04-07
feat: implement gauge and counter support for OpenMetrics 2.0

dashpole:
Part 1 of https://github.com/prometheus/common/issues/893

Histograms and summaries are left for follow-ups.

Open Questions:

* How do we indicate that OM 2.0 support is experimental?

@ywwg @bwplotka @krajorama

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/munnerz/goautoneg"
 	dto "github.com/prometheus/client_model/go"
@@ -118,8 +119,10 @@ func NegotiateIncludingOpenMetrics(h http.Header) Format {
 		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
 			return FmtText + escapingScheme
 		}
-		if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion_0_0_1 || ver == OpenMetricsVersion_1_0_0 || ver == "") {
+		if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion_0_0_1 || ver == OpenMetricsVersion_1_0_0 || ver == OpenMetricsVersion_2_0_0 || ver == "") {
 			switch ver {
+			case OpenMetricsVersion_2_0_0:
+				return FmtOpenMetrics_2_0_0 + escapingScheme
 			case OpenMetricsVersion_1_0_0:
 				return FmtOpenMetrics_1_0_0 + escapingScheme
 			default:
@@ -181,6 +184,18 @@ func NewEncoder(w io.Writer, format Format, options ...EncoderOption) Encoder {
 			close: func() error { return nil },
 		}
 	case TypeOpenMetrics:
+		if strings.Contains(string(format), "version="+OpenMetricsVersion_2_0_0) {
+			return encoderCloser{
+				encode: func(v *dto.MetricFamily) error {
+					_, err := MetricFamilyToOpenMetrics20(w, model.EscapeMetricFamily(v, escapingScheme), options...)
+					return err
+				},
+				close: func() error {
+					_, err := FinalizeOpenMetrics(w)
+					return err
+				},
+			}
+		}
 		return encoderCloser{
 			encode: func(v *dto.MetricFamily) error {
 				_, err := MetricFamilyToOpenMetrics(w, model.EscapeMetricFamily(v, escapingScheme), options...)
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -60,8 +60,25 @@ func (ec encoderCloser) Close() error {
 // appropriate accepted type is found, FmtText is returned (which is the
 // Prometheus text format). This function will never negotiate FmtOpenMetrics,
 // as the support is still experimental. To include the option to negotiate
-// FmtOpenMetrics, use NegotiateIncludingOpenMetrics.
+// FmtOpenMetrics, use NegotiateIncludingOpenMetrics or NegotiateIncluding.
 func Negotiate(h http.Header) Format {
+	return NegotiateIncluding(h)
+}
+
+// NegotiateIncludingOpenMetrics works like Negotiate but includes
+// FmtOpenMetrics as an option for the result. Note that this function is
+// temporary and will disappear once FmtOpenMetrics is fully supported and as
+// such may be negotiated by the normal Negotiate function.
+func NegotiateIncludingOpenMetrics(h http.Header) Format {
+	return NegotiateIncluding(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1)
+}
+
+// NegotiateIncluding returns the Content-Type based on the given Accept header.
+// It automatically includes the default formats (Protobuf and Text) as options.
+// Additional formats provided in the arguments are also included as options,
+// and are checked in the order they are provided, respecting the preference
+// order in the Accept header.
+func NegotiateIncluding(h http.Header, additionalFormats ...Format) Format {
 	escapingScheme := Format(fmt.Sprintf("; escaping=%s", Format(model.NameEscapingScheme.String())))
 	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
 		if escapeParam := ac.Params[model.EscapingKey]; escapeParam != "" {
@@ -73,6 +90,13 @@ func Negotiate(h http.Header) Format {
 			}
 		}
 		ver := ac.Params["version"]
+
+		for _, f := range additionalFormats {
+			if matchFormat(ac, f) {
+				return f + escapingScheme
+			}
+		}
+
 		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
 			switch ac.Params["encoding"] {
 			case "delimited":
@@ -90,47 +114,39 @@ func Negotiate(h http.Header) Format {
 	return FmtText + escapingScheme
 }
 
-// NegotiateIncludingOpenMetrics works like Negotiate but includes
-// FmtOpenMetrics as an option for the result. Note that this function is
-// temporary and will disappear once FmtOpenMetrics is fully supported and as
-// such may be negotiated by the normal Negotiate function.
-func NegotiateIncludingOpenMetrics(h http.Header) Format {
-	escapingScheme := Format(fmt.Sprintf("; escaping=%s", Format(model.NameEscapingScheme.String())))
-	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
-		if escapeParam := ac.Params[model.EscapingKey]; escapeParam != "" {
-			switch Format(escapeParam) {
-			case model.AllowUTF8, model.EscapeUnderscores, model.EscapeDots, model.EscapeValues:
-				escapingScheme = Format("; escaping=" + escapeParam)
-			default:
-				// If the escaping parameter is unknown, ignore it.
-			}
+// matchFormat checks if a parsed accept clause matches a given Format.
+func matchFormat(ac goautoneg.Accept, f Format) bool {
+	parsed := goautoneg.ParseAccept(string(f))
+	if len(parsed) == 0 {
+		return false
+	}
+	target := parsed[0]
+
+	if ac.Type != target.Type || ac.SubType != target.SubType {
+		return false
+	}
+
+	// Default OpenMetrics version to OpenMetricsVersion_0_0_1
+	acVersion := ac.Params["version"]
+	if acVersion == "" && ac.Type+"/"+ac.SubType == OpenMetricsType {
+		acVersion = OpenMetricsVersion_0_0_1
+	}
+
+	// General param matching
+	for k, v := range target.Params {
+		if k == "charset" {
+			continue
 		}
-		ver := ac.Params["version"]
-		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
-			switch ac.Params["encoding"] {
-			case "delimited":
-				return FmtProtoDelim + escapingScheme
-			case "text":
-				return FmtProtoText + escapingScheme
-			case "compact-text":
-				return FmtProtoCompact + escapingScheme
-			}
+		acVal := ac.Params[k]
+		if k == "version" {
+			acVal = acVersion
 		}
-		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
-			return FmtText + escapingScheme
-		}
-		if ac.Type+"/"+ac.SubType == OpenMetricsType && (ver == OpenMetricsVersion_0_0_1 || ver == OpenMetricsVersion_1_0_0 || ver == OpenMetricsVersion_2_0_0 || ver == "") {
-			switch ver {
-			case OpenMetricsVersion_2_0_0:
-				return FmtOpenMetrics_2_0_0 + escapingScheme
-			case OpenMetricsVersion_1_0_0:
-				return FmtOpenMetrics_1_0_0 + escapingScheme
-			default:
-				return FmtOpenMetrics_0_0_1 + escapingScheme
-			}
+		if acVal != v {
+			return false
 		}
 	}
-	return FmtText + escapingScheme
+
+	return true
 }
 
 // NewEncoder returns a new encoder based on content type negotiation. All
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -58,27 +58,29 @@ func (ec encoderCloser) Close() error {
 
 // Negotiate returns the Content-Type based on the given Accept header. If no
 // appropriate accepted type is found, FmtText is returned (which is the
-// Prometheus text format). This function will never negotiate FmtOpenMetrics,
-// as the support is still experimental. To include the option to negotiate
-// FmtOpenMetrics, use NegotiateIncludingOpenMetrics or NegotiateIncluding.
+// Prometheus text format).
+//
+// Deprecated: Use NegotiateAccept(h, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
+// or specify only the formats supported by your server.
 func Negotiate(h http.Header) Format {
-	return NegotiateIncluding(h)
+	return NegotiateAccept(h, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
 }
 
 // NegotiateIncludingOpenMetrics works like Negotiate but includes
-// FmtOpenMetrics as an option for the result. Note that this function is
-// temporary and will disappear once FmtOpenMetrics is fully supported and as
-// such may be negotiated by the normal Negotiate function.
+// FmtOpenMetrics as an option for the result.
+//
+// Deprecated: Use NegotiateAccept(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
+// or specify only the formats supported by your server.
 func NegotiateIncludingOpenMetrics(h http.Header) Format {
-	return NegotiateIncluding(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1)
+	return NegotiateAccept(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
 }
 
-// NegotiateIncluding returns the Content-Type based on the given Accept header.
-// It automatically includes the default formats (Protobuf and Text) as options.
-// Additional formats provided in the arguments are also included as options,
-// and are checked in the order they are provided, respecting the preference
-// order in the Accept header.
-func NegotiateIncluding(h http.Header, additionalFormats ...Format) Format {
+// NegotiateAccept returns the Content-Type based on the given Accept header
+// and the list of accepted Formats provided by the caller, in order of preference.
+// If no accepted format matches the Accept header, it falls back to the text
+// format if present in the accepted list, or the first accepted format (or FmtText
+// if accepted is empty).
+func NegotiateAccept(h http.Header, accepted ...Format) Format {
 	escapingScheme := Format(fmt.Sprintf("; escaping=%s", Format(model.NameEscapingScheme.String())))
 	for _, ac := range goautoneg.ParseAccept(h.Get(hdrAccept)) {
 		if escapeParam := ac.Params[model.EscapingKey]; escapeParam != "" {
@@ -89,27 +91,20 @@ func NegotiateIncluding(h http.Header, additionalFormats ...Format) Format {
 				// If the escaping parameter is unknown, ignore it.
 			}
 		}
-		ver := ac.Params["version"]
 
-		for _, f := range additionalFormats {
+		for _, f := range accepted {
 			if matchFormat(ac, f) {
 				return f + escapingScheme
 			}
 		}
-
-		if ac.Type+"/"+ac.SubType == ProtoType && ac.Params["proto"] == ProtoProtocol {
-			switch ac.Params["encoding"] {
-			case "delimited":
-				return FmtProtoDelim + escapingScheme
-			case "text":
-				return FmtProtoText + escapingScheme
-			case "compact-text":
-				return FmtProtoCompact + escapingScheme
-			}
+	}
+	for _, f := range accepted {
+		if f.FormatType() == TypeTextPlain {
+			return f + escapingScheme
 		}
-		if ac.Type == "text" && ac.SubType == "plain" && (ver == TextVersion || ver == "") {
-			return FmtText + escapingScheme
-		}
+	}
+	if len(accepted) > 0 {
+		return accepted[0] + escapingScheme
 	}
 	return FmtText + escapingScheme
 }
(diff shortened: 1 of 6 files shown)


expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,285 @@
...
+		n, err = w.WriteString(" gauge\n")
+	case dto.MetricType_SUMMARY:
+		n, err = w.WriteString(" summary\n")
+	case dto.MetricType_UNTYPED:
+		n, err = w.WriteString(" unknown\n")
+	case dto.MetricType_HISTOGRAM:
+		n, err = w.WriteString(" histogram\n")
+	case dto.MetricType_GAUGE_HISTOGRAM:
+		n, err = w.WriteString(" gaugehistogram\n")
+	default:
```
bwplotka, line 118: Also info and stateset

expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,285 @@
...
+	"math"
+	"strconv"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// MetricFamilyToOpenMetrics20 converts a MetricFamily proto message into the
+// OpenMetrics text format version 2.0.0 and writes the resulting lines to 'out'.
+// It returns the number of bytes written and any error encountered.
+func MetricFamilyToOpenMetrics20(out io.Writer, in *dto.MetricFamily, options ...EncoderOption) (written int, err error) {
```
bwplotka, line 36: One way of letting users know it's experimental is by adding clear 

"// NOTE: This method implements OpenMetrics 2.0-rc.0 which is experimental. Breaking changes might happen in the future."


... assuming it even implements RC.0 at this point. If not let's mention what's implemented.

expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,285 @@
...
+		case dto.MetricType_SUMMARY:
+			if metric.Summary == nil {
+				return written, fmt.Errorf("expected summary in metric %s %s", name, metric)
+			}
+			n, err = writeCompositeSummary(w, name, metric)
+		case dto.MetricType_HISTOGRAM, dto.MetricType_GAUGE_HISTOGRAM:
+			if metric.Histogram == nil {
+				return written, fmt.Errorf("expected histogram in metric %s %s", name, metric)
+			}
+			n, err = writeCompositeHistogram(w, name, metric, metricType == dto.MetricType_GAUGE_HISTOGRAM)
```
bwplotka, line 193: Ditto (missing stateset and info)

expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,285 @@
...
+		written, err := w.Write(*bp)
+		numBufPool.Put(bp)
+		return written, err
+	}
+}
+
+// Stubs for Summary and Histogram
+
+func writeCompositeSummary(w enhancedWriter, name string, metric *dto.Metric) (int, error) {
+	_ = w
```
bwplotka, line 348: Sad that AI didn't find an easier way to do this but 🤷‍♀️ - doesn't matter here - we will replace later

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,115 @@
...
+			},
+			out: `# HELP http_requests_total Total number of HTTP requests.
+# TYPE http_requests_total counter
+http_requests_total{method="GET",code="200"} 1027 st@1234567890
+`,
+		},
+		{
+			name: "Gauge",
+			in: &dto.MetricFamily{
+				Name: proto.String("node_memory_Active_bytes"),
```
bwplotka:
```suggestion
				Name: proto.String("node_memory_active_bytes"),
```

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"bytes"
+	"math"
+	"testing"
+
+	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func TestCreateOpenMetrics20(t *testing.T) {
```
bwplotka, line 28: We probably need way more tests for edge cases - but  good start

bwplotka (commented):
For common, given v0 I would just comment on the method. If we want to be fancy we can generate comment in the exposition format too (:

Looks good so far, good start (see comments) - only question is do we iterate on this on main or some feature branch - no strong opinion, assuming commentary is clear.

Thanks

expfmt/openmetrics_2_0_create.go
dashpole: https://github.com/prometheus/client_model/blob/master/go/metrics.pb.go does not include info or stateset types.  I'll add TODOs instead.

expfmt/openmetrics_2_0_create.go
dashpole: Added

expfmt/openmetrics_2_0_create.go
dashpole: Added a TODO to support once it is supported in the protobuf version.

expfmt/openmetrics_2_0_create_test.go
dashpole: done

expfmt/openmetrics_2_0_create_test.go
dashpole: Added a few more tests.

ywwg (commented):
looks good to me so far. My general opinion is that if some scraper asks for openmetrics 2.0, they should get it... Hypothetically I could see requiring the version number "2.0.0rc" or "1.9.9" just in case?

dashpole:
Lots of discussion at the OM wg.  We will add a NegotiateIncluding() function that takes a list of supported protocols so we don't end up with NegotiateIncludingOpenMetricsAndOpenMetrics2 ... 😆 .  Then the client can decide when it wants to add support, and how it wants to gate it (behind `HandlerOpts.EnableOpenMetrics`?).

dashpole:
I added a NegotiateIncluding function.  I didn't deprecate the old functions (e.g. NegotiateIncludingOpenMetrics), but we can do that in a follow-up if we want to.

bwplotka:
I think that's great, the experimental factor at this level should be a commentary.

As per discussion on Slack I think we decided to follow:
1. Exposer should eventually switch to OM2 default, but only once ready (which means we recommend opt-in "exposed protocol" option) 
3. Scrapers should be opt-in for now, while OM2 is experimental

expfmt/encode.go
```diff
@@ -59,41 +60,25 @@ func (ec encoderCloser) Close() error {
...
-	return FmtText + escapingScheme
+	return NegotiateIncluding(h)
 }
 
 // NegotiateIncludingOpenMetrics works like Negotiate but includes
 // FmtOpenMetrics as an option for the result. Note that this function is
 // temporary and will disappear once FmtOpenMetrics is fully supported and as
 // such may be negotiated by the normal Negotiate function.
 func NegotiateIncludingOpenMetrics(h http.Header) Format {
+	return NegotiateIncluding(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1)
```
bwplotka: 👍🏽

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,389 @@
...
+				Type: dto.MetricType_COUNTER.Enum(),
+				Metric: []*dto.Metric{
+					{
+						Label: []*dto.LabelPair{
+							{Name: proto.String("method"), Value: proto.String("GET")},
+							{Name: proto.String("code"), Value: proto.String("200")},
+						},
+						Counter: &dto.Counter{
+							Value:            proto.Float64(1027),
+							CreatedTimestamp: &timestamppb.Timestamp{Seconds: 1234567890},
```
krajorama, line 48: Let's add a test for counters where ST has ms resolution, e.g.:
```
CreatedTimestamp: &timestamppb.Timestamp{Seconds: 1234567890, Nanos: 987654321},
```

To test that the result has the precision:
```
http_requests_total{method="GET",code="200"} 1027 st@1234567890.9876542
```

Maybe we can discuss if we only want prec==3 for this case. But keep that discussion separate from the PR.

expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,291 @@
...
+		ts := metric.Counter.CreatedTimestamp
+		n, err = writeOpenMetrics20Timestamp(w, float64(ts.GetSeconds())+float64(ts.GetNanos())/1e9)
+		written += n
+		if err != nil {
+			return written, err
+		}
+	}
+
+	if exemplar != nil && len(exemplar.Label) > 0 {
+		n, err = writeExemplar(w, exemplar)
```
krajorama: I don't think we can call this function as is. It treats the exemplar timestamp as optional. I think we need to drop exemplars without timestamp for om2.0. https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#exemplars

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,389 @@
...
+		{
+			name: "CounterWithExemplar",
+			in: &dto.MetricFamily{
+				Name: proto.String("http_requests_total"),
+				Type: dto.MetricType_COUNTER.Enum(),
+				Metric: []*dto.Metric{
+					{
+						Counter: &dto.Counter{
+							Value: proto.Float64(1027),
+							Exemplar: &dto.Exemplar{
```
krajorama, line 175: Invalid test, the exemplar must have a timestamp.

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,389 @@
...
+									{Name: proto.String("trace_id"), Value: proto.String("1234")},
+								},
+								Value: proto.Float64(1),
+							},
+						},
+					},
+				},
+			},
+			out: `# TYPE http_requests_total counter
+http_requests_total 1027 # {trace_id="1234"} 1.0
```
krajorama: Let's also add start time to verify the order.
And possibly the sample timestamp as well, although I don't know how to do that.

expfmt/encode.go
```diff
@@ -59,41 +60,25 @@ func (ec encoderCloser) Close() error {
...
+}
+
+// NegotiateIncluding returns the Content-Type based on the given Accept header.
+// It automatically includes the default formats (Protobuf and Text) as options.
+// Additional formats provided in the arguments are also included as options,
+// and are checked in the order they are provided, respecting the preference
+// order in the Accept header.
+func NegotiateIncluding(h http.Header, additionalFormats ...Format) Format {
```
krajorama: The format we pass has different string from protocols setting in Prometheus scrape config. We'll make the mapping from user string to these formats in client_golang, right?

krajorama (changes requested):

expfmt/openmetrics_2_0_create.go
krajorama: According to LLM, client_golang at least always sets the timestamp, so it will not be breaking to do this, i.e. drop exemplars without timestamp.

expfmt/encode.go
dashpole: Yes, we will have to do that mapping if we choose to allow client_golang users to provide strings instead of these constants. We can evaluate that in the client_golang PR.

expfmt/openmetrics_2_0_create_test.go
dashpole: fixed

expfmt/openmetrics_2_0_create.go
dashpole: Added a new function for OM 2.0 timestamps.

expfmt/openmetrics_2_0_create_test.go
dashpole: done

expfmt/openmetrics_2_0_create_test.go
dashpole: added

dashpole:
In the spirit of @krajorama's feedback, i've updated the implementation to perform some additional validation. It checks for newline characters, NaN and negative counter values, valid timestamps, empty exemplar label keys.  It does not validate UTF-8 for strings, or duplicate label keys or duplicate label sets because those are expensive to validate. This does more validation than OM 1.0, but still leaves some things not validated.  I also ended up writing benchmarks as part of that work, which i've split into https://github.com/prometheus/common/pull/943.

The benchmark results from #943 run against this PR are:

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
                                             │ /tmp/bench.txt │
                                             │     sec/op     │
ConvertMetricFamily/TEXT/COUNTER-24               775.5n ± 1%
ConvertMetricFamily/OM1.0/COUNTER-24              1.664µ ± 1%
ConvertMetricFamily/OM2.0/COUNTER-24              1.617µ ± 1%
ConvertMetricFamily/TEXT/GAUGE-24                 761.5n ± 1%
ConvertMetricFamily/OM1.0/GAUGE-24                1.091µ ± 6%
ConvertMetricFamily/OM2.0/GAUGE-24                1.003µ ± 3%
ConvertMetricFamily/TEXT/UNTYPED-24               680.6n ± 1%
ConvertMetricFamily/OM1.0/UNTYPED-24              923.8n ± 3%
ConvertMetricFamily/OM2.0/UNTYPED-24              883.1n ± 1%
ConvertMetricFamily/TEXT/SUMMARY-24               3.207µ ± 1%
ConvertMetricFamily/OM1.0/SUMMARY-24              3.965µ ± 1%
ConvertMetricFamily/TEXT/HISTOGRAM-24             4.532µ ± 2%
ConvertMetricFamily/OM1.0/HISTOGRAM-24            5.582µ ± 2%
ConvertMetricFamily/TEXT/GAUGE_HISTOGRAM-24       3.414µ ± 1%
ConvertMetricFamily/OM1.0/GAUGE_HISTOGRAM-24      3.852µ ± 1%
geomean                                           1.741µ

                                             │ /tmp/bench.txt │
                                             │      B/op      │
ConvertMetricFamily/TEXT/COUNTER-24              0.000 ± 0%
ConvertMetricFamily/OM1.0/COUNTER-24             33.00 ± 0%
ConvertMetricFamily/OM2.0/COUNTER-24             0.000 ± 0%
ConvertMetricFamily/TEXT/GAUGE-24                0.000 ± 0%
ConvertMetricFamily/OM1.0/GAUGE-24               1.000 ± 0%
ConvertMetricFamily/OM2.0/GAUGE-24               0.000 ± 0%
ConvertMetricFamily/TEXT/UNTYPED-24              0.000 ± 0%
ConvertMetricFamily/OM1.0/UNTYPED-24             1.000 ± 0%
ConvertMetricFamily/OM2.0/UNTYPED-24             0.000 ± 0%
ConvertMetricFamily/TEXT/SUMMARY-24              56.00 ± 0%
ConvertMetricFamily/OM1.0/SUMMARY-24             57.00 ± 0%
ConvertMetricFamily/TEXT/HISTOGRAM-24            288.0 ± 0%
ConvertMetricFamily/OM1.0/HISTOGRAM-24           289.0 ± 0%
ConvertMetricFamily/TEXT/GAUGE_HISTOGRAM-24      160.0 ± 0%
ConvertMetricFamily/OM1.0/GAUGE_HISTOGRAM-24     161.0 ± 0%
geomean                                                     ¹
¹ summaries must be >0 to compute geomean

                                             │ /tmp/bench.txt │
                                             │   allocs/op    │
ConvertMetricFamily/TEXT/COUNTER-24              0.000 ± 0%
ConvertMetricFamily/OM1.0/COUNTER-24             2.000 ± 0%
ConvertMetricFamily/OM2.0/COUNTER-24             0.000 ± 0%
ConvertMetricFamily/TEXT/GAUGE-24                0.000 ± 0%
ConvertMetricFamily/OM1.0/GAUGE-24               1.000 ± 0%
ConvertMetricFamily/OM2.0/GAUGE-24               0.000 ± 0%
ConvertMetricFamily/TEXT/UNTYPED-24              0.000 ± 0%
ConvertMetricFamily/OM1.0/UNTYPED-24             1.000 ± 0%
ConvertMetricFamily/OM2.0/UNTYPED-24             0.000 ± 0%
ConvertMetricFamily/TEXT/SUMMARY-24              2.000 ± 0%
ConvertMetricFamily/OM1.0/SUMMARY-24             3.000 ± 0%
ConvertMetricFamily/TEXT/HISTOGRAM-24            6.000 ± 0%
ConvertMetricFamily/OM1.0/HISTOGRAM-24           7.000 ± 0%
ConvertMetricFamily/TEXT/GAUGE_HISTOGRAM-24      5.000 ± 0%
ConvertMetricFamily/OM1.0/GAUGE_HISTOGRAM-24     6.000 ± 0%
geomean                                                     ¹
¹ summaries must be >0 to compute geomean
```

expfmt/expfmt.go
```diff
@@ -59,6 +61,8 @@ const (
 	// Deprecated: Use expfmt.NewFormat(expfmt.TypeOpenMetrics) instead.
 	//nolint:revive // Allow for underscores.
 	FmtOpenMetrics_1_0_0 Format = OpenMetricsType + `; version=` + OpenMetricsVersion_1_0_0 + `; charset=utf-8`
+	//nolint:revive // Allow for underscores.
+	FmtOpenMetrics_2_0_0 Format = OpenMetricsType + `; version=` + OpenMetricsVersion_2_0_0 + `; charset=utf-8`
```
bwplotka: The PR introduces `FmtOpenMetrics_2_0_0` as an exported constant. Adjacent older formats are explicitly marked as `// Deprecated: Use expfmt.NewFormat(expfmt.TypeOpenMetrics) instead.`. Introducing a new format-specific exported constant reproduces the very API footprint that maintainers actively deprecated. Consider changing to unexported `fmtOpenMetrics_2_0_0`.

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,529 @@
...
+				t.Fatal(err)
+			}
+			if buf.String() != tc.out {
+				t.Errorf("expected %q, got %q", tc.out, buf.String())
+			}
+		})
+	}
+}
+
+func TestCreateOpenMetrics20_Errors(t *testing.T) {
```
bwplotka, line 315: Missing test for Counter CreatedTimestamp validation (Edge case coverage). Add a test case in `TestCreateOpenMetrics20_Errors` where `Metric.Counter.CreatedTimestamp` is mathematically invalid (e.g., negative seconds or out-of-bounds) to assert it cleanly returns the validation error message.

expfmt/openmetrics_2_0_create.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+	}
+	return nil
+}
+
+func containsRawNewline(s string) bool {
+	return strings.IndexByte(s, '\n') >= 0 || strings.IndexByte(s, '\r') >= 0
+}
+
+func validateExemplar20(e *dto.Exemplar) error {
```
bwplotka, line 382: Missing test for Exemplar Timestamp validation (Edge case coverage). Add a test case with an invalid Exemplar Timestamp (e.g. invalid protobuf timestamp values) to ensure `validateExemplar20` correctly captures and bubbles up the `CheckValid()` error.

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,529 @@
...
+				},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			_, err := MetricFamilyToOpenMetrics20(&buf, tc.in)
+			if err == nil {
```
bwplotka, line 544: Update the `err == nil` logic to strictly check if the actual `err.Error()` matches the *exact* expected error message or a substring representing it. Currently, the test blindly accepts any generic non-nil error, masking regressions.

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -0,0 +1,529 @@
...
+			out: `# TYPE "你好_total" counter
+{"你好_total","🌎"="🌍"} 1027
+`,
+		},
+	}
+
+	for _, scenario := range scenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			_, err := MetricFamilyToOpenMetrics20(&buf, scenario.in)
```
bwplotka: Weak Unit Test Assertions on Written Bytes count. The functions return the number of bytes written, but this test suppresses it with `_, err := ...` . Verify that the returned byte count equals `len(buf.String())` / `len(expected)` to ensure the returned byte counting arithmetic functions properly.

expfmt/encode.go
```diff
@@ -59,41 +60,25 @@ func (ec encoderCloser) Close() error {
...
-	}
-	return FmtText + escapingScheme
+	return NegotiateIncluding(h)
 }
 
 // NegotiateIncludingOpenMetrics works like Negotiate but includes
 // FmtOpenMetrics as an option for the result. Note that this function is
 // temporary and will disappear once FmtOpenMetrics is fully supported and as
 // such may be negotiated by the normal Negotiate function.
 func NegotiateIncludingOpenMetrics(h http.Header) Format {
```
bwplotka, line 74: Can we deprecate Negotiate favor of the new function?

expfmt/encode.go
bwplotka: nit: Since we bake new API, what about something more explicit like: `NegotiateAccept(h http.Header, accepted ...Format) Format {` that would allow us to provide full list vs speculating what's default what's not (especially when default change?)

expfmt/encode.go
```diff
@@ -118,18 +110,45 @@ func NegotiateIncludingOpenMetrics(h http.Header) Format {
...
+	if len(parsed) == 0 {
+		return false
+	}
+	target := parsed[0]
+
+	if ac.Type != target.Type || ac.SubType != target.SubType {
+		return false
+	}
+
+	// Default OpenMetrics version to OpenMetricsVersion_0_0_1
```
bwplotka:
```suggestion
	// Default OpenMetrics version to OpenMetricsVersion_0_0_1.
```

expfmt/encode.go
```diff
@@ -118,18 +110,45 @@ func NegotiateIncludingOpenMetrics(h http.Header) Format {
...
+	acVersion := ac.Params["version"]
+	if acVersion == "" && ac.Type+"/"+ac.SubType == OpenMetricsType {
+		acVersion = OpenMetricsVersion_0_0_1
+	}
+
+	// General param matching
```
bwplotka:
```suggestion
	// General param matching.
```

bwplotka (approved):
Looks good, just one last API question

expfmt/encode.go
dashpole: @krajorama @ywwg LMK if either of you have objections.

expfmt/encode.go
dashpole: We will still have to have PrometheusText0.0.4 as a final fallback, but I think this is a good change.

From https://github.com/prometheus/docs/blob/main/docs/instrumenting/content_negotiation.md#selection-of-format:

> If no fallback is specified, the target MUST use PrometheusText0.0.4 as a last resort.

expfmt/encode.go
dashpole: addressed in 75aa192

expfmt/encode.go
dashpole: addressed in 75aa192

expfmt/expfmt.go
dashpole: addressed in a28874d

expfmt/openmetrics_2_0_create_test.go
dashpole: addressed in 8d97fb3

expfmt/openmetrics_2_0_create_test.go
dashpole: addressed in 9362a83

expfmt/openmetrics_2_0_create.go
dashpole: addressed in f9d3970

expfmt/openmetrics_2_0_create_test.go
dashpole: addressed in 525f78e

expfmt/encode.go
```diff
@@ -57,10 +58,29 @@ func (ec encoderCloser) Close() error {
...
 func Negotiate(h http.Header) Format {
+	return NegotiateAccept(h, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
+}
+
+// NegotiateIncludingOpenMetrics works like Negotiate but includes
+// FmtOpenMetrics as an option for the result.
+//
+// Deprecated: Use NegotiateAccept(h, FmtOpenMetrics_1_0_0, FmtOpenMetrics_0_0_1, FmtProtoDelim, FmtProtoText, FmtProtoCompact, FmtText)
+// or specify only the formats supported by your server.
+func NegotiateIncludingOpenMetrics(h http.Header) Format {
```
bwplotka, line 74: Let's remove if we want to use NegotiateAccept - this is a new function

bwplotka (approved):
One thing otherwise LGTM, let's iterate. Thanks!

expfmt/encode.go
dashpole: This was added for OpenMetrics 1.0, so its been around a while. Are you sure you want this removed without deprecation?

bwplotka: ah, I didn't notice you move it around -- no let's do this properly

bwplotka (approved):
Thanks!

PR #897, opened 2026-04-13
config: guard against nil oauth2 credential in RoundTrip

### LINKED ISSUE #790: toSecret' "return nil, nil" is error-prone
here https://github.com/prometheus/common/blob/098669c08af8eee3da33d1d4a4f242a5242cccb8/config/http_config.go#L781-L797

because the error is nil, we may/would be moving nils around, leading to nil pointer dereference

may be the cause of https://github.com/prometheus/prometheus/issues/16622

alliasgher:
## Summary

`toSecret` returns `(nil, nil)` when no credential source is configured. Most callers guarded the returned `SecretReader` with `!= nil` before invoking `Fetch`, but `oauth2RoundTripper.RoundTrip` reached directly into `rt.oauthCredential.Immutable()` and would nil-deref panic when someone supplied an `oauth2` block with no client-secret source (likely the proximate cause referenced by prometheus/prometheus#16622).

This PR:

1. Adds an explicit nil-guard in `oauth2RoundTripper.RoundTrip` that returns `errors.New("oauth2 client secret is required")` instead of panicking.
2. Documents `toSecret`'s nil-return contract so future callers explicitly acknowledge it.

Fixes #790

## Test plan

- [x] `go build ./...`
- [x] `go test ./config/`
- [x] `gofmt` clean

Signed-off-by: Ali <alliasgher123@gmail.com>

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -828,10 +828,7 @@ func (*refSecret) Immutable() bool {
 }
 
 // toSecret returns a SecretReader from one of the given sources, assuming exactly
-// one or none of the sources are provided. When no source is provided it
-// returns (nil, nil); callers MUST guard the returned reader with a nil check
-// before invoking any method on it — see the nil-deref issue tracked at
-// https://github.com/prometheus/common/issues/790.
+// one or none of the sources are provided.
 func toSecret(secretManager SecretManager, text Secret, file, ref string) (SecretReader, error) {
 	if text != "" {
 		return NewInlineSecret(string(text)), nil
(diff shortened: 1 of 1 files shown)


config/http_config.go
```diff
@@ -828,7 +828,10 @@ func (*refSecret) Immutable() bool {
 }
 
 // toSecret returns a SecretReader from one of the given sources, assuming exactly
-// one or none of the sources are provided.
+// one or none of the sources are provided. When no source is provided it
```
roidelapluie: I would not add this comment, it seems un-necessary.

config/http_config.go
```diff
@@ -1053,6 +1056,15 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		needsInit bool
 	)
 
+	// oauthCredential can be nil when the caller constructed an oauth2
```
roidelapluie: I think this comment is also not useful -- maybe it should just mention "this should not happen when config is the normal prometheus path as config should be valid"

config/http_config.go
alliasgher: Done — dropped the extended documentation from the `toSecret` comment; the original one-liner is back.

config/http_config.go
alliasgher: Updated to: "This should not happen when config goes through the normal Prometheus validation path, but guard against a nil credential to avoid a panic."

roidelapluie:
Thanks!

machine424:
Thanks for this.
But I'm afraid, as I mentioned in the issue, what needs to be fixed is the `toSecret()` itself; returning `nil, nil` is confusing and error prone. And nothing would prevent other callers from misusing it.

PR #898, opened 2026-04-15
config: change NewOAuth2RoundTripper to accept variadic HTTPClientOption

alliasgher:
## Summary

`NewOAuth2RoundTripper` previously accepted `*httpClientOptions` as its fourth
parameter — an **unexported** type. This made the function effectively
uncallable with a non-nil options argument from any package outside
`prometheus/common/config`, forcing downstream users (e.g. Grafana Alloy) to
resort to reflection or `unsafe` pointer tricks.

Reported upstream in prometheus/prometheus#18329.

## Change

Replace the unexported `*httpClientOptions` parameter with variadic
`...HTTPClientOption` — the **exported** interface already used by
`NewRoundTripperFromConfigWithContext`:

```go
// Before (unusable externally with non-nil opts):
func NewOAuth2RoundTripper(cred SecretReader, cfg *OAuth2, next http.RoundTripper, opts *httpClientOptions) http.RoundTripper

// After (fully usable from any package):
func NewOAuth2RoundTripper(cred SecretReader, cfg *OAuth2, next http.RoundTripper, optFuncs ...HTTPClientOption) http.RoundTripper
```

Internally the function builds its own `httpClientOptions` copy from the supplied options starting from `defaultHTTPClientOptions`, so behaviour is identical to before when no options are passed.

## Call sites updated

- Internal call in `NewRoundTripperFromConfigWithContext`: passes `optFuncs...` instead of `&opts`
- Three test call sites that passed `&defaultHTTPClientOptions` now pass no options (equivalent — defaults are unchanged)

## Tests

```
go test ./config/...   # all pass
```

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -717,7 +717,7 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
 					return nil, fmt.Errorf("unable to use client secret: %w", err)
 				}
 			}
-			rt = NewOAuth2RoundTripper(oauthCredential, cfg.OAuth2, rt, &opts)
+			rt = NewOAuth2RoundTripper(oauthCredential, cfg.OAuth2, rt, optFuncs...)
 		}
 
 		if cfg.HTTPHeaders != nil {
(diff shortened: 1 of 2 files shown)


roidelapluie (approved):

roidelapluie:
Thanks!

PR #901, opened 2026-04-17
config: strip credentials on cross-host redirects

roidelapluie:
When FollowRedirects is true, credentials (Authorization header, Cookie headers set via HTTPHeaders) were forwarded to any redirect target, including cross-host redirects.

Fix by marking cross-host redirects in the request context inside CheckRedirect, then having each credential round-tripper (bearer, basic auth, OAuth2) skip adding credentials for marked requests. A new sensitiveHeadersStripRT also strips sensitive headers added by headersRoundTripper on cross-host redirects.

This aligns to Go's HTTP client behaviour.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -721,6 +721,14 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
 		}
 
 		if cfg.HTTPHeaders != nil {
+			// Strip sensitive headers added by headersRoundTripper on cross-host
+			// redirects before they reach the transport. Only needed when
+			// redirects are actually followed; when FollowRedirects is false
+			// CheckRedirect returns ErrUseLastResponse immediately so there are
+			// no subsequent requests.
+			if cfg.FollowRedirects {
+				rt = &sensitiveHeadersStripRT{next: rt}
+			}
 			rt = NewHeadersRoundTripper(cfg.HTTPHeaders, rt)
 		}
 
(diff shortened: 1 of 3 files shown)


config/http_config.go
```diff
@@ -717,10 +717,18 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
...
+			rt = NewOAuth2RoundTripper(oauthCredential, cfg.OAuth2, rt, &opts)
 		}
 
 		if cfg.HTTPHeaders != nil {
+			// Strip sensitive headers added by headersRoundTripper on cross-host
+			// redirects before they reach the transport. Only needed when
+			// redirects are actually followed; when FollowRedirects is false
+			// CheckRedirect returns ErrUseLastResponse immediately so there are
+			// no subsequent requests.
+			if cfg.FollowRedirects {
```
bwplotka, line 729: I wonder, if this is a safe practice, why Go is not doing this by default? 🤔

roidelapluie: > I wonder, if this is a safe practice, why Go is not doing this by default? 🤔

Go IS doing this by default.

roidelapluie: https://github.com/golang/go/blob/go1.24.0/src/net/http/client.go#L41-L50
https://github.com/golang/go/blob/go1.24.0/src/net/http/client.go#L682-L687
https://github.com/golang/go/blob/go1.24.0/src/net/http/client.go#L806-L813
https://github.com/golang/go/blob/go1.24.0/src/net/http/client.go#L990-L1031

bwplotka: nice, thanks

config/http_config.go
```diff
@@ -1101,6 +1093,9 @@ func (rt *oauth2RoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 	rt.mtx.RLock()
 	currentRT := rt.lastRT
 	rt.mtx.RUnlock()
+	if isCrossHostRedirect(req) {
```
bwplotka, line 1096: I was surprised we can't cut off sooner, but I guess we don't know Base here .. or is it rt.client?

bwplotka (approved):
Thanks!

PR #907, opened 2026-05-12
model: fix EscapeLabelName to not preserve colon in label names

### LINKED ISSUE #18380: `EscapeMetricFamily` incorrectly preserves `:` in label names

ogulcanaydogan:
Colons are reserved for metric names only per the Prometheus data model. `EscapeMetricFamily` and `metricNeedsEscaping` were using `IsValidLegacyMetricName` and `EscapeName` for label name contexts, causing colons to be preserved when they should be escaped.

Fixes prometheus/prometheus#18380

## Changes

- Add `isValidLegacyLabelRune`: same as `isValidLegacyRune` but excludes `:`
- Add `IsValidLegacyLabelName`: uses the label-specific rune validator
- Add `EscapeLabelName`: mirrors `EscapeName` but applies label name rules
- Update `EscapeMetricFamily` to use `IsValidLegacyLabelName` and `EscapeLabelName` for label names
- Update `metricNeedsEscaping` to use `IsValidLegacyLabelName` for label names

## Behavior

| Input | Scheme | Before | After |
|---|---|---|---|
| `app:instance-id` | underscores | `app:instance_id` | `app_instance_id` |
| `http.status:sum` | underscores | `http_status:sum` | `http_status_sum` |
| `app:instance-id` | values | `U__app:instance_2d_id` | `U__app_3a_instance_2d_id` |
| `requests:total` (metric name) | underscores | `requests:total` | `requests:total` (unchanged) |

Metric names with colons continue to work correctly. Only label name escaping is affected.

Builds on the approach discussed in #889.

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -321,6 +321,75 @@ func IsValidLegacyMetricName(n string) bool {
 	return LegacyValidation.IsValidMetricName(n)
 }
 
+// IsValidLegacyLabelName reports whether n is a valid label name under the
+// legacy Prometheus data model. Unlike metric names, label names must not
+// contain ':'.
+func IsValidLegacyLabelName(n string) bool {
+	return LegacyValidation.IsValidLabelName(n)
+}
+
+// EscapeLabelName escapes the incoming label name according to the provided
+// escaping scheme. It behaves like EscapeName but applies label-name rules:
+// ':' is not a valid character in a label name and is always escaped.
+func EscapeLabelName(name string, scheme EscapingScheme) string {
+	if len(name) == 0 {
+		return name
+	}
+	var escaped strings.Builder
+	switch scheme {
+	case NoEscaping:
+		return name
+	case UnderscoreEscaping:
+		if IsValidLegacyLabelName(name) {
+			return name
+		}
+		for i, b := range name {
+			if isValidLegacyLabelRune(b, i) {
+				escaped.WriteRune(b)
+			} else {
+				escaped.WriteRune('_')
+			}
+		}
+		return escaped.String()
+	case DotsEscaping:
+		for i, b := range name {
+			switch {
+			case b == '_':
+				escaped.WriteString("__")
+			case b == '.':
+				escaped.WriteString("_dot_")
+			case isValidLegacyLabelRune(b, i):
+				escaped.WriteRune(b)
+			default:
+				escaped.WriteString("__")
+			}
+		}
+		return escaped.String()
+	case ValueEncodingEscaping:
+		if IsValidLegacyLabelName(name) {
+			return name
+		}
+		escaped.WriteString("U__")
+		for i, b := range name {
+			switch {
+			case b == '_':
+				escaped.WriteString("__")
+			case isValidLegacyLabelRune(b, i):
+				escaped.WriteRune(b)
+			case !utf8.ValidRune(b):
+				escaped.WriteString("_FFFD_")
+			default:
+				escaped.WriteRune('_')
+				escaped.WriteString(strconv.FormatInt(int64(b), 16))
+				escaped.WriteRune('_')
+			}
+		}
+		return escaped.String()
+	default:
+		panic(fmt.Sprintf("invalid escaping scheme %d", scheme))
+	}
+}
+
 // EscapeMetricFamily escapes the given metric names and labels with the given
 // escaping scheme. Returns a new object that uses the same pointers to fields
 // when possible and creates new escaped versions so as not to mutate the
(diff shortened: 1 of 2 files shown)


ogulcanaydogan:
Friendly bump. This is a small correctness fix for EscapeLabelName and CI is green. Let me know if you'd like any changes.

PR #911, opened 2026-05-27
Fix: apply DialContextFunc to OAuth2 token-fetch transport

yuri-tceretian:
### Problem

function `newOauth2TokenSource` builds its own `http.Transport` to fetch tokens but doesn't set DialContext on it. Any DialContextFunc passed via WithDialContextFunc is applied to the main request transport but silently skipped for the token endpoint.

### Solution
The fix is a one-liner - set DialContext on the token transport the same way it's set on the main one. When dialContextFunc is nil the behaviour is unchanged since http.Transport falls back to its default dialer.


bwplotka (approved):
Nice, thanks!

PR #924, opened 2026-06-17
config: clarify sensitive redirect headers match net/http

roidelapluie:
The stripped header list (including Proxy-Authorization and Proxy-Authenticate) now matches makeHeadersCopier in net/http, which gained the Proxy-* entries in the fix for CVE-2025-4673.


krajorama (approved):
stamp

PR #925, opened 2026-06-17
config: resolve LoadHTTPConfigFile paths relative to the config file

roidelapluie:
LoadHTTPConfigFile called SetDirectory with filepath.Dir(filepath.Dir(filename)), resolving relative file paths (http_headers files, *_file credentials) against the config file's grandparent directory instead of its own directory. This contradicts the SetDirectory contract and every other config loader.

The behavior was masked by the testdata fixtures, which prefixed file paths with the config file's own directory name (e.g. files: [testdata/headers-file] for a config living in testdata/) to compensate. Drop those redundant prefixes and add a regression test that loads a config from a temp dir and asserts a plain relative path resolves next to the config file.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -325,7 +325,7 @@ func LoadHTTPConfigFile(filename string) (*HTTPClientConfig, []byte, error) {
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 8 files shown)


krajorama (commented):
Fix looks ok, but it should be mentioned in the changelog as it is breaking.

PR #932, opened 2026-06-24
route: add support for the QUERY HTTP method

roidelapluie:
QUERY is a safe, idempotent method that carries a request body (RFC 10008). It is not yet part of net/http, so register it using a locally defined MethodQuery constant.


krajorama (approved):
LGTM, but I'm no expert in go networking

PR #935, opened 2026-06-30
version: fix Slog godoc example to compile

sueun-dev:
The `Example` in the godoc for `version.Slog()` passes `promslog.Config{}` by value:

```go
logger := promslog.New(promslog.Config{})
logger.Info("Starting Prometheus Server", version.Slog()...)
```

but `promslog.New` takes a pointer:

```go
func New(config *Config) *slog.Logger
```

so the documented snippet does not compile. Building it verbatim fails with:

```
cannot use promslog.Config{} (value of struct type promslog.Config)
  as *promslog.Config value in argument to promslog.New
```

This changes it to `&promslog.Config{}`, matching the real call sites (`promslog/slog.go`, `promslog/slog_test.go`).

I also added a compiled `ExampleSlog` so the documented usage stays in sync with the API. It imports `promslog` only from the test file, so `version` gains no new runtime dependency. `gofmt`, `go vet ./version/...`, and `go test ./...` pass.

### DIFF
--- a/version/example_test.go
+++ b/version/example_test.go
@@ -1,0 +1,26 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


version/example_test.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2024 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright The Prometheus Authors
```

SuperQ (approved):
Thanks! Minor nit on the new file copyright.

SuperQ (changes requested):
Actually, I don't see the point of adding the test file. How does it actively enforce the documentation match?

PR #944, opened 2026-07-07
config: add OAuth2 token_url_file

dtrejod:
Adds a `token_url_file` field to the OAuth2 HTTP client config, alongside the existing `token_url` config. The token URL is read from the file when the OAuth2 token source is (re)built, so deploy-time-injected values (e.g. Helm-managed files, mounted Secrets/ConfigMaps) work without needing the URL to be known at config-authoring time. 

Exactly one of `token_url` or `token_url_file` must be set. Configs that set both, or neither, fail validation.

See https://github.com/prometheus/alertmanager/issues/4759


dtrejod:
@SuperQ / @roidelapluie Would appreciate a review on this contribution 🙏

PR #951, opened 2026-07-21
expfmt: add NegotiateFormatsIncludingOpenMetrics

dpacgdm:
## Summary
- Add `NegotiateFormatsIncludingOpenMetrics` returning all recognised Accept formats in preference order
- Keep `NegotiateIncludingOpenMetrics` as a wrapper that returns the first entry
- Add a regression test for proto → OpenMetrics → text fallback

Addresses prometheus/common#928. This gives servers that support only a subset of formats a clean way to walk the client's preference list (example in the issue: protobuf preferred but unsupported → OpenMetrics → text).

Callers such as kube-state-metrics can use this for content negotiation; a local workaround is also possible without this API.

## Test plan
- [x] `go test ./expfmt/ -run Negotiate`

### DIFF
--- a/expfmt/encode.go
+++ b/expfmt/encode.go
@@ -94,7 +94,21 @@ func Negotiate(h http.Header) Format {
 // temporary and will disappear once FmtOpenMetrics is fully supported and as
 // such may be negotiated by the normal Negotiate function.
 func NegotiateIncludingOpenMetrics(h http.Header) Format {
+	formats := NegotiateFormatsIncludingOpenMetrics(h)
+	return formats[0]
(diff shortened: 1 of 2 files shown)


bwplotka:
Duplicate of https://github.com/prometheus/common/pull/894?

Would you mind to help review https://github.com/prometheus/common/pull/894 instead?

bwplotka:
Ah I misunderstood, I think we solved the issue already https://github.com/prometheus/common/issues/928#issuecomment-5395660948 

Closing, we can always reopen, thanks!

PR #952, opened 2026-07-22
model: add Duration unit constants and conversion methods

### LINKED ISSUE #222: Extend model.Duration with `Miliseconds()` method as well as `model.{Milisecond,Second, ...}` vars.
Hi :wave: 

I would propose extending `model.Duration`. It's super painful to work with and prone to errors. Examples of lack of `Seconds` or `model.Milisecond`:

* `int64(db.opts.RetentionDuration)/int64(time.Millisecond)` to convert to miliseconds in Prometheus
* 15 * 24 * model.Duration(time.Hour)
*  `model.Duration(time.Duration(DefaultBlockDuration) * time.Millisecond),` :scream: 

I was thinking about adding `Miliseconds` method (we use miliseconds in Prometheus a lot!) as well as adding all `model.Milisecond` `model.Days` `model.Hours` `model.Seconds` `model.Microsecond` `model.Nanosecond` vars.

What do you think? Would we accept just contribution? @beorn7 @brian-brazil @SuperQ

locker95:
## Summary

Extend `model.Duration` with unit constants and conversion methods so callers no longer need awkward casts to/from `time.Duration`.

**Constants** (mirroring `time` package, plus Day/Week which `ParseDuration` already supports):

- `model.Nanosecond`, `model.Microsecond`, `model.Millisecond`
- `model.Second`, `model.Minute`, `model.Hour`
- `model.Day`, `model.Week`

**Methods** (same signatures as `time.Duration`):

- `Nanoseconds()`, `Microseconds()`, `Milliseconds()` → `int64`
- `Seconds()`, `Minutes()`, `Hours()` → `float64`

### Before / after

```go
// Before
int64(d) / int64(time.Millisecond)
15 * 24 * model.Duration(time.Hour)
model.Duration(time.Duration(n) * time.Millisecond)

// After
d.Milliseconds()
15 * model.Day
model.Duration(n) * model.Millisecond
```

Fixes #222

cc @roidelapluie @gotjosh

## Test plan

- [x] `go test ./model/ -run 'TestDuration' -count=1`
- [x] New tests cover constant values vs `time.*` equivalents, multiply usage, conversion methods (including negative durations)

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -164,6 +164,47 @@ func (t *Time) UnmarshalJSON(b []byte) error {
 // This type should not propagate beyond the scope of input/output processing.
 type Duration time.Duration
 
+// Common durations. Day and Week are included because ParseDuration supports
+// those units (unlike the standard time package).
+//
+// To count the number of units in a Duration, divide:
+//
+//	second := model.Second
+//	fmt.Print(int64(second/model.Millisecond)) // prints 1000
+//
+// To convert an integer number of units to a Duration, multiply:
+//
+//	seconds := 10
+//	fmt.Print(model.Duration(seconds)*model.Second) // prints 10s
+const (
+	Nanosecond  Duration = 1
+	Microsecond          = 1000 * Nanosecond
+	Millisecond          = 1000 * Microsecond
+	Second               = 1000 * Millisecond
+	Minute               = 60 * Second
+	Hour                 = 60 * Minute
+	Day                  = 24 * Hour
+	Week                 = 7 * Day
+)
+
+// Nanoseconds returns the duration as an integer nanosecond count.
+func (d Duration) Nanoseconds() int64 { return time.Duration(d).Nanoseconds() }
+
+// Microseconds returns the duration as an integer microsecond count.
+func (d Duration) Microseconds() int64 { return time.Duration(d).Microseconds() }
+
+// Milliseconds returns the duration as an integer millisecond count.
+func (d Duration) Milliseconds() int64 { return time.Duration(d).Milliseconds() }
+
+// Seconds returns the duration as a floating point number of seconds.
+func (d Duration) Seconds() float64 { return time.Duration(d).Seconds() }
+
+// Minutes returns the duration as a floating point number of minutes.
+func (d Duration) Minutes() float64 { return time.Duration(d).Minutes() }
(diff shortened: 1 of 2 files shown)


model/time.go
```diff
@@ -164,6 +164,47 @@ func (t *Time) UnmarshalJSON(b []byte) error {
...
+)
+
+// Nanoseconds returns the duration as an integer nanosecond count.
+func (d Duration) Nanoseconds() int64 { return time.Duration(d).Nanoseconds() }
+
+// Microseconds returns the duration as an integer microsecond count.
+func (d Duration) Microseconds() int64 { return time.Duration(d).Microseconds() }
+
+// Milliseconds returns the duration as an integer millisecond count.
+func (d Duration) Milliseconds() int64 { return time.Duration(d).Milliseconds() }
```
bwplotka, line 193: I vote week only Milliseconds as this is what we use in Prometheus. We don't expect to have this used as a generic time duration (time.Duration is good enough)

model/time.go
```diff
@@ -164,6 +164,47 @@ func (t *Time) UnmarshalJSON(b []byte) error {
...
+//	fmt.Print(model.Duration(seconds)*model.Second) // prints 10s
+const (
+	Nanosecond  Duration = 1
+	Microsecond          = 1000 * Nanosecond
+	Millisecond          = 1000 * Microsecond
+	Second               = 1000 * Millisecond
+	Minute               = 60 * Second
+	Hour                 = 60 * Minute
+	Day                  = 24 * Hour
+	Week                 = 7 * Day
```
bwplotka, line 187: Technically we also define a year, should we add it?

bwplotka (approved):
Thanks!

model/time.go
locker95: dropped the extra wrappers — left Milliseconds since that's the unit we actually use.

model/time.go
locker95: skipped it. ParseDuration already takes `y`, and a calendar year isn't a fixed duration like day/week.

bwplotka: I know, but we treat it as fixed in some calculations, anyway let's skip to not surface confusion

bwplotka (approved):

PR #954, opened 2026-07-28
Cache FileSecret reads

prymitive:
FileSecret reads files on every request, which means that if I configure a scrape job with:

```yaml
tls_config:
  ca_file: ca.pem
  cert_file: cert.pem
  key_file: key.pem
```

All three files will be read using FileSecret every time a scrape starts for that target. With 1k targets and 1m scrape interval this means 50 file reads per second (1000 targets * 3 files / 60s).
File reads require a blocking syscall so having to re-read files all the time isn't ideal on busy instances with a lot of targets. These files are normally small and reads are fast, but the overhead is still there.

Make FileSecret cache file reads, for 1m by default, configurable via cache_ttl YAML key.

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -40,10 +40,16 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
+
+	"github.com/prometheus/common/model"
 )
 
 const (
 	grantTypeJWTBearer = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+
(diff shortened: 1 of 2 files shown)


bwplotka (commented):
While more scalable, this will break scrapes for maximum of 1m on key rotation, just to be fair. Perhaps ok trade-off.

cc @roidelapluie

SuperQ:
I would recommend changing the default to be 0, no caching, to preserve the existing behavior.

These reads should pull from page cache and be essentially free.

roidelapluie:
I agree with Ben's comment. Is there any numbers to bake that this is noticeable? I do not think we had complaints in the past 10 years.

PR #962, opened 2026-08-18
Update http_config to support cfaccess

jaymzh:
Adds a new auth type cf-access which allows using this tool with
Cloudflare Access. Defines the new auth type, create a new RoundTripper
to handle the login / token fetch flow, and set it as the client
transport if enabled.

I originally planned to add this in prometheus/prometheus for
`promtool`, but (1) The implementation is cleaner doing it here and (2)
I want to do it for `amtool` as well, so implementing it in `common`
means not doing it twice.

If you were curious what it looks like to put directly in prometheus/prometheus,
there's a commit here for comparison: https://github.com/jaymzh/prometheus/commit/6ea12c1be6cf06b6af9ec19e6d04e69fe85d1edf

Signed-off-by: Phil Dibowitz <phil@ipom.com>


roidelapluie:
Thank you. This seems to drag a lot of dependencies to this project, which is broadly used.

Do you think it could fit in its own repo/module? We have a precedent: https://github.com/prometheus/sigv4/ . Happy to have that under `prometheus/` if you think it's useful

jaymzh:
Thanks for the reply!

Yes, since this is implemented as a RoundTripper, extracting it into a separate module should be straightforward.

I originally put it in common to avoid requiring each consumer to interpret authorization.type: cf-access and install an additional RoundTripper. However, a standalone module could keep that integration small while preventing common from inheriting `cloudflared`s dependency tree.

I envision the module providing a helper such as:

```golang
cfg, cfAccessEnabled, err := cfaccess.PrepareHTTPClientConfig(cfg)
if err != nil {
    return err
}
```

This would detect and validate the cf-access configuration and return a copy with the authorization entry removed, preventing common from treating cf-access as a literal HTTP authorization scheme.

The consumer would then construct its normal transport and conditionally wrap it:

```golang
rt, err := commonconfig.NewRoundTripperFromConfig(cfg, "promtool")
if err != nil {
    return err
}
if cfAccessEnabled {
    rt = cfaccess.NewRoundTripper(rt)
}
```

`amtool` would do the equivalent around the transport returned by `NewClientFromConfig`.

Before doing the extraction, I'd like to confirm that this is the intended architecture and that follow-up integrations for `promtool` and `amtool` would be welcome. The standalone module would not provide user-facing functionality until consumers adopt it.

 If so, wanna create that repo and I'll make modify this a bit and PR over there?

jaymzh:
@roidelapluie - would the above work?

roidelapluie:
How should I name the repo ?

(I need to check with new governance how I am allowed to create a new repo)

jaymzh:
@roidelapluie It's your repo, your call, but perhaps one of:

* prom-cfaccess  (I like this because it is specific to prom's http libs)
* httpconfig-cfaccess (also clear, but I prefer the above)
* cfaccess (matches sigv4, though I still prefer prom-access)

roidelapluie:
Repo is created and can be PR'd:

https://github.com/prometheus/cfaccess

After a week and a few PR's, I will look to promote you as maintainer of that repo, if that is fine.

jaymzh:
Yeah, that sounds great, thanks.

PR created: https://github.com/prometheus/cfaccess/pull/3

PR #964, opened 2026-08-21
feat: implement histogram and gauge histogram support for OpenMetrics 2.0

dashpole:
Part 2 of https://github.com/prometheus/common/issues/893

Implement serialization and strict validation for classic, native, dual, and gauge histograms in OpenMetrics 2.0 format according to the specification and ABNF grammar.

```
goos: linux
goarch: amd64
pkg: github.com/prometheus/common/expfmt
cpu: AMD EPYC 7B12
                                             │    sec/op     │     B/op      │   allocs/op   │
ConvertMetricFamily/TEXT/COUNTER-24              502.8n ± 2%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/OM1.0/COUNTER-24             1.211µ ± 4%      33.00 ± 0%     2.000 ± 0%
ConvertMetricFamily/OM2.0/COUNTER-24             1.100µ ± 5%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/TEXT/GAUGE-24                487.5n ± 5%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/OM1.0/GAUGE-24               735.8n ± 5%      1.000 ± 0%     1.000 ± 0%
ConvertMetricFamily/OM2.0/GAUGE-24               653.2n ± 3%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/TEXT/UNTYPED-24              428.9n ± 4%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/OM1.0/UNTYPED-24             599.5n ± 8%      1.000 ± 0%     1.000 ± 0%
ConvertMetricFamily/OM2.0/UNTYPED-24             573.3n ± 7%      0.000 ± 0%     0.000 ± 0%
ConvertMetricFamily/TEXT/SUMMARY-24              2.125µ ± 4%      56.00 ± 0%     2.000 ± 0%
ConvertMetricFamily/OM1.0/SUMMARY-24             2.745µ ± 7%      57.00 ± 0%     3.000 ± 0%
ConvertMetricFamily/TEXT/HISTOGRAM-24            4.298µ ± 7%      288.0 ± 0%     6.000 ± 0%
ConvertMetricFamily/OM1.0/HISTOGRAM-24           5.265µ ± 5%      289.0 ± 0%     7.000 ± 0%
[...]

PR #965, opened 2026-08-21
fix: support exemplars with empty label sets in OpenMetrics 2.0

dashpole:
Part of https://github.com/prometheus/common/issues/893

Split out of https://github.com/prometheus/common/pull/964

OpenMetrics 2.0 specification represents exemplars with an empty label set ({}). Previously, exemplars without labels were dropped.  From the [spec](https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#exemplars-1):

> "Exemplars without Labels MUST represent an empty LabelSet as {}."


bwplotka (approved):
Great, thanks!

PR #966, opened 2026-08-24
expfmt: fix wildcard content negotiation regression for Negotiate and NegotiateIncludingOpenMetrics

dashpole:
Fixes a regression introduced by https://github.com/prometheus/common/pull/894.

Clients sending Accept: */* or browser Accept headers containing wildcards should receive standard Prometheus text format (FmtText) when using Negotiate or NegotiateIncludingOpenMetrics.

Previously, matchFormat short-circuited on */*, immediately returning the first format in the accepted list (Protobuf for Negotiate, and OpenMetrics 1.0 for NegotiateIncludingOpenMetrics).

Remove the */* short-circuit from matchFormat so wildcard Accept headers fall back to the text-preferring fallback logic in NegotiateAccept.

### DIFF
--- a/expfmt/encode_test.go
+++ b/expfmt/encode_test.go
@@ -78,6 +78,21 @@ func TestNegotiate(t *testing.T) {
 			acceptHeaderValue: "text/plain;version=0.0.4; escaping=values;",
 			expectedFmt:       "text/plain; version=0.0.4; charset=utf-8; escaping=values",
 		},
+		{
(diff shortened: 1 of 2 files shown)


expfmt/encode_test.go
bwplotka, line 29: observation: Probably a duplicated test, we could DRY here

bwplotka (approved):
Thanks!

PR #970, opened 2026-08-24
fix: drop invalid OpenMetrics 2.0 exemplars instead of failing exposition

dashpole:
Small follow-up to https://github.com/prometheus/common/pull/894.

Per the OpenMetrics 2.0 specification failure modes, failures specific to exemplars should not cause the entire exposition to fail. Invalid exemplars (e.g. invalid timestamps, malformed labels) are now dropped so that the rest of the metric exposition succeeds.

> "This specification advocates for transactional processing: any encoding, decoding, or validation errors must reject the whole MetricSet ingestion. A failed scrape is better than an inaccurate scrape or a partial metric view that breaks transactionality (e.g., scraping a portion of a StateSet MetricGroup, or scraping only one Counter out of two that are aggregated in a single alert expression).

> There's one exception to this rule: failures specific to exemplars should not cause the entire exposition to fail. If an exemplar is malformed or invalid, it should be dropped or ignored, allowing the valid metric data to be ingested."

This also consolidates exemplar timestamp validation in `validateExemplar20`.

### DIFF
--- a/expfmt/openmetrics_2_0_create.go
+++ b/expfmt/openmetrics_2_0_create.go
@@ -263,7 +263,7 @@ func writeOpenMetrics20Sample(w enhancedWriter, name string, metric *dto.Metric,
 		}
 	}
 
-	if exemplar != nil && exemplar.Timestamp != nil {
+	if exemplar != nil {
 		n, err = writeExemplar20(w, exemplar)
 		written += n
 		if err != nil {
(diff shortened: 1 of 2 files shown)


expfmt/openmetrics_2_0_create.go
```diff
@@ -280,13 +280,13 @@ func writeOpenMetrics20Sample(w enhancedWriter, name string, metric *dto.Metric,
...
-// In OpenMetrics 2.0, exemplars without a timestamp are dropped.
+// In OpenMetrics 2.0, invalid exemplars or exemplars without a timestamp are dropped.
 func writeExemplar20(w enhancedWriter, e *dto.Exemplar) (int, error) {
-	if e == nil || e.Timestamp == nil {
+	if e == nil {
 		return 0, nil
 	}
 	if err := validateExemplar20(e); err != nil {
-		return 0, err
+		return 0, nil
```
bwplotka, line 290: Deserves a comment, easy to miss

dashpole: addressed in 223e26b

bwplotka (approved):
Thanks!

PR #972, opened 2026-08-27
chore: update golangci-lint to v2.13 for go1.27 compatibility

thaJeztah:
- fix linting
- ci: remove obsolete GODEBUG settings

  Remove the x509sha1 and tls10default compatibility settings from the
  test environment. Neither is supported by the minimum tested Go version,
  and x509sha1 causes binaries built with recent Go versions to fail at
  startup.

- chore: update golangci-lint to v2.13 for go1.27 compatibility

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -61,7 +61,7 @@ PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_
 SKIP_GOLANGCI_LINT :=
 GOLANGCI_LINT :=
 GOLANGCI_LINT_OPTS ?=
-GOLANGCI_LINT_VERSION ?= v2.12.2
+GOLANGCI_LINT_VERSION ?= v2.13.1
 GOLANGCI_FMT_OPTS ?=
 # golangci-lint only supports linux, darwin and windows platforms on i386/amd64/arm64.
 # windows isn't included here because of the path separator being different.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,6 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    env:
-      # Override Go 1.18 security deprecations.
-      GODEBUG: "x509sha1=1,tls10default=1"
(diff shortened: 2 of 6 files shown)


thaJeztah:
Ah; it doesn't like a GODEBUG, because it's deprecated in go1.24, and go1.27 now rejects obsolete ones 🤔 

```
golangci/golangci-lint info checking GitHub for tag 'v2.13.1'
golangci/golangci-lint info found version: 2.13.1 for v2.13.1/linux/amd64
golangci/golangci-lint info installed /home/runner/go/bin/golangci-lint
>> running golangci-lint
/home/runner/go/bin/golangci-lint run  ./...
fatal error: removed GODEBUG "x509sha1" set to old value "1" in environment (https://go.dev/doc/godebug#go-124)

goroutine 1 [running, locked to thread]:
runtime.fatal({0xfea13f16230, 0x62})
	runtime/panic.go:1267 +0x74
runtime.main()
	runtime/proc.go:224 +0x2aa
runtime.goexit({})
	runtime/asm_amd64.s:1264 +0x1
make: *** [Makefile.common:205: common-lint] Error 2
```

Makefile.common
```diff
@@ -61,7 +61,7 @@ PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_
 SKIP_GOLANGCI_LINT :=
 GOLANGCI_LINT :=
 GOLANGCI_LINT_OPTS ?=
-GOLANGCI_LINT_VERSION ?= v2.12.2
+GOLANGCI_LINT_VERSION ?= v2.13.1
```
SuperQ, line 64: This file is managed and synced from upstream Prometheus.

SuperQ (approved):
Thanks. It's probably best to just apply the pending fixes and we'll pick up the `Makefile.common` update soon when we apply the latest version upstream.

Makefile.common
SuperQ:
```suggestion
GOLANGCI_LINT_VERSION ?= v2.12.2
```

PR #982, opened 2026-09-01
fix(config): deep copy the header in cloneRequest

### LINKED ISSUE #981: cloneRequest does not copy the header, so round trippers mutate the caller's request
`cloneRequest` in `config/http_config.go` intends a deep copy of the header but does not make one:

```go
r2 := new(http.Request)
*r2 = *r
// Deep copy of the Header.
maps.Copy(r.Header, r2.Header)
```

After `*r2 = *r`, `r2.Header` is the same map as `r.Header`, so `maps.Copy` copies it onto itself. The clone shares the caller's header map.

Every round tripper that touches headers is affected, since they all clone first and then modify. `headersRoundTripper` uses `Header.Add`, so a reused request gains another copy of every configured header on each round trip. It grows without bound until the server rejects it — a downstream report in Alloy ([grafana/alloy#7016](https://github.com/grafana/alloy/issues/7016)) shows ~1000 identical `X-Api-Key` lines in one request and scrapes failing with 431.

There is a second effect: `Header.Set` on the clone of a request built without a header panics with `assignment to entry in nil map`, because the shared nil map is never replaced.

Introduced in 56870db ("Modernize Go"), which replaced the manual copy loop:

```go
-	r2.Header = make(http.Header)
-	for k, s := range r.Header {
-		r2.Header[k] = s
-	}
+	maps.Copy(r.Header, r2.Header)
```

The arguments are also the wrong way round, but swapping them is not enough — the clone needs its own map.

Fix in #982.

SaiPisey2:
Fixes #981.

`cloneRequest` intends a deep copy of the header but does not make one:

```go
r2 := new(http.Request)
*r2 = *r
// Deep copy of the Header.
maps.Copy(r.Header, r2.Header)
```

After `*r2 = *r`, `r2.Header` is already the same map as `r.Header`, so `maps.Copy` copies it onto itself and the clone shares the caller's header map.

Every round tripper here clones and then modifies, so each one mutates the request it was given. `headersRoundTripper` uses `Header.Add`, so a reused request gains another copy of every configured header on each round trip, without bound. [grafana/alloy#7016](https://github.com/grafana/alloy/issues/7016) reports ~1000 identical `X-Api-Key` lines in a single request, with the target returning 431 and profiling stopping.

Introduced in 56870db, which replaced the manual copy loop with `maps.Copy`. The arguments are also the wrong way round, but swapping them is not enough — the clone needs its own map.

`Header.Clone()` returns nil for a nil header, while the previous code always produced a usable map, so the nil case is kept: `userAgentRoundTripper` and others call `Header.Set` straight after cloning, which panics with `assignment to entry in nil map` on a request built without one. That path is broken on main today as well.

## Tests

Two tests, both failing on main and passing here:

- `TestCloneRequest` — the clone must not share the caller's map, and a request without a header must still be usable afterwards.
- `TestHeadersRoundTripperReusedRequest` — reusing a request must not accumulate headers, and must not modify the caller's request.

`go build ./...`, `go test ./...` and `go test -race ./config/` are green on darwin/arm64, go1.26.5.

### DIFF
--- a/config/headers_test.go
+++ b/config/headers_test.go
@@ -69,6 +69,35 @@ func TestHeadersRoundTripperSameHost(t *testing.T) {
 	}
 }
 
+func TestHeadersRoundTripperReusedRequest(t *testing.T) {
+	// The round tripper must not mutate the request it is given: reusing the
+	// same request must not accumulate another copy of every header.
+	var received []string
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		received = r.Header.Values("X-Custom-Header")
+		fmt.Fprint(w, "ok")
+	}))
+	t.Cleanup(server.Close)
+
+	headers := &Headers{
+		Headers: map[string]Header{
+			"X-Custom-Header": {Values: []string{"testvalue"}},
+		},
+	}
+	rt := NewHeadersRoundTripper(headers, http.DefaultTransport)
+
+	req, err := http.NewRequest(http.MethodGet, server.URL, nil)
+	require.NoError(t, err)
+
+	for i := range 3 {
+		resp, err := rt.RoundTrip(req)
+		require.NoError(t, err)
+		resp.Body.Close()
+		require.Equalf(t, []string{"testvalue"}, received, "header duplicated on request %d", i+1)
+		require.Empty(t, req.Header.Values("X-Custom-Header"), "the caller's request was modified")
+	}
+}
+
 func TestHeadersRoundTripperCrossHostRedirect(t *testing.T) {
 	// Cookie must be set on the initial request but stripped on cross-host redirects.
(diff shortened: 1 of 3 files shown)


SaiPisey2:
Bump — green and mergeable.

This is a regression on main from 56870db: `maps.Copy(r.Header, r2.Header)` runs after `*r2 = *r`, so it copies the map onto itself and the clone keeps the caller's header. grafana/alloy#7016 sees ~1000 duplicated `X-Api-Key` lines and a 431 from it. @roidelapluie @gotjosh could one of you take a look?

SaiPisey2:
Still green and mergeable. @roidelapluie @gotjosh this one is a regression on main from 56870db, where `maps.Copy` copies the header map onto itself, so every round tripper mutates the caller request. Happy to rebase if it has gone stale.

PR #983, opened 2026-09-02
feat: implement summary support for OpenMetrics 2.0

dashpole:
Part of https://github.com/prometheus/common/issues/893

Implement serialization and strict validation for summaries in OpenMetrics 2.0 format according to the [specification](https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/) and ABNF grammar.

### Benchmark Results

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkConvertMetricFamily/TEXT/SUMMARY-24         	  345410	      3426 ns/op	      56 B/op	       2 allocs/op
BenchmarkConvertMetricFamily/OM1.0/SUMMARY-24        	  273915	      4486 ns/op	      57 B/op	       3 allocs/op
BenchmarkConvertMetricFamily/OM2.0/SUMMARY-24        	  501536	      2026 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/expfmt/openmetrics_2_0_create.go
+++ b/expfmt/openmetrics_2_0_create.go
@@ -31,7 +31,7 @@ import (
 // It returns the number of bytes written and any error encountered.
 //
 // NOTE: This method targets OpenMetrics 2.0.0 (currently aligned with 2.0-rc.0) which is experimental and
-// encode-only (currently supporting counter, gauge, and untyped metric types).
+// encode-only (currently supporting counter, gauge, summary, untyped, histogram, and gaugehistogram metric types).
 // Breaking changes might happen in the future. This implementation is still a
 // work-in-progress, and does not yet support all features of the format.
 // EncoderOptions are accepted for signature compatibility with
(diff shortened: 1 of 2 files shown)


expfmt/openmetrics_2_0_create.go
```diff
@@ -340,13 +340,159 @@ func writeOpenMetrics20Timestamp(w enhancedWriter, f float64) (int, error) {
...
+		}
+	}
+
+	sum := s.GetSampleSum()
+	if math.IsNaN(sum) {
+		return 0, fmt.Errorf("summary sum cannot be NaN in metric %s", name)
+	}
+	if sum < 0 {
+		return 0, fmt.Errorf("summary sum cannot be negative (%g) in metric %s", sum, name)
+	}
```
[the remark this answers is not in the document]
dashpole: This validation directly follows the OpenMetrics 2.0 specification. 

According to Section [4.1.7 Summary](https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#summary):

> "A Summary Sample MUST contain a Count, Sum and a set of quantiles.  
> **Semantically, Count and Sum values are counters so MUST NOT be NaN or negative. Count MUST be an integer.**  
> ...  
> **Quantile values MUST NOT be negative.**"

Unlike histograms—where Section 4.1.6 explicitly permits negative observation values affecting the sum ("*The Sum is only a counter semantically as long as there are no negative event values measured by the Histogram*")—the specification for summaries explicitly treats both Count and Sum as counters that must not be negative or NaN, and forbids negative quantile values. 

(Note: Line 380 validates that histogram *count* cannot be negative, not sum).

expfmt/openmetrics_2_0_create.go
```diff
@@ -340,13 +340,159 @@ func writeOpenMetrics20Timestamp(w enhancedWriter, f float64) (int, error) {
...
+		if lp.GetName() == "quantile" {
+			return 0, fmt.Errorf("metric %s is a summary but label set contains %q label", name, "quantile")
+		}
+	}
+
+	sum := s.GetSampleSum()
+	if math.IsNaN(sum) {
+		return 0, fmt.Errorf("summary sum cannot be NaN in metric %s", name)
+	}
+	if sum < 0 {
```
dashpole: The spec does say "Quantile values MUST NOT be negative" and "Count and Sum ... MUST NOT be NaN or negative", so these checks are correct — but note that `summary.Observe()` in client_golang accepts negative values with no guard (unlike `counter.Add`, which panics), and this repo's own OM 1.0 fixture (`openmetrics_create_test.go:294`, `SampleSum: -3.4567`) encodes fine as OM 1.0 and now fails here, taking the rest of the family with it. Worth documenting on `MetricFamilyToOpenMetrics20` that OM 2.0 rejects families the other two encoders accept.

expfmt/openmetrics_2_0_create.go
```diff
@@ -340,13 +345,159 @@ func writeOpenMetrics20Timestamp(w enhancedWriter, f float64) (int, error) {
...
+	var prevQuantile float64
+	for i, q := range s.Quantile {
+		if q == nil {
+			return 0, fmt.Errorf("expected non-nil quantile in metric %s", name)
+		}
+		qv := q.GetQuantile()
+		if math.IsNaN(qv) {
+			return 0, fmt.Errorf("summary quantile cannot be NaN in metric %s", name)
+		}
+		if math.IsInf(qv, 0) || qv < 0 || qv > 1 {
```
bwplotka: Quick nit: `math.IsInf(qv, 0)` is redundant here since `qv < 0 || qv > 1` already catches both `+Inf` and `-Inf`.
```suggestion
		if qv < 0 || qv > 1 {
```

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -923,6 +923,244 @@ empty_histogram {count:0,sum:0,bucket:[+Inf:0]}
...
+			in: &dto.MetricFamily{
+				Name: proto.String("foo"),
+				Type: dto.MetricType_SUMMARY.Enum(),
+				Metric: []*dto.Metric{
+					{
+						Summary: &dto.Summary{
+							SampleCount: proto.Uint64(0),
+							SampleSum:   proto.Float64(0),
+							Quantile: []*dto.Quantile{
+								{Quantile: proto.Float64(0.95), Value: proto.Float64(123.7)},
```
bwplotka, line 988: Could we include quantile `0` (with value `0`) and quantile `1` in this test case? That would cover the inclusive `[0, 1]` quantile bounds, zero quantile values, and verify the `i > 0` check when `prevQuantile` starts at `0`.

expfmt/openmetrics_2_0_create_test.go
```diff
@@ -1909,3 +2333,28 @@ func TestCreateOpenMetrics20_HistogramError_NoPartialBytes(t *testing.T) {
...
+					SampleCount: proto.Uint64(1),
+					SampleSum:   proto.Float64(-1.0), // invalid sum
+				},
+			},
+		},
+	}
+
+	var buf bytes.Buffer
+	w := enhancedWriter(&buf)
+	_, err := writeCompositeSummary(w, in.GetName(), in.Metric[0])
```
bwplotka: nit: Consider capturing `n, err := writeCompositeSummary(...)` and asserting `n == 0` as well, and testing a failure in the last validation step (like an invalid `CreatedTimestamp`) to verify that all validation paths return zero written bytes.

bwplotka (approved):
Looks great, tiny nits only.

expfmt/openmetrics_2_0_create_test.go
dashpole: addressed in 29b91d7

expfmt/openmetrics_2_0_create_test.go
dashpole: addressed in 2a75737

PR #985, opened 2026-09-03
Add http debugging roundtripper

jaymzh:
When developing things (like https://github.com/prometheus/common/pull/962/), it's useful to be able
to debug the http roundtrips. This adds `debugRoundTripper` that other prom 
tools can use (behind a `--http.debug` flag) which automatically
logs the HTTP requests while redacting credentials.

Implemented downstream usage for promtool and amtool: https://github.com/prometheus/prometheus/pull/19577 and https://github.com/prometheus/alertmanager/pull/5541

Signed-off-by: Phil Dibowitz <phil@ipom.com>

### DIFF
--- a/config/http_debug.go
+++ b/config/http_debug.go
@@ -1,0 +1,113 @@
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


bboreham:
Why would this live in a Prometheus package?  Seems quite general.

What similar existing tools are out there?

jaymzh:
There's not a decent equivalent I can find. There's replacements for the entire HTTP client that have similar things, but it seems extreme to rip out the entire http client.

The closest thing I've found is net/http/httputil's httputil.DumpRequestOut and DumpResponse. However, it has no redaction or body-size limit. Further, DumpResponse(..., true) buffers the entire body, so we would still need a wrapper for safe previews.

Adding a simple round-tripper to do this didn't seem like a huge problem, and I implemented not for fun, but because I needed something like it to implement #962

jaymzh:
(that said, if this isn't desired, I have no hard need for it now that #962 is written... the cfaccess support was the goal, and that's the PR I'm most interested in getting merged. I just thought it'd be a good-citizen thing to do to make this generic and upstream it as well)

jaymzh:
(which is my subtle attempt to get attention on #962  ;) )

PR #986, opened 2026-09-07
chore: Replace sort package with slices and cmp packages

dongjiang1989:
Replace all usage of Go's `sort` package with the modern `slices` and `cmp` packages (Go 1.21+) for more concise and readable sorting code.

## Changes

- Remove `sort.Interface` methods (`Len`/`Less`/`Swap`) from `LabelNames`, `LabelValues`, `LabelPairs`, `Fingerprints`, `Alerts`, `Samples`, `Vector`, and `Matrix` types.
- Add `Compare` methods on `*Sample`, `*SampleStream`, `*Alert`, and `*LabelPair` for use with `slices.SortFunc`, eliminating duplicated comparator code at call sites.
- Replace `sort.Sort` calls with `slices.SortFunc` (using the new `Compare` methods or `cmp.Compare` for simple ordered types), and `sort.Strings` with `slices.Sort`.
- Add a `depguard` rule in `.golangci.yml` to forbid the `sort` package and prevent future regressions.
- Fix pre-existing `gofumpt` formatting issues in `config` and `route` packages (uncovered by `make lint`).

## Note

The original `Alerts.Less` was not a proper strict weak ordering (it used `StartsBefore || EndsBefore || FPLess`). The new `(*Alert).Compare` implements a correct hierarchical comparison (StartsAt → EndsAt → Fingerprint). `TestSortAlerts` expected values were updated accordingly.

## Verification

- `go build ./...` ✅
- `go vet ./...` ✅
- `make lint` ✅
- `go test ./model/... ./expfmt/...` ✅

### DIFF
--- a/config/http_config.go
+++ b/config/http_config.go
@@ -71,10 +71,10 @@ type closeIdler interface {
 type TLSVersion uint16
 
 var TLSVersions = map[string]TLSVersion{
-	"TLS13": (TLSVersion)(tls.VersionTLS13),
-	"TLS12": (TLSVersion)(tls.VersionTLS12),
-	"TLS11": (TLSVersion)(tls.VersionTLS11),
-	"TLS10": (TLSVersion)(tls.VersionTLS10),
+	"TLS13": TLSVersion(tls.VersionTLS13),
+	"TLS12": TLSVersion(tls.VersionTLS12),
+	"TLS11": TLSVersion(tls.VersionTLS11),
+	"TLS10": TLSVersion(tls.VersionTLS10),
 }
 
 func (tv *TLSVersion) UnmarshalYAML(unmarshal func(any) error) error {
@@ -639,11 +639,13 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
 		dialContext = conntrack.NewDialContextFunc(
 			conntrack.DialWithDialContextFunc((func(context.Context, string, string) (net.Conn, error))(opts.dialContextFunc)),
 			conntrack.DialWithTracing(),
-			conntrack.DialWithName(name))
+			conntrack.DialWithName(name),
+		)
 	} else {
 		dialContext = conntrack.NewDialContextFunc(
 			conntrack.DialWithTracing(),
-			conntrack.DialWithName(name))
+			conntrack.DialWithName(name),
+		)
 	}
 
 	newRT := func(tlsConfig *tls.Config) (http.RoundTripper, error) {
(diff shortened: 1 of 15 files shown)


bwplotka:
NOTE: This is a breaking change for downstream - we should do it ideally only when we have to

config/http_config.go
```diff
@@ -71,10 +71,10 @@ type closeIdler interface {
 type TLSVersion uint16
 
 var TLSVersions = map[string]TLSVersion{
-	"TLS13": (TLSVersion)(tls.VersionTLS13),
-	"TLS12": (TLSVersion)(tls.VersionTLS12),
-	"TLS11": (TLSVersion)(tls.VersionTLS11),
-	"TLS10": (TLSVersion)(tls.VersionTLS10),
+	"TLS13": TLSVersion(tls.VersionTLS13),
```
bboreham, line 74: This change is not covered by the PR title.

config/http_config.go
```diff
@@ -639,11 +639,13 @@ func NewRoundTripperFromConfigWithContext(ctx context.Context, cfg HTTPClientCon
 		dialContext = conntrack.NewDialContextFunc(
 			conntrack.DialWithDialContextFunc((func(context.Context, string, string) (net.Conn, error))(opts.dialContextFunc)),
 			conntrack.DialWithTracing(),
-			conntrack.DialWithName(name))
+			conntrack.DialWithName(name),
+		)
```
bboreham, line 643: Spurious change

bboreham (commented):
Agree with Bartek - it's quite a breaking change to make all these types no longer implement `sort.Interface`.  Suggest to leave them in, maybe with a comment that they are for backwards-compatibility only. You could have `Less` call the new `Compare`.

Can you explain a bit more about the change to `Alert` ordering?  This will be another breaking change; we'd need a strong justification.

> Fix pre-existing gofumpt formatting issues in config and route packages (uncovered by make lint).

Do this in a separate PR.  It's extra work to review changes for multiple different reasons at the same time.

Also, it might be neater to use [`cmp.Or`](https://pkg.go.dev/cmp#Or) in the `Compare` methods.

dongjiang1989:
Thanks for the review @bwplotka @bboreham.

Understood. I'll restore Len/Less/Swap methods as a compatibility layer, delegating to the new Compare().

I will make these changes in the next push:
1. Fix doc comments and remove the wrong cmp.Comparator reference.
2. Restore sort.Interface methods (Len/Less/Swap) as a compatibility layer, delegating to Compare().
3. Rewrite multi-level comparisons using cmp.Or.
4. Strip unrelated gofumpt formatting changes to a separate PR, clean up spurious diffs.
5. The old Alert sort had invalid ordering. The new ordering fixes this bug, which is a behavioral change. I will add a changelog note for this.

PR #987, opened 2026-09-12
chore: Update gofumpt config and apply new formatting

dongjiang1989:
Migrate gofumpt configuration to the new schema: replace the deprecated `extra-rules: true` with explicit `extra` sub-options (`module-path`, `group-params`, `clothe-returns`, `balance-calls`) and reformat the affected call sites and type conversions accordingly.

ref: https://github.com/prometheus/common/pull/986#pullrequestreview-5144682931

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -177,6 +177,7 @@ formatters:
         - default
         - prefix(github.com/prometheus/common)
     gofumpt:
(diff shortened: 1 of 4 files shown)


SuperQ (approved):
Thanks.

PR #991, opened 2026-09-20
optimize sample unmarshaling for json/v2

liggitt:
Part of https://github.com/prometheus/client_golang/issues/2105

* Adds UnmarshalJSONFrom optimized streaming unmarshal implementations
* Implementations are in build-tagged Go 1.27 files, previous version of Go still use less efficient unmarshalers
* Adds a copy of the sample unmarshaling benchmark from client_golang
* Adds unit tests to check parity between UnmarshalJSON and UnmarshalJSONFrom implementations

Benchmarks at https://gist.github.com/liggitt/ad7346af10affd28a83881f42f851c07

Go 1.27 performance is significantly improved:
```
benchstat before-go1.27 after-go1.27 
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                                                            │  1-go1.27   │              3-go1.27              │
                                                            │   sec/op    │   sec/op     vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14           56.57µ ± 2%   15.17µ ± 3%  -73.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14       538.6µ ± 1%   252.9µ ± 1%  -53.04% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14          499.9µ ± 0%   106.1µ ± 4%  -78.78% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14      5.379m ± 1%   2.491m ± 1%  -53.68% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14         5.015m ± 1%   1.029m ± 1%  -79.48% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14     54.06m ± 1%   25.23m ± 1%  -53.32% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14          571.9µ ± 1%   148.8µ ± 1%  -73.98% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14      5.478m ± 1%   2.547m ± 5%  -53.50% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14         5.132m ± 1%   1.127m ± 2%  -78.05% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14     54.76m ± 1%   26.00m ± 3%  -52.53% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14        50.94m ± 1%   11.03m ± 2%  -78.34% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14    551.0m ± 1%   267.8m ± 3%  -51.39% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14         5.744m ± 1%   1.604m ± 4%  -72.07% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14     55.89m ± 1%   26.66m ± 4%  -52.30% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14        51.48m ± 1%   11.36m ± 2%  -77.94% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14    557.5m ± 1%   266.1m ± 3%  -52.27% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14       511.5m ± 1%   112.4m ± 3%  -78.02% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14    5.584 ± 1%    2.638 ± 1%  -52.75% (p=0.002 n=6)
geomean                                                       17.03m        5.638m       -66.89%

                                                            │   1-go1.27    │              3-go1.27               │
                                                            │     B/op      │     B/op      vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14           14.816Ki ± 0%   9.490Ki ± 0%  -35.95% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14       122.76Ki ± 0%   44.67Ki ± 0%  -63.61% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14           92.10Ki ± 0%   44.52Ki ± 0%  -51.67% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14      1175.4Ki ± 0%   400.1Ki ± 0%  -65.96% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14          967.2Ki ± 0%   497.2Ki ± 0%  -48.59% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14     11.335Mi ± 0%   3.775Mi ± 0%  -66.70% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14          147.68Ki ± 0%   94.40Ki ± 0%  -36.08% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14      1227.3Ki ± 0%   446.2Ki ± 0%  -63.64% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14          920.1Ki ± 0%   444.5Ki ± 0%  -51.69% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14     11.471Mi ± 0%   3.905Mi ± 0%  -65.95% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14         9.438Mi ± 0%   4.853Mi ± 0%  -48.58% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14    113.28Mi ± 0%   37.74Mi ± 0%  -66.69% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14         1471.7Ki ± 0%   939.6Ki ± 0%  -36.16% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14     11.973Mi ± 0%   4.352Mi ± 0%  -63.65% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14         8.975Mi ± 0%   4.335Mi ± 0%  -51.69% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14    114.63Mi ± 0%   39.04Mi ± 0%  -65.95% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14        94.37Mi ± 0%   48.53Mi ± 0%  -48.58% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14   1132.7Mi ± 0%   377.3Mi ± 0%  -66.69% (p=0.002 n=6)
geomean                                                        3.521Mi        1.523Mi       -56.74%

                                                            │   1-go1.27    │              3-go1.27              │
                                                            │   allocs/op   │  allocs/op   vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14              426.0 ± 0%    116.0 ± 0%  -72.77% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14         5.226k ± 0%   1.216k ± 0%  -76.73% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14            3156.0 ± 0%    146.0 ± 0%  -95.37% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14        51.16k ± 0%   11.15k ± 0%  -78.21% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14          30197.0 ± 0%    186.0 ± 0%  -99.38% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14       510.2k ± 0%   110.2k ± 0%  -78.41% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14            4.209k ± 0%   1.109k ± 0%  -73.65% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14        52.21k ± 0%   12.11k ± 0%  -76.81% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14          31.510k ± 0%   1.409k ± 0%  -95.53% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14       511.5k ± 0%   111.4k ± 0%  -78.22% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14        301.911k ± 0%   1.809k ± 0%  -99.40% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14      5.102M ± 0%   1.102M ± 0%  -78.41% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14           42.03k ± 0%   11.03k ± 0%  -73.76% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14       522.1k ± 0%   121.0k ± 0%  -76.82% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14         315.03k ± 0%   14.03k ± 0%  -95.55% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14      5.115M ± 0%   1.114M ± 0%  -78.22% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14       3019.04k ± 0%   18.03k ± 0%  -99.40% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14     51.02M ± 0%   11.02M ± 0%  -78.41% (p=0.002 n=6)
geomean                                                         132.8k        12.77k       -90.38%
```

Go 1.26 CPU performance is slightly improved by dropping an unnecessary custom UnmarshalJSON method:
```
benchstat before-go1.26 after-go1.26 
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                                                            │   1-go1.26   │              3-go1.26              │
                                                            │    sec/op    │   sec/op     vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14           59.13µ ±  3%   44.60µ ± 3%  -24.58% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14       760.4µ ±  1%   586.7µ ± 0%  -22.84% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14          476.0µ ±  4%   382.1µ ± 1%  -19.73% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14      7.449m ±  1%   5.880m ± 1%  -21.07% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14         4.698m ±  1%   3.799m ± 2%  -19.13% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14     76.79m ±  3%   59.98m ± 1%  -21.89% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14          586.1µ ±  2%   451.7µ ± 1%  -22.94% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14      7.550m ±  1%   5.947m ± 0%  -21.22% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14         4.714m ±  0%   3.855m ± 2%  -18.22% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14     75.83m ±  1%   59.81m ± 0%  -21.14% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14        47.63m ±  3%   38.68m ± 3%  -18.80% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14    839.1m ± 43%   617.9m ± 0%  -26.36% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14         6.479m ± 11%   4.566m ± 1%  -29.53% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14     81.97m ±  2%   60.11m ± 1%  -26.67% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14        51.16m ±  2%   39.35m ± 1%  -23.10% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14    804.5m ±  1%   619.6m ± 1%  -22.98% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14       478.6m ±  2%   394.4m ± 1%  -17.59% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14    7.861 ±  2%    6.218 ± 1%  -20.89% (p=0.002 n=6)
geomean                                                       20.15m         15.67m       -22.21%

                                                            │   1-go1.26   │              3-go1.26               │
                                                            │     B/op     │     B/op      vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14           31.16Ki ± 0%   28.02Ki ± 0%  -10.08% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14       220.9Ki ± 0%   217.1Ki ± 0%   -1.71% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14          227.9Ki ± 0%   224.7Ki ± 0%   -1.38% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14      2.073Mi ± 0%   2.070Mi ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14         2.244Mi ± 0%   2.241Mi ± 0%   -0.14% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14     20.56Mi ± 0%   20.56Mi ± 0%   -0.02% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14          309.4Ki ± 0%   277.4Ki ± 0%  -10.33% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14      2.155Mi ± 0%   2.117Mi ± 0%   -1.73% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14         2.223Mi ± 0%   2.192Mi ± 0%   -1.40% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14     20.73Mi ± 0%   20.69Mi ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14        22.43Mi ± 0%   22.40Mi ± 0%   -0.14% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14    205.6Mi ± 0%   205.6Mi ± 0%   -0.02% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14         3.015Mi ± 0%   2.703Mi ± 0%  -10.37% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14     21.54Mi ± 0%   21.17Mi ± 0%   -1.74% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14        22.23Mi ± 0%   21.91Mi ± 0%   -1.41% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14    207.3Mi ± 0%   206.9Mi ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14       224.3Mi ± 0%   224.0Mi ± 0%   -0.14% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14   2.008Gi ± 0%   2.008Gi ± 0%   -0.02% (p=0.002 n=6)
geomean                                                       7.195Mi        7.025Mi        -2.36%

                                                            │  1-go1.26   │              3-go1.26              │
                                                            │  allocs/op  │  allocs/op   vs base               │
SamplesJSONUnmarshal/series=10/dp=10/type=floats-14            530.0 ± 0%    462.0 ± 0%  -12.83% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=10/type=histograms-14       4.541k ± 0%   4.463k ± 0%   -1.72% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=floats-14          3.260k ± 0%   3.192k ± 0%   -2.09% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=100/type=histograms-14      43.27k ± 0%   43.19k ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=floats-14         30.30k ± 0%   30.23k ± 0%   -0.22% (p=0.002 n=6)
SamplesJSONUnmarshal/series=10/dp=1000/type=histograms-14     430.3k ± 0%   430.2k ± 0%   -0.02% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=floats-14          5.213k ± 0%   4.515k ± 0%  -13.39% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=10/type=histograms-14      45.31k ± 0%   44.52k ± 0%   -1.76% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=floats-14         32.51k ± 0%   31.82k ± 0%   -2.15% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=100/type=histograms-14     432.6k ± 0%   431.8k ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=floats-14        302.9k ± 0%   302.2k ± 0%   -0.23% (p=0.002 n=6)
SamplesJSONUnmarshal/series=100/dp=1000/type=histograms-14    4.303M ± 0%   4.302M ± 0%   -0.02% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=floats-14         52.02k ± 0%   45.02k ± 0%  -13.45% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=10/type=histograms-14     453.0k ± 0%   445.0k ± 0%   -1.77% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=floats-14        325.0k ± 0%   318.0k ± 0%   -2.15% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=100/type=histograms-14    4.326M ± 0%   4.318M ± 0%   -0.18% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=floats-14       3.029M ± 0%   3.022M ± 0%   -0.23% (p=0.002 n=6)
SamplesJSONUnmarshal/series=1000/dp=1000/type=histograms-14   43.03M ± 0%   43.02M ± 0%   -0.02% (p=0.002 n=6)
geomean                                                       127.8k        123.9k        -3.04%
```

cc @bwplotka


.github/workflows/ci.yml
```diff
@@ -39,14 +39,12 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    env:
-      # Override Go 1.18 security deprecations.
-      GODEBUG: "x509sha1=1,tls10default=1"
```
liggitt, line 44: these are removed in Go 1.27 and are no longer functional, looks like they were only added to keep unit tests green when upgrading to Go 1.18, and the test certificates have since been updated to no longer need this

model/value_histogram127.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (s *HistogramBucket) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	if t, err := dec.ReadToken(); err != nil {
+		return err
+	} else if t.Kind() != jsontext.KindBeginArray {
+		return errors.New("expected [")
+	}
+
+	if err := json.UnmarshalDecode(dec, &s.Boundaries); err != nil {
+		return nil
```
bwplotka: These four branches `return nil` instead of `return err` (also lines 62, 65, 68), so the real parse error gets thrown away. Today json/v2's "read exactly one value" guard still surfaces an error, but callers see `must read or write exactly one value` instead of e.g. `strconv.ParseFloat: parsing "bogus": invalid syntax`, and a later refactor could turn this into a silent success.
```suggestion
		return err
```

model/value_histogram127.go
```diff
@@ -0,0 +1,77 @@
...
+		return errors.New("expected [")
+	}
+
+	if v, err := dec.ReadValue(); err != nil {
+		return err
+	} else if err := s.Timestamp.UnmarshalJSON(v); err != nil {
+		return err
+	}
+
+	if err := json.UnmarshalDecode(dec, &s.Histogram); err != nil {
```
bwplotka, line 40: observation: This moves the nested histogram onto native v2 name matching, which is case-sensitive. I checked on go1.27: `jsonv2.Unmarshal` of `[1,{"Count":"5","Sum":"6"}]` used to go through the v1 fallback and decode fine, and now returns a zeroed histogram with no error. If parity is the goal, `json.UnmarshalDecode(dec, &s.Histogram, json.MatchCaseInsensitiveNames(true))` restores it.

I think since jsonv2 is new it's ok to keep in case sensitive (I checked and we marshal lower case on server side)

model/value.go
```diff
@@ -224,28 +224,6 @@ func (ss SampleStream) MarshalJSON() ([]byte, error) {
 	}
 }
 
-func (ss *SampleStream) UnmarshalJSON(b []byte) error {
```
bwplotka: Same case-sensitivity catch as the nested histogram: with this gone, `jsonv2.Unmarshal` of `{"Metric":{"a":"b"},"Values":[[1,"2"]]}` silently gives an empty SampleStream where the v1 unmarshaler matched case-insensitively. Behaviour on the plain `encoding/json` path is unchanged, but `*SampleStream` also stops implementing `json.Unmarshaler`, so this probably deserves a release note.

model/value_float127.go
```diff
@@ -0,0 +1,63 @@
+//go:build go1.27
```
bwplotka: TIL you can turn off the `nojsonv2` experiment 🤔 

Should we add `//go:build go1.27 && goexperiment.jsonv2` on all of those files or not worth it?

---- 

`GOEXPERIMENT=nojsonv2 go build ./model/` fails on go1.27 with `build constraints exclude all Go files in .../encoding/json/v2`, since the stdlib files are still guarded by the experiment tag. Adding the tag here (and in the other 127 files) fixes it and keeps the default build and tests green, I verified both.
```suggestion
//go:build go1.27 && goexperiment.jsonv2
```

model/value_float127_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{Value: SampleValue(1)},
+		{Timestamp: Time(1), Value: SampleValue(1)},
+	}
+
+	for _, v := range cases {
+		testV1V2Marshal(t, fmt.Sprintf("%#v", v), v)
+		testRoundTrip(t, fmt.Sprintf("%#v", v), v)
+	}
+
+	for _, input := range []string{`null`, `[]`, `[1.123]`, `[1.123,"2"]`, `[1.123,"2","bogus","trailing","data"]`} {
```
bwplotka: Worth adding `[null,"2"]` and `[1.123,null]` here: v1 never calls UnmarshalJSON for a `null` array element so it succeeds and leaves the field untouched, while the new path feeds the literal `null` to Time/SampleValue and errors. Looks unintentional given you emulate the other v1 quirks.

model/time127_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Run("decode_error_"+input, func(t *testing.T) {
+		s1 := newObj()
+		v1err := s1.UnmarshalJSON([]byte(input))
+		if v1err == nil {
+			t.Fatalf("expected error, got none from v1")
+		}
+
+		s2 := newObj()
+		v2err := jsonv2.Unmarshal([]byte(input), s2)
+		if v2err == nil {
```
bwplotka: This only checks that both sides returned something non-nil, which is why the swallowed errors in `HistogramBucket.UnmarshalJSONFrom` slip through. I flipped those four `return nil` to `return err` locally and not a single test changed result. A `wantErrContains` per case would catch it.

model/value127_test.go
```diff
@@ -0,0 +1,62 @@
...
+		{Values: []SamplePair{}},
+		{Values: []SamplePair{{}}},
+		{Values: []SamplePair{{Timestamp: Time(1), Value: SampleValue(1)}}},
+		{Histograms: []SampleHistogramPair{{Timestamp: Time(1), Histogram: &SampleHistogram{}}}},
+		{Metric: Metric{}, Values: []SamplePair{}},
+		{Metric: Metric{}, Values: []SamplePair{{}}, Histograms: []SampleHistogramPair{{Timestamp: Time(1), Histogram: &SampleHistogram{}}}},
+	}
+
+	for i, v := range cases {
+		testV1V2Marshal(t, fmt.Sprintf("%d_%#v", i, v), v)
```
bwplotka: Quick nit: `%#v` on a SampleStream with histograms bakes the pointer address into the subtest name, so names differ on every run and `-run` filters aren't reusable. An index plus a short label ("values_only", "hist_only") reads better.

model/api_bench_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		floatMatrix = append(floatMatrix, fss)
+		histogramMatrix = append(histogramMatrix, hss)
+	}
+	return floatMatrix, histogramMatrix
+}
+
+func BenchmarkSamplesJSONUnmarshal(b *testing.B) {
+	for _, timeseriesCount := range []int{10, 100, 1000} {
```
bwplotka: The 1000x1000 leaf is ~39s/op and a single `-count=1` run of this benchmark takes ~155s and peaks around 1.8GB RSS, so the `-count=6` runs the project asks for would exceed the default 10m test timeout. Trimming to something like series {1,100} x dp {10,1000} would keep it usable.

Makefile.common
```diff
@@ -61,7 +61,7 @@ PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_
 SKIP_GOLANGCI_LINT :=
 GOLANGCI_LINT :=
 GOLANGCI_LINT_OPTS ?=
-GOLANGCI_LINT_VERSION ?= v2.12.2
+GOLANGCI_LINT_VERSION ?= v2.13.2
```
bwplotka: Makefile.common says at the top to only open PRs against prometheus/prometheus, so this bump gets reverted by the next sync. Probably needs to land upstream and come back via the sync instead.

model/time127_test.go
```diff
@@ -0,0 +1,176 @@
+//go:build go1.27
+
+// Copyright The Prometheus Authors
```
bwplotka, line 1: Copyright should be before go build? (at least that's our convention).

model/value127_test.go
```diff
@@ -0,0 +1,62 @@
+//go:build go1.27
+
+// Copyright The Prometheus Authors
```
bwplotka, line 1: ditto, Copyright first?

model/value_float127.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"encoding/json/jsontext"
+	"errors"
+)
+
+func (s *SamplePair) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	if t, err := dec.ReadToken(); err != nil {
+		return err
+	} else if t.Kind() == jsontext.KindNull {
```
bwplotka, line 29: readability: no need for else here and everywhere we return early below (simpler indent, what we follow in Prometheus) -- or do we see some compiler perf improvements from this?

model/value127_test.go
bwplotka: Will stop commenting but this is repeated everywhere else

model/value_histogram127.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"encoding/json/jsontext"
+	"encoding/json/v2"
+	"errors"
+)
+
+func (s *SampleHistogramPair) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	if t, err := dec.ReadToken(); err != nil {
+		return err
```
bwplotka, line 27: ditto early returns allow us to avoid elses

model/value127_test.go
```diff
@@ -0,0 +1,62 @@
...
+		{Metric: Metric{}, Values: []SamplePair{{}}, Histograms: []SampleHistogramPair{{Timestamp: Time(1), Histogram: &SampleHistogram{}}}},
+	}
+
+	for i, v := range cases {
+		testV1V2Marshal(t, fmt.Sprintf("%d_%#v", i, v), v)
+		testRoundTrip(t, fmt.Sprintf("%d_%#v", i, v), v)
+	}
+}
+
+func TestSampleHistogramPairJSONV2(t *testing.T) {
```
bwplotka: Should this be in value_histogram127_test.go for consistency?

bwplotka (commented):
Thanks! Generally ok, but I have some suggestion. Especially around error handling, see comments.

model/value_float127.go
liggitt: doing that required breaking  `t, err := dec.ReadToken()` into its own statement, and leaks the `t` variable to subsequent code... the `else` construct let me keep it scoped to this block.

Keeping each token contained to a small if/else scope seemed cleaner, but I don't feel strongly, if you want me to break it apart I can, just let me know

model/value_histogram127.go
liggitt: same reason as https://github.com/prometheus/common/pull/991#discussion_r4072519383

model/value_float127.go
liggitt: I don't think that's worth it ... anyone who cares about json/v2 will use Go 1.27 now.

model/value_histogram127.go
liggitt: Actually, it uses the options baked into the `dec` decoder. When the caller decoded using `encoding/json`, those options default to compatibility with json/v1 behavior, including case-insensitivity, toleration for duplicate fields, and unknown fields.

I added a test to demonstrate that.

model/value_histogram127.go
liggitt: good catch, fixed

model/value_float127_test.go
liggitt: added tests for these cases to all the types I added custom decoding for, fixed the impls, and re-benchmarked... no change in allocs, CPU was somewhat better on histogram samples, slightly (~4-9%) worse on float samples, but still far better than unoptimized jsonv2 (~60-75% CPU improvements across the board)

Makefile.common
liggitt: hmm... that is already at v2.13.1, which would work fine, but the auto-bump in https://github.com/prometheus/common/pull/984 failed CI for some other reason ... if I set this to v2.13.1 to match current prometheus/prometheus, it won't get reverted whenever the auto-bump gets in, right?

liggitt: updated this to match the current version in prometheus/prometheus

liggitt:
all comments except https://github.com/prometheus/common/pull/991#discussion_r4072519383 addressed in `fixup!` commits, will squash once re-reviewed

Let me know if you want me to break apart the if/else blocks for https://github.com/prometheus/common/pull/991#discussion_r4072519383 ... I feel like it makes things *less* readable by letting token / value variables escape scope, but it's up to you

model/value_float127.go
bwplotka: Yea, that's the trade-off but is scoping giving us a lot here?

I do prefer no else from readability standpoint (https://github.com/prometheus/common/pull/991/changes#r4079951263) but not a big deal (:

model/value_float127.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Read the final end array token.
+	if t, err := dec.ReadToken(); err != nil {
+		return err
+	} else if t.Kind() != jsontext.KindEndArray {
+		return errors.New("expected ]")
+	}
+
+	return nil
+}
```
bwplotka, line 70:
```suggestion
    t, err := dec.ReadToken()
	if err != nil {
		return err
	}
	if t.Kind() == jsontext.KindNull {
		return nil
	}
	if t.Kind() != jsontext.KindBeginArray {
		return errors.New("expected [")
	}
	// Loop until we see the end of the array (tolerate arrays of any length).
	for i := 0; dec.PeekKind() != jsontext.KindEndArray; i++ {
		switch i {
		case 0:
			v, err := dec.ReadValue()
			if err != nil {
				return err
			}
			if bytes.Equal(v, nullBytes) {
				// Skip null values to match SamplePair#UnmarshalJSON behavior.
				break
			}
			if err := s.Timestamp.UnmarshalJSON(v); err != nil {
				return err
			}
		case 1:
			v, err := dec.ReadValue()
			if err != nil {
				return err
			}
			if bytes.Equal(v, nullBytes) {
				// Skip null values to match SamplePair#UnmarshalJSON behavior.
				break
			}
			if err := s.Value.UnmarshalJSON(v); err != nil {
				return err
			}
		default:
			// Skip any remaining array items to match UnmarshalJSON behavior.
			if err := dec.SkipValue(); err != nil {
				return err
			}
		}
	}
	// Read the final end array token.
	t, err = dec.ReadToken()
	if err != nil {
		return err
	}
	if t.Kind() != jsontext.KindEndArray {
		return errors.New("expected ]")
	}
	return nil
}
```

bwplotka (approved):
Thanks! I'd personally still unroll else cases, but not a big deal.

model/value_float127.go
bwplotka: Ok, let's decide --  let's use your version since you benchmarked all, we can always change form slightly later

bwplotka:
Let's stick to your version https://github.com/prometheus/common/pull/991#discussion_r4081037198

bwplotka:
> all comments except https://github.com/prometheus/common/pull/991#discussion_r4072519383 addressed in fixup! commits, will squash once re-reviewed

I assume you want to clean the history before merging

liggitt:
rebased and squashed

bwplotka (approved):

